PR #11, opened 2015-10-25
Start a list of (G)UI options.

deech:


alanz:
Thanks for the PR. Would you like me to add you as a member of the project?

This project is actually badly named. It is intended to provide backend services for existing IDEs, rather than to be an IDE is its own right.

As such, I think the intent of this list may be misplaced.

deech:
Sure. That would be great. :) Feel free to decline the PR.

alanz:
Added as project member, declining PR

PR #21, opened 2015-11-05
fixed typo

ankhers:


alanz:
Thanks, good to you someone is actually reading it. I will merge it once I have pushed my current WIP.

PR #29, opened 2015-11-06
Is build system deterministic

gracjan:


alanz:
Not completely, as there is a cache. Which means you may get second-time-around effects.

It is possible to clear the cache on travis. See https://travis-ci.org/haskell/haskell-ide-engine/caches

gracjan:
I see this as a real problem. I won't feel save clicking 'Merge' button knowing that it might show green but not be really green.

Is there a way to make it not happen ever?

I do not understand the invalidation rules for Travis but then I see that if `.travis.yml` has changed it should invalidate cache, should it not?

gracjan:
I see that the issue in this case is that `stack` does not recompile if flag `--pedantic` is added or removed, is that right?

alanz:
I think we can do a `stack clean` immediately prior to the test itself, then run the test with --pedantic

alanz:
I think the problem was the dependency HaRe requiring too much RAM, which broke things. A new version has been released that compiles with -O0, reducing memory usage.

PR #42, opened 2015-11-08
Make web interface optional and port configurable

cocreature:
I'm not super happy with "web" as an option name but I couldn't come up with anything better.


alanz:
If you are not happy with web, how about http?

cocreature:
That's better, updated.

PR #44, opened 2015-11-08
Remove Context, simplifying IdeRequest

### LINKED ISSUE #40: Remove Context from IdeRequest
The context variables should rather come through as standard parameters

alanz:
Closes #40


cocreature:
lgtm now

Issue #45, opened 2015-11-08
Encourage stateless programming style


gracjan:
Each command should be as self-contained as possible, It should preferably not depend on hidden state at all. Visible state, like contents of visible files, is less of a problem.

Encourage stateless programming in what is supposed to be async system.

Encourage using full absolute paths to files because current working directory is state.

bitemyapp:
Tend to agree, although for performance (later?) may want to mediate access to project state (probably namespaced by project path root or some-such).

alanz:
I also tend to agree, in particular with the approach of always running hie wrt a given, known directory, and using absolute file paths.

PR #47, opened 2015-11-08
Remove state cwd

gracjan:
Do not suggest or encourage using state in plugins, not in the form of current working directory.

Reference #45.

### DIFF
--- a/src/Haskell/Ide/Engine/BasePlugin.hs
+++ b/src/Haskell/Ide/Engine/BasePlugin.hs
@@ -4,7 +4,6 @@
 module Haskell.Ide.Engine.BasePlugin where
 
 import           Control.Monad
-import           Control.Monad.IO.Class
 import           Data.Aeson
 import           Data.Foldable
 import           Data.List
@@ -20,7 +19,6 @@ import           Haskell.Ide.Engine.PluginUtils
 import           Options.Applicative.Simple (simpleVersion)
 import qualified Paths_haskell_ide_engine as Meta
 import           Prelude hiding (log)
-import           System.Directory
 
 -- ---------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


src/Haskell/Ide/Engine/BasePlugin.hs
```diff
@@ -67,26 +65,6 @@ baseDescriptor = PluginDescriptor
                         }
           , cmdFunc = commandDetailCmd
           }
-      , Command
-          { cmdDesc = CommandDesc
-                        { cmdName = "pwd"
```
alanz: I think it may be worthwhile keeping this one, but removing "cwd"

gracjan:
@cocreature: Refactored it out to a function.

src/Haskell/Ide/Engine/BasePlugin.hs
gracjan: Why? pwd is a constant in this case.

alanz:
There is some logic in setting the initial directory, with a "give up leave it unchanged" option.

I do not believe there is any harm in being able to request the current working directory. It is not a stateful operation.

alanz:
BTW, I am just working through some ghc-mod commands, in particular the "find" command, which works in the context of the current project, which it detects using the find cradle process.

The cradle is defined as the point where the cabal file is found, and the search starts at the current directory.

So in the normal  (non-HIE) case, ghc-mod is started up in the project directory and can then find the cradle by looking in the current directory or moving up.

In the stateless mode, we can work around this by setting the `CtxProject` to require a directory name, and then doing an explicit cd in the ghc-mod plugin, or modifying ghc-mod to take a directory as its root.

We also need to consider the case where there is more than one HIE instance running at a time, attached to different projects.

gracjan:
Yes, `find` command should get an explicit `ctxProject`, an absolute path
to where to begin searching.

2015-11-11 9:14 GMT+01:00 Alan Zimmerman <notifications@github.com>:

gracjan:
And about having a `pwd` without a `cwd`: somebody will come along and contribute the obvious missing part, that is `cwd`. I would not want to give people suggestions for that path.

alanz:
ok, you've worn me down. I'll have to sneak it into a private plugin if I
want to debug.

Maybe we should just log the directory chosen on startup. Then we are both
happy.

alanz:
Can you rebase this, then merge?

PR #51, opened 2015-11-09
Dispatcher validation

alanz:
Move request validation into the dispatcher.

Now the `cmdFunc` is only ever called if the parameters in its request comply with at least one of its `AcceptedContext` values and all additional reuired parameters are present.


alanz:
@gracjan @DanielG @mgsloan @cocreature and all others, this is quite a big change, please review

PR #53, opened 2015-11-10
Improve console

alanz:
Add some basic help, and parse parameters according to the `CommandDescription` so that the types line up when passed to the dispatcher.


alanz:
@JPMoresmau this should improve your console experience

JPMoresmau:
Great, works!

alanz:
And see also
https://github.com/haskell/haskell-ide-engine/blob/master/docs/IdeIntegration.md

It is the start of a howto for integrating into an IDE

PR #56, opened 2015-11-11
Use a GADT parametrized by param type, fixes #54

### LINKED ISSUE #54: Worker function parameter passing
A `CommandFunc` is defined as

```haskell
type CommandFunc = forall m. (MonadIO m,GHC.GhcMonad m,HasIdeState m)
                => [AcceptedContext] -> IdeRequest -> m IdeResponse
```

It will only be called by the dispatcher if at least one context matches, and and the required parameters are present and of the right type.

At the moment the command processing function is very messy

```haskell
checkCmd :: CommandFunc
checkCmd _ctxs req = do
  case getParams ["file"] req of
    Left err -> return err
    Right [ParamFile fileName] -> do
      liftIO $ runGhcModCommand (GM.checkSyntax [T.unpack fileName])
    Right x -> error $ "GhcModPlugin.checkCmd: got unexpected file param:" ++ show x
```

In this case there is only one valid context, so it it is called at all the "file" parameter is guaranteed to be valid. The `getParams` call returns a list of `ParamValue` in the order of the param names given to it, or a error result if they are missing. This call is actually redundant now.

It may be simpler to define the following types

```haskell
data ActualRequiredParams =
    AR0
  | AR1 ParamVal
  | AR2 ParamVal ParamVal
 ...
```
 
```haskell
data ActualOptionalParams =
    AO0
  | AO1 (Maybe ParamVal)
  | AO2 (Maybe ParamVal) (Maybe ParamVal)
 ...
```

as well as something similar for the matching contexts.

The `CommandFunc` would then be defined as

```haskell
type CommandFunc = forall m. (MonadIO m,GHC.GhcMonad m,HasIdeState m)
                => [ActualContextContext] -> ActualRequiredParams -> ActualOptionalParams -> m IdeResponse
```

This will still leave is with "missing case" warning messages though.

Is there some clever way to do this with types?

cocreature:
So apart from my terrible names (that's why it's marked as WIP), there are a couple of things here.

- The `Rec` datatype is the same as the one from Vinyl, I am not sure if it's worth pulling that in as a dependency for only one type.
- The duplication is not nice, the problem is that we need to stuff that in the map and maps want a monomorphic type, so you would need to put it in a RankNType to hide the tag, but that still doesn't work because now you can no longer derive Generic and Show. Show is easily fixable using StandaloneDeriving, Generic is not because you can't derive Generic for RankNTypes. We can just not derive Generic, but we need that for the json stuff, so we would need to manually write a ToJSON and FromJSON instance for ParamMap which doesn't seem to be easily possible since you can't make instances of RankNTypes. I'll try to find a solution later, but I doubt we'll get around the duplication.
- We can't use standard list syntax (overloaded lists also don't allow for that kind of list).
- We have to specify the types of the parameters passed to `getParams`. I actually consider that to be a good thing, but it might be annoying sometimes.

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -54,6 +54,7 @@ library
                      , transformers
                      , wai
                      , warp
+                     , vinyl >= 0.5 && < 0.6
   ghc-options:         -Wall
   default-language:    Haskell2010
 
--- a/haskell-ide-example-plugin2/Haskell/Ide/ExamplePlugin2.hs
+++ b/haskell-ide-example-plugin2/Haskell/Ide/ExamplePlugin2.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE GADTs #-}
 module Haskell.Ide.ExamplePlugin2 where
 
 import Haskell.Ide.Engine.PluginDescriptor
@@ -50,7 +51,7 @@ sayHelloToCmd :: CommandFunc
 sayHelloToCmd _ req = do
   case Map.lookup "name" (ideParams req) of
     Nothing -> return $ IdeResponseFail "expecting parameter `name`"
-    Just (ParamText n) -> do
+    Just (ParamValP (ParamText n)) -> do
       r <- liftIO $ sayHelloTo n
       return $ IdeResponseOk (String r)
     Just x -> return $ IdeResponseFail (toJSON $ T.pack $ "got wrong param type:" ++ show x)
--- a/haskell-ide-example-plugin2/Haskell/Ide/ExamplePlugin2.hs
+++ b/haskell-ide-example-plugin2/Haskell/Ide/ExamplePlugin2.hs
@@ -51,7 +51,7 @@ sayHelloToCmd :: CommandFunc
 sayHelloToCmd _ req = do
   case Map.lookup "name" (ideParams req) of
     Nothing -> return $ IdeResponseFail "expecting parameter `name`"
-    Just (ParamValP (ParamText n)) -> do
+    Just (ParamTextP n) -> do
       r <- liftIO $ sayHelloTo n
       return $ IdeResponseOk (String r)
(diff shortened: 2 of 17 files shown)


alanz:
In terms of Generic, I think we decided to hand-roll all JSON instances to make sure that external interfaces stay stable as compiler versions change, so we can do away with it for now.

alanz:
Looks good though.

cocreature:
Yeah the problem is not actually Generic, but that we don't get a `ToJSON` and `FromJSON` instance of `ParamMap`. I think we can work around that by making it a newtype. I'll play around with it a bit more.

alanz:
What we can do is have the dispatcher convert from the original ParamMap to
the new GADT version when it does its type checking. That way we separate
the external transport level interaction from the internal plugin one.

In other  words a plugin will always be called with valid params.

Because the next step is to define the CommandFunc as one of those Rec
instances, so we can only call it with the right stuff.

alanz:
@cocreature Please indicate when you are done with this, so we can do a final review/merge

cocreature:
Will do, I want to clean up some minor things before.

cocreature:
I'm done with this now. If you want I can squash the commits, not sure what's easier for you to review.

alanz:
I am happy to keep the history, and github shows a diff for the whole PR, so no reason to squash it

cocreature:
Also `ParamValP` is a pretty crappy name, but I couldn't come up with something that made sense.

alanz:
If we are reusing the `Rec` type from vinyl, is there any way we can use it directly?
And if not we should put in a comment noting its origin/inspiration.

Otherwise lgtm. Thanks.

alanz:
Well, it just adds a 'P' to what we had before, so it drops in neatly.

cocreature:
We could depend on vinyl, I just wasn't sure if it's worth depending on it for a single type. If you think that's a good idea I'll change it.

alanz:
By depending on vinyl we expose the documentation, so people can trace it
through and see how it works.

cocreature:
Updated.

PR #57, opened 2015-11-11
Ghcmod commands

alanz:
Adding some more ghc-mod commands

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -96,6 +96,7 @@ test-suite haskell-ide-test
   build-depends:       base
                      , aeson
                      , containers
+                     , directory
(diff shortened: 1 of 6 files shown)


alanz:
@gracjan, @JPMoresmau, @cocreature 

Please review this and merge if happy

gracjan:
@alanz: Can you rebase it in a way so that commit 400fac7 goes completely away? `git rebase origin/master` should do the trick.

gracjan:
Same with commits 3990a62 and f3f3c93.

gracjan:
I think you should just squash it as one single commit. There is still some unnecessary history left in here.

gracjan:
Otherwise it looks great.

PR #59, opened 2015-11-11
Expect json input separated by STX, fix #43

### LINKED ISSUE #43: Consume invalid input
pipes-parse doesn't consume the input if it doesn't parse which results in us repeatedly trying to parse the same input which results in an perpetual flood of error messages.

My pipes foo is too weak to figure out how to fix this right now.

cocreature:
So we now exit if we get invalid input which means either a parse failure or some random stuff between a parse and a separator. in particular that means that each message must be followed by a separator directly which seems reasonable to me.

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/JsonStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/JsonStdio.hs
@@ -5,12 +5,14 @@
 
 module Haskell.Ide.Engine.Transport.JsonStdio where
 
+import           Control.Applicative
 import           Control.Concurrent
 import           Control.Lens (view)
 import           Control.Logging
 import           Control.Monad.State.Strict
 import qualified Data.Aeson as A
 import qualified Data.Attoparsec.ByteString as AB
+import qualified Data.Attoparsec.ByteString.Char8 as AB
 import qualified Data.ByteString.Char8 as B
 import qualified Data.ByteString.Lazy as BL
 import           Data.Char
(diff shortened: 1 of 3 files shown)


src/Haskell/Ide/Engine/Transport/JsonStdio.hs
```diff
@@ -25,19 +27,27 @@ jsonStdioTransport cin = do
...
+      loop (cid + 1) (P.drop (1::Int) stream)
     loop cid stream = do
       debug "jsonStdioTransport:calling go"
-      (req,stream') <- runStateT decodeMsg stream
+      let splitStream = view (P.break (== fromIntegral (ord '\STX'))) stream
+      (req,res) <- runStateT decodeMsg splitStream
       debug $ T.pack $ "jsonStdioTransport:got:" ++ show req
       case req of
         Just (Left err) -> do
           putStr $ show (HieError (A.String $ T.pack $ show err))
```
alanz: Can we change this to be 

```haskell
rsp = IdeResponseError (toJSON (HieError .......))
BL.putStr $ A.encode (channelToWire rsp)
```

i.e. send it back on the wire as a properly formatted message.

And perhaps also log it at debug level.

src/Haskell/Ide/Engine/Transport/JsonStdio.hs
```diff
@@ -25,19 +27,27 @@ jsonStdioTransport cin = do
...
-          writeChan cin (wireToChannel cout cid r)
-          rsp <- readChan cout
-          BL.putStr $ A.encode (channelToWire rsp)
-          loop (cid + 1) stream'
+          nextStreamOrError <- next res
+          case nextStreamOrError of
+            Left nextStream -> processParse cid r nextStream
+            Right (a,restConsumer) -> do (restLength,nextStream) <- P.fold' (\n bs -> n + fromIntegral (B.length bs)) 0 id restConsumer
+                                         if (B.length a + restLength > 0) then
+                                           putStr $ show (HieError (A.String $ "Unexpected input before separator"))
```
alanz: Perhaps debug log, and send back wire format error

src/Haskell/Ide/Engine/Transport/JsonStdio.hs
cocreature: Good point, I was planning to change the error handling but somehow forgot about it. I'll try to get to it later today.

cocreature:
Updated the error handling, I chose to for an empty string as the plugin name since we don't have anything reasonable to use.

gracjan:
Hmm... is this one way only? I mean, does HIE also produce json-stream that is \STX separated?

I think both ways streaming should look the same and use same separator character (if any). Note that HIE might also accidentally spit garbage to stdout because it is pretty easy to do this in Haskell while coding a plugin, or when a plugin calls an external executable.

cocreature:
@gracjan It would be great if you could figure out why this is failing on travis. The tests ran through flawlessly locally.

PR #63, opened 2015-11-12
Short cut sleeping when result is available

gracjan:
Significantly speed up Emacs testing.


alanz:
Looks fine to me, based on my limited elisp knowledge

gracjan:
Hmm, it fails... I'll debug this.

PR #65, opened 2015-11-12
Enrich plugins call

gracjan:
An Editor's side it is rare to work with a single command so it makes sense to batch queries about all available commands and command details.

This commits puts all the information about a plugin into 'plugins' call and remove the (now redundant) 'commands' and 'commandDetails' calls.


alanz:
I think a top-level "plugins" command makes sense, but I also think it is worth keeping the other two.

I would rather have a richer command set for editor integrators

gracjan:
Okay, in that case I'm not touching 'commands' and 'commandDetail'.

alanz:
thanks

PR #72, opened 2015-11-13
Async dispatch

### LINKED ISSUE #67: Bring in async processing
```
<alanz> And then to experiment with some sort of ghci/inferior repl option
<gracjan> guy, I need the async part to get going forward, I mean request-id, response-id
<gracjan> Id like to setup emacs to handle responses in async way only
<gracjan> so this has to be in before I start
<alanz> ok, that is actually one of the things I have been thinking quite a bit about.
<alanz> I know how to do it in the dispatcher, maybe I should make a version. I have been waiting for the other stuff going through the dispatcher to settle a bit first.
* fgaz (~fgaz@5.90.35.97) has joined #haskell-ide-engine
<alanz> The key is the the reply Chan in the  ChannelRequest.
<gracjan> emacs is fully callback based
<gracjan> so I’ll make it completely async from day one
<alanz> We can let the plugin dispatcher decide whether to do a sync or async call into the plugin, and so long as the reply eventually goes to the Chan, which goes into the transport, we are good.
<gracjan> okay
<alanz> Once the framing in the stdio is settled, it is a simple matter to split it into a pair of listeners.
<alanz> One listens on stdin and sends to dispatcher. The other listens on reply Chan and forwards to stdout.
<alanz> The only think is to decide what the management process around async is at the plugin level.
<alanz> One part of me says every plugin should run in its own thread as a matter of routine, so the dispatcher is never blocked
<gracjan> true
<alanz> but it is coming....
<gracjan> or even every request should work in a fresh thread
<alanz> that is up to the plugin dispatcher.
<alanz> So we repeat the process. a plugin listens on a chan, then either does a sync process or spawns off a new thread, based on any plugin-specific criteria.
<alanz> Either way the response goes back into the Chan that ends up at the transpot
<alanz> It is quite a general way of doing it
<alanz> so it can cater for whatever has to happen.
<alanz> We will probably have to put syncronisation points in somewhere so we do not trigger a whole lot of project rebuilds for example
<alanz> But that is why the service layer is in the plugin descriptor
```

alanz:
Closes #67 

A plugin can decide on a per-command basis whether it should have the option to return an async result.

If so, it takes responsibility for managing whatever threads are required to create the async processing, but each message has a callback function for when the result is ready.

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/JsonStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/JsonStdio.hs
@@ -6,9 +6,11 @@
 module Haskell.Ide.Engine.Transport.JsonStdio where
 
 import           Control.Applicative
-import           Control.Concurrent
+import           Control.Concurrent.STM.TChan
 import           Control.Lens (view)
 import           Control.Logging
+import           Control.Monad.IO.Class
+import           Control.Monad.STM
 import           Control.Monad.State.Strict
 import qualified Data.Aeson as A
 import qualified Data.Attoparsec.ByteString as AB
@@ -28,9 +30,9 @@ import qualified Pipes.Prelude as P
 import           System.IO
 
 -- TODO: Can pass in a handle, then it is general
-jsonStdioTransport :: Chan ChannelRequest -> IO ()
+jsonStdioTransport :: TChan ChannelRequest -> IO ()
 jsonStdioTransport cin = do
-  cout <- newChan :: IO (Chan ChannelResponse)
+  cout <- atomically $ newTChan :: IO (TChan ChannelResponse)
   hSetBuffering stdout NoBuffering
   P.runEffect (parseFrames PB.stdin P.>-> parseToJsonPipe cin cout 1 P.>-> jsonConsumer)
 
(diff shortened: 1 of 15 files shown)


src/Haskell/Ide/Engine/Transport/JsonStdio.hs
```diff
@@ -34,8 +36,8 @@ jsonStdioTransport cin = do
           putStr $ show (HieError (A.String $ T.pack $ show err))
           loop (cid + 1) stream'
         Just (Right r) -> do
-          writeChan cin (wireToChannel cout cid r)
-          rsp <- readChan cout
+          atomically $ writeTChan cin (wireToChannel cout cid r)
```
cocreature: Even if the plugin is async we block completely at this point, which seems like the wrong thing to do, since we don't gain anything from having async that way. But we should probably try to get the reworked `jsonStdioTransport` before changing that.

cocreature: Also we can do that in a separate pr.

alanz: I have not touched the transport side, pending the STX PR.

Once it is done, this will have to be split into two processes, one which reads on stdio and writes to the Chan to the dispatcher, the other waits on the return Chan and relays on to stdout.

But better not to mix two things at once, it can wait.

PR #73, opened 2015-11-13
Remove invalid input even if not handled

gracjan:


cocreature:
lgtm

gracjan:
I see you like living on the edge.

PR #88, opened 2015-11-22
Support commands that only require a context

cocreature:
This supports commands like ghc-mod’s type command. Currently the information is only printed to the mini buffer.

Also there are some ugly workaround for what I consider flaws in our protocol. I’ll clean it up and try to fix the protocol, where I feel necessary and then we can merge this.

I’m not an elisp wizard so feel free to tell me what I should change.

### DIFF
--- a/elisp/haskell-ide-engine.el
+++ b/elisp/haskell-ide-engine.el
@@ -164,9 +164,9 @@ association lists and count on HIE to use default values there."
   (format "%s:%s" (car cmd) (cdr cmd)))
 
 (defun hie-get-context (context)
-  (let ((start (save-excursion (goto-char (region-beginning))
+  (let ((start (save-excursion (if (use-region-p) (goto-char (region-beginning)))
                                (list (line-number-at-pos) (current-column))))
-        (end (save-excursion (goto-char (region-end))
+        (end (save-excursion (if (use-region-p) (goto-char (region-end)))
                              (list (line-number-at-pos) (current-column))))
         (filename (buffer-file-name)))
     `(("file" . (("file" . ,filename)))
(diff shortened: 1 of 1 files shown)


gracjan:
Nice! Is it merge worthy already?

elisp/haskell-ide-engine.el
```diff
@@ -145,13 +151,38 @@ association lists and count on HIE to use default values there."
...
+                               (list (line-number-at-pos) (current-column))))
+        (end (save-excursion (goto-char (region-end))
+                             (list (line-number-at-pos) (current-column))))
+        (filename (buffer-file-name)))
+    `(("file" . (("file" . ,filename)))
+      ("start_pos" . (("pos" . ,(apply 'vector start))))
+      ("end_pos" . (("pos" . ,(apply 'vector end)))))))
 
 (defun hie-run-command (plugin command)
   (setq haskell-ide-engine-process-handle-message
```
gracjan: `haskell-ide-engine-process-handle-message` was meant for debugging purposes, really. There should be a map from send command ids to functions handling responses.

cocreature: Definitely, but we can do that separately imho. Moving in small steps is easier and faster.

cocreature:
I wanted to wait for the plugin name changes in #90 which are now in master and I adapted this pr to work with them. So I consider it to be merge worthy now. It’s definitely not perfect, but it’s a step in the right direction and a clear improvement over the previous state which was completely broken by changes to the json format.

alanz:
@gracjan are you ok with this?

gracjan:
Let we review the new commits, okay? Should be today.

cocreature:
Just did another commit to adapt to the changes in #93.

gracjan:
There is a need for one change, that is use region only when region is active. Otherwise looks good. You write very clean nice lisp!

PR #89, opened 2015-11-22
Add plugin name to command descriptor

cocreature:
The reason i want this is so that I don’t have to store the plugin name on the IDE side when calling `commandDetail`. This also ties in with our "stateless philosophy".

I am not super happy with sprinkling it all over the place, but I couldn’t come up with something better. Suggestions welcome.

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -46,11 +46,11 @@ plugins = Map.fromList
   [
     -- Note: statically including known plugins. In future this map could be set
     -- up via a config file of some kind.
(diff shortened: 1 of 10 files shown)


alanz:
I am not sure this is needed.

In a given configuration we already have it, and we should be sending back a structure which is basically "for plugin 'hare' we have these commands"

cocreature:
Ah yeah, I just saw that we already have the plugin name in `commandDetail`, then we only need to change the format of the response there.

PR #99, opened 2015-11-25
Create elisp functions for all commands

cocreature:
The fact that this uses `eval` is kinda ugly, but we can’t get away with a macro here, since that would expanded when the function is defined and not when it’s applied, so we can’t use any local parameters here which makes it sadly useless for this purpose.

### DIFF
--- a/elisp/haskell-ide-engine.el
+++ b/elisp/haskell-ide-engine.el
@@ -184,21 +184,47 @@ association lists and count on HIE to use default values there."
 
 (defun haskell-ide-engine-handle-first-plugins-command (json)
   "Handle first plugins call."
-
   (let ((menu-items
          (mapcar
           (lambda (plugin)
             (cons (symbol-name (car plugin))
                   (mapcar
                    (lambda (command)
-                     (vector (cdr (assq 'ui_description command)) (list 'hie-run-command (symbol-name (car plugin)) (cdr (assq 'name command)))))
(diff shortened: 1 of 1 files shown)


elisp/haskell-ide-engine.el
```diff
@@ -184,21 +184,47 @@ association lists and count on HIE to use default values there."
...
                    (cdr plugin))))
           (cdr (assq 'plugins json)))))
     (setq haskell-ide-engine-plugins (cdr (assq 'plugins json)))
+    (haskell-ide-engine-create-all-commands command-names)
     (easy-menu-define hie-menu hie-mode-map
       "Menu for Haskell IDE Engine"
       (cons "HIE" menu-items))))
 
+(defun haskell-ide-engine-create-command (plugin command)
+  `(defun ,(intern (concat "haskell-ide-engine-" (symbol-name plugin) command)) ()
```
alanz: Should we perhaps use `hie-` as the prefix instead of `haskell-ide-engine`?

cocreature: Sounds like a good idea, I just went with what was already there. I would do this in a different pr, since it’s a large diff (we need to rename the file to hie.el to conform to the elisp naming conventions). which only clutters what is going on here.

alanz: I don't think it is a big deal, and with auto-compietion not much of a problem in reality. The names will just get very long.

PR #103, opened 2015-11-26
Use dash.el

cocreature:
This is a bit controversial, in particular @gracjan doesn’t seem to like it a lot, but let me explain my reasoning and then see where the disucssion goes

- It makes the code cleaner and simpler by providing combinators for commonly used patterns. That’s the same reason why haskellers love their `Data.List`.
- It gives us a form of pattern matching which is amazing when working with alists which is what json is represented as in emacs.

Now to the counter arguments and why I don’t fully buy them:

- It introduces an external dependency.

elpa gives us dependency management and given that this seems to be how most people install packages these days it’s not a burden for users. People who use git submodules or whatever for managing their packages are expert enough to figure out how to get dash.el. Also a lot of packages already depend on it, so most likely everybody has it anyway.

- It makes it harder for newcomers to contribute.

Our target audience is haskell programmers. Even if they are fairly new to haskell they know what `mapcat` means and understand the basics of pattern matching. If someone understands those combinators I think it makes the code actually easier to read and so it’s in fact easier to contribute instead of harder.

### DIFF
--- a/elisp/hie.el
+++ b/elisp/hie.el
@@ -1,6 +1,7 @@
 ;;; hie.el --- Haskell IDE Engine process -*- lexical-binding: t -*-
 
 ;; Copyright (c) 2015 Haskell Ide Contributors
+;; Package-Requires: ((dash "2.12.1"))
 
 ;; See LICENSE for details.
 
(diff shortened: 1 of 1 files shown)


gracjan:
It removes the ugly `(cdr (assq 'xxx lst))` so I think I'm convinced.

Please do not overuse this. Seems that `dash` has the power to seriously obfuscate code.

cocreature:
:+1: I'll try not to overuse it :)

Issue #79, opened 2015-11-14
Response content type negotiation


alanz:
Based on an IRC discussion with @hamishmack
#61 And #66 are also relevant.

There are two different but related issues
1. Given that we end up with a specific set of well-known types that can be managed by a given IDE, we need to be able to manage versioning of these, if the format changes.
2. If we end up with a set of semantic types / widgets that can be meaningfully applied in an IDE, we will end up with uneven implementation of these in the set of supporting IDEs.

Both of these seem to require some kind of content negotiation.

A possible solution would be to include a description of the content types (in the sense of semantic type / widget) together with explicit versions in each request from the IDE for a given command.  The supported types for a given command could potentially be returned as part of the `CommandDesc`.

mgsloan:
This makes sense to me.  Annotations are a good concrete example of the need to support multiple response types.  We probably wouldn't want to send all the extra annotation metadata in the case that the editor doesn't do anything with it.

cocreature:
@alanz Do you think we should support multiple versions or just add a version parameter to let the IDE detect if it supports the version of hie running or not?

Personally I would prefer the lattter since being able to create multiple versions is a lot of work on our side and might even be impossible since we can't necessarily reconstruct the old information if we remove something.

Apart from that something like a pvp version scheme for our json might make sense. So increment a minor version if only additions or other non breaking changes are introduced which might enable tho IDE to do new things, but doesn't break the previous behavior and major versions for breaking changes.

alanz:
There are at least three things that have versions they know/care about

* The IDE
* HIE
* A Plugin

If there are well-known semantic/widget types, these should probably be versioned from the beginning, as part of the JSON definition. And I think at the JSON message level. (Using JSON as a standing for  "whatever wire protocol is used in this transport")

In the case of a mismatch, depending on what it is, we could have HIE do some kind of mapping between versions to keep a given plugin/IDE combination happy.

Initially, supplying a list of known semantic/widget types to a plugin should allow it to either return the requested type, or fall back to the standard serialisation, with the proviso that the IDE handling of it may not be optimal.

Beyond that, I think we need to get a feel for real world cases. We do not want to overthink it in the beginning, rather get something simple out that works, and then evolve.

alanz:
Superseded by move to LSP

PR #105, opened 2015-11-26
nicer parameters and command return type

JPMoresmau:
The JSON for parameters really leaked our internal structure without being understandable. Changed it to something saner in my opinion.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginDescriptor.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginDescriptor.hs
@@ -302,19 +302,19 @@ ok :: T.Text
 ok = "ok"
 
 instance ValidResponse String where
-  jsWrite s = H.fromList [ok .= toJSON s]
+  jsWrite s = H.fromList [ok .= s]
   jsRead o = o .: ok
 
 instance ValidResponse T.Text where
-  jsWrite s = H.fromList [ok .= toJSON s]
+  jsWrite s = H.fromList [ok .= s]
   jsRead o = o .: ok
 
 instance ValidResponse [String] where
-  jsWrite ss = H.fromList [ok .= toJSON ss]
+  jsWrite ss = H.fromList [ok .= ss]
   jsRead o = o .: ok
 
 instance ValidResponse [T.Text] where
-  jsWrite ss = H.fromList [ok .= toJSON ss]
+  jsWrite ss = H.fromList [ok .= ss]
   jsRead o = o .: ok
 
 instance ValidResponse () where
@@ -363,7 +363,7 @@ instance ValidResponse CommandDescriptor where
 
 instance ValidResponse IdePlugins where
   jsWrite m = H.fromList ["plugins" .= H.fromList
-                ( map (\(k,v)-> k .= toJSON v)
+                ( map (uncurry (.=))
                 $ Map.assocs m)]
   jsRead v = do
     ps <- v .: "plugins"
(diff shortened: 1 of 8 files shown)


JPMoresmau:
I've added another thing: each ValidResponse instance must define a name. This name is used as the return type of the command in the command description, so that IDEs can decide what to do with the result. I've changed all of hare operation to return a RefactorResult, so that it's clear that they all return the same thing (the list of impacted files)

alanz:
I like the return type name

JPMoresmau:
I see you haven't merged the PR. Any concern?

hie-plugin-api/Haskell/Ide/Engine/PluginDescriptor.hs
```diff
@@ -214,7 +233,8 @@ data ParamVal (t :: ParamType) where
 class ValidResponse a where
   jsWrite :: a -> Object -- ^ Serialize to JSON Object
   jsRead  :: Object -> Parser a -- ^ Read from JSON Object
-
+  responseTypeName :: a -> T.Text -- ^ Type name in command description.
```
alanz: I wonder if this should not be the `TypeRep` of the return type, so it actually tags the return type?

This means we are stuck with long names in the serialisation, but at least we get a guarantee about the tag matching the type.

alanz:
Just made a comment. No real concerns, just did not have time to look in
detail.

hie-plugin-api/Haskell/Ide/Engine/PluginDescriptor.hs
cocreature: I like the idea of using `TypeRep`s here. This should also make it easier to find the source producing a particular response.

alanz: Bringing in the discussion from #109, if the type is clearly serialised, will this be necessary?

In other words, do we need a tag for the command context as well as the type?

The one example is a HaRe Diff, which is currently supplied as just a list of files, but will be a full featured diff soon.

Is it important to know that it is a refactor diff rather than just a diff? Can this knowledge be accessed from the command that was sent, and hence the known reply? Should we tag it with the "plugin:command name" instead?

To me this comes back to the discussion about semantic types / widgets.

JPMoresmau: I explicitly put a specific type name for two reasons:
- be able to have several types act as one in the API (String and Text, for example)
- have one layer of indirection between the Haskell types and the types we expose, to avoid leaking our Haskell into our JSON (see what I fixed earlier on parameters RP/OP)

But of course the flip side is that it's easy to define two different structures that return the same responseTypeName with different JSON, so I'll agree we could use TypeRep and avoid types like String.
@alanz  I don't understand your comment about the type being clearly serialized. For me, an IDE needs to know the return type of a command BEFORE it even offers it to the user, because potentially it needs to know what to do with the response. For example, a refactor operation changes the files on disk and return the paths of the change files. the IDE needs to understand that, so that it can reload the files. A generic method that will offer the command to the user without doing anything with the answer will cause issues (the IDE file representation being out of sync with the disk). But at the same time, if I know how to handle a specific return type, I can handle any command that return that type. So you can add loads of Hare refactor commands, if they all return the same type I know the IDE can support them all, without a priori knowledge of which commands to support.

alanz: I think what you are describing as a return type of a command is what I have been calling a semantic type.

From that perspective it makes sense to return something that identifies the semantic type that is returned, but to me it should be a property of hie, not just a string that gets defined ad hoc on a command.

For a semantic type it would make sense to distinguish between a list of files that may have been changed and a list of files that were not changed, for example, even if the underlying type is the same `[FilePath]`.

But I see the semantic types as being a set that grows over time and as we understand what is needed, and become something that every IDE should support in some way as part of hie integration. And we should be able to fall back to a vanilla semantic type that says something like "raw".

And this possibly has a bearing on #79

JPMoresmau: The command return type needs to be a ValidResponse anyway, so if we use TypeRep there'll be an exact correspondance between that string and the response structure the IDE can expect. I'll change to use the TypeRep.

PR #111, opened 2015-11-28
add ShortName and Overview to PluginDescriptor

### LINKED ISSUE #98: Add user visible info to PluginDescriptor
At the moment a `PluginDescriptor` is defined as

```haskell
data PluginDescriptor = PluginDescriptor
  { pdCommands        :: ![Command]
  , pdExposedServices :: ![Service]
  , pdUsedServices    :: ![Service]
  }
```

Add some fields to be used in the IDE UI as follows

```haskell
data PluginDescriptor = PluginDescriptor
  { pdUIShortName     :: !T.Text
  , pdUIOverview      :: !T.Text
  , pdCommands        :: ![Command]
  , pdExposedServices :: ![Service]
  , pdUsedServices    :: ![Service]
  }
```

The `pdUIShortName` should be displayed in menus and the like, rather than the `PluginId`.
The `pdUIOverview` can be displayed in any kind of "more information" scenario about the plugin.

bergey:
Text descriptions intended for the editor user, rather than for
messages between the editor and HIE.

closes #98 

I took the Overview text from the tools' Hackage descriptions, but no doubt they could be improved.

### DIFF
--- a/hie-example-plugin2/Haskell/Ide/ExamplePlugin2.hs
+++ b/hie-example-plugin2/Haskell/Ide/ExamplePlugin2.hs
@@ -15,7 +15,9 @@ import qualified Data.Text as T
 example2Descriptor :: PluginDescriptor
 example2Descriptor = PluginDescriptor
   {
-    pdCommands =
+    pdUIShortName = "Hello World"
+  , pdUIOverview = "An example of writing an HIE plugin"
+  , pdCommands =
       [
         Command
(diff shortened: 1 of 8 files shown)


alanz:
Thanks, looks good, but you need to update the quickcheck tests by adding `arbitrary` instances for the two new fields.

bergey:
Thanks for pointing that out.  So good to have tests.

PR #115, opened 2015-11-30
Detect plugin param name collisions

### LINKED ISSUE #52: Check PluginDescriptor on loading
When HIE starts up, it is presented with a list of `Plugin`s.

Each of these defines commands in a `CommandDescriptor`

For each `CommandDescriptor`, check that the parameter names introduced via `cmdContexts` and `cmdAdditionalParameters` are unique.

If this is not the case, exit with an error.

tobiasgwaaler:
Fixes #52
I highly appreciate feedback on style/readability etc. :)


hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
```diff
@@ -9,10 +9,16 @@ module Haskell.Ide.Engine.PluginUtils
   , mapEithers
   , missingParameter
   , incorrectParameter
+  , validatePlugins
```
alanz: I do not believe this funcionality belongs in `hie-plugin-api`, it will only be used when a plugin is integrated, so belongs either in `Haskell.Ide.Engine.Dispatcher` or a `Haskell.Ide.Engine.Utils` file in the same place.

src/Haskell/Ide/Engine/Monad.hs
```diff
@@ -46,6 +47,9 @@ data IdeState = IdeState
 
 runIdeM :: IdeState -> IdeM a -> IO a
 runIdeM initState f = do
+    case validatePlugins (idePlugins initState) of
```
alanz: I think the plugin validation should probably happen in `MainHie`, before it starts spawning anything off.

So if there is a plugin error, it does not even start. And one day we will be asked to add a flag to skip the test, or run anyway even if there are errors. But I would say terminate on failure for now.

alanz:
Looks good. Just a few housekeeping issues.

tobiasgwaaler:
Good points, @alanz, I've made the changes :)

alanz:
Thanks, looks great

PR #120, opened 2015-12-03
New logger

### LINKED ISSUE #112: Replace logging package
At the moment the logging has issues with not flushing if the exe is killed, and brings in `pcre` which complicated windows/mac installs.

Replace it with something else, or roll our own.

alanz:
Updated logging to remove PCRE dependency

It now makes use of a blend of Control.Monad.Logger and Control.Logging,
in that it uses the monad logger log formatting functions, which
includes using TH for a location, and the Control.Logging IORef process,
so logging is available in any thread that has access to IO.

At the moment it has the same flushing behaviour as before.

Closes #112 

cc @bergey

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -52,14 +52,14 @@ library
                      , pipes-attoparsec >= 0.5
                      , pipes-bytestring
                      , pipes-parse
-                     , stm
                      , servant-server
+                     , stm
                      , text
                      , time
                      , transformers
+                     , vinyl >= 0.5 && < 0.6
                      , wai
                      , warp
-                     , vinyl >= 0.5 && < 0.6
   ghc-options:         -Wall
   default-language:    Haskell2010
 
@@ -78,9 +78,9 @@ executable hie
                      , gitrev >= 1.1
                      , haskell-ide-engine
                      , hie-example-plugin2
-                     , hie-plugin-api
                      , hie-ghc-mod
                      , hie-hare
+                     , hie-plugin-api
                      , monad-logger
                      , optparse-applicative
                      , optparse-simple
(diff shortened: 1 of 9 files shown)


haskell-ide-engine.cabal
```diff
@@ -105,12 +108,14 @@ test-suite haskell-ide-test
...
                      , unordered-containers
                      , directory
+                     , fast-logger
                      , haskell-ide-engine
                      , hie-ghc-mod
                      , hie-hare
                      , hie-plugin-api
                      , hspec
-                     , logging
+                     , fast-logger
```
bergey: duplicate of line 111?

alanz: Oops

src/Haskell/Ide/Engine/MonadFunctions.hs
```diff
@@ -1,21 +1,163 @@
...
+
+logTimeFormat :: IORef String
+{-# NOINLINE logTimeFormat #-}
+logTimeFormat = unsafePerformIO $ newIORef "%Y %b-%d %H:%M:%S%Q"
+
+-- | Set the format used for log timestamps.
+setLogTimeFormat :: String -> IO ()
+setLogTimeFormat = atomicWriteIORef logTimeFormat
+
+-- | This function, or 'withStderrLogging', must be wrapped around whatever
```
mgsloan: While perhaps it's clear from the type, it might be good to mention in the docs that these modify global state.  So, if you `Control.Concurrent.Async.race (withStdoutLogging (logm "a")) (withFileLogging "log" (logm "b"))`, it's not deterministic which messages go to which logs.

alanz: Makes sense, because you should only ever run one of the `withXXXLogging` functions. Perhaps that should be documented to.

mgsloan:
LGTM!

I once did something similar, but it also associated the log info with threads (how to log, log level, etc).  This lets you selectively log different portions of the process.  There's certainly a case to be made that [thread-local storage](https://wiki.haskell.org/Thread-local_storage) is inadvisable, but it certainly would be cool for things like this.

Lets keep it simple till we really find the need to do something more specialized.

PR #119, opened 2015-12-03
(emacs) show log of process input / output

mgsloan:
@gracjan Please review.  I'm rather new to elisp.

This makes the `*hie*` buffer accumulate a log of `<- input` and `-> output`, like:

```
-> {"cmd":"base:commandDetail","params":{"command":{"text":"type"},"plugin":{"text":"ghcmod"}}}
<- {"contexts":["point"],"return_type":"TypeInfo","plugin_name":"ghcmod","name":"type","additional_params":[],"ui_description":"Get the type of the expression under (LINE,COL)","file_extensions":[".hs",".lhs"]}
-> {"cmd":"ghcmod:type","params":{"file":{"file":"\/home\/mgsloan\/fpco\/stack\/src\/Stack\/BuildPlan.hs"},"start_pos":{"line":272,"col":79},"end_pos":{"line":272,"col":79}}}
[...]

PR #123, opened 2015-12-05
(elisp) Add log of hie process input / output

mgsloan:
Different approach than #119

### DIFF
--- a/elisp/hie.el
+++ b/elisp/hie.el
@@ -25,8 +25,11 @@
 (defvar hie-process nil
   "Variable holding current Haskell IDE Engine process")
 
-(defvar hie-buffer nil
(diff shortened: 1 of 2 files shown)


mgsloan:
@cocreature / @gracjan Please review.

I imagine in the future we might want to put a space in front of the name `*hie-process*`, to hide the buffer, since that buffer isn't very interesting to look at.

cocreature:
I don’t think a space is necessary. There are a ton of packages that use `*whatever*` buffer names for logs, so it’s completely fine to use that here. I’ll take a look at the actual PR sometime later today.

alanz:
This is great, just tried it now

PR #130, opened 2015-12-11
Ghc mod session

### LINKED ISSUE #127: Sort out ghc-mod session management
ghc-mod has a caching strategy for GHC sessions.

At the moment, the way the `IdeM` is implemented in HIE does not take advantage of this.

Update the monad/ghc-mod to do this.

alanz:
Simplify IdeM so that ghc-mod session caching can be used.

This provides a significant performance boost.

Closes #127

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,17 +27,15 @@ install:
   # Bring in GHC if not already present
   - ./travis_long stack +RTS -N2 -RTS setup
   # Build the dependencies only.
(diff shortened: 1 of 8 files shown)


DanielG:
Looks good to me though I still recommend we move the whole `setCurrentDirectory =<< root` hakery in the dispatcher at some point so it only has to run once on startup.

alanz:
Well, I would rather get rid of it via https://github.com/kazu-yamamoto/ghc-mod/pull/690#issuecomment-163871713

Issue #131, opened 2015-12-11
apply-refact plugin


alanz:
Create a plugin to expose the commands from [apply-refact](https://github.com/mpickering/apply-refact), basing it on [hlint-refactor-mode](https://github.com/mpickering/hlint-refactor-mode)

JPMoresmau:
Closed via #133

PR #133, opened 2015-12-14
Apply refact plugin

alanz:


alanz:
Closes #131

stack.yaml
```diff
@@ -13,6 +14,10 @@ packages:
     commit: b9bd4ebf77b22d2d9061d647d7799ddcc7c51228
     # commit: bff86be69f556f80a8dcd9dd42774ab77cb00eba
   extra-dep: true
+# - location:
+#     git: https://github.com/alanz/hlint.git
+#     commit: 6fb531d5ba32a4ba0e1a1c189e5f81ac7c43803f
+#   extra-dep: true
```
JPMoresmau: Are these lines still needed for reference?

JPMoresmau:
I see the value of applying hlint suggestions via refact, but what's the pipeline there? The ghc-mod lint command we expose return a Text, since ghc-mod applies show to the list of hlint suggestions. Since we seem to call hlint ourselves in that plugin, shouldn't we

- remove the ghc-mod lint command
- provide a lint command that returns a properly typed result, with proper location information?

stack.yaml
alanz: Nope, will delete them

alanz:
Its a bit of a strange pipeline, but models what happens outside of hie.

In the normal case, ghc-mod or flyecheck or whatever runs hlint and displays the warnings.

Then [hlint-refactor-mode](https://github.com/mpickering/hlint-refactor-mode) which actually knows nothing about the displayed hints, invokes `hlint` itself, in an integrated pipeline mode which calls the `refactor` command from `apply-refact`.

So the two could potentially be collapsed, but have no specific requirement that this is so. Essentially the one runs hlint to display the changes, the other runs it in a different mode to apply one or more of them

hie-apply-refact/hie-apply-refact.cabal
```diff
@@ -0,0 +1,30 @@
...
+category:            Web
+build-type:          Simple
+-- extra-source-files:
+cabal-version:       >=1.10
+
+library
+  exposed-modules:     Haskell.Ide.ApplyRefactPlugin
+  build-depends:       base >= 4.7 && < 5
+                     , aeson
+                     , apply-refact
```
JPMoresmau: I thought apply-refact was integrated into hlint, and we used it via hlint invocation, do we need that dependency? I don't see any usage of apply-refact modules.

alanz: See the end of the log in this build for what happens without this
dependency: https://travis-ci.org/alanz/haskell-ide-engine/builds/96739271

hlint does manage the application of the hints, but does so by calling the
`refactor` executable from `apply-refact`.

JPMoresmau: OK!

PR #135, opened 2015-12-15
Ghc mod find

JPMoresmau:
Implement the find command since it was a placeholder. The issue was that the find command in ghc-mod launches another copy of the executable. The source doesn't say why, and if dumping the symbol file in the same session should be avoided. It works without it: I basically just copied the code from findSymbol. The return type is a `ModuleList`, a specific type added to SemanticTypes.
A side effect is that CtxProject does require the project directory, since we need it for ghc-mod. So there's a `dir` parameter for CtxProject commands. Its type is still `PtFile`, not sure it's worth creating a different `PtDir` parameter type.


DanielG:
We exec() the dump symbol table function because of a memory leak in GHC FYI.

alanz:
@DanielG does the leak occur in all GHC versions?

PR #141, opened 2015-12-17
Hare context

JPMoresmau:
As discussed via IRC, hare needs to be in the right folder. With this change, running a refactor command from Leksah generates a right answer (a diff). Progress!

### DIFF
--- a/hie-hare/Haskell/Ide/HaRePlugin.hs
+++ b/hie-hare/Haskell/Ide/HaRePlugin.hs
@@ -2,17 +2,20 @@
 {-# LANGUAGE GADTs #-}
 module Haskell.Ide.HaRePlugin where
 
-import           Control.Exception
 import           Control.Monad.IO.Class
 import qualified Data.Text as T
 import           Data.Vinyl
 import           Haskell.Ide.Engine.PluginDescriptor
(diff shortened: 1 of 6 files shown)


hie-hare/Haskell/Ide/HaRePlugin.hs
```diff
@@ -180,10 +147,28 @@ makeRefactorResult changedFiles = do
...
-  return res
-  where
-    handler:: SomeException -> IO (Either String t)
-    handler e = return (Left (show e))
+
+runHareCommand :: T.Text -- ^ The file name we'll operate on
+                 -> String -- ^ command name for log
+                 -> (RefactSettings -> GM.Options -> FilePath -> IO [FilePath])
+                 -> IdeM (IdeResponse RefactorResult)
+runHareCommand fp name cmd = do
```
alanz: I wonder if we shouldn't pull this into `PluginUtils` or `MonadFunctions`, to run a `GhcModT` action in an appropriate directory.

JPMoresmau: Yes, I've started putting stuff in there (fileInfo, catchException).

PR #143, opened 2015-12-18
Initial example async process plugin

alanz:
I am not sure if this should be in master, but I am getting tired of rebasing it.

Comments?


cocreature:
Looks reasonable, why do you think it should not be in master?

PR #144, opened 2015-12-20
Base split

JPMoresmau:
Puts the plugin and semantic types into their own package, so that a Haskell client can easily reuse them with minimal dependencies. Tested with Leksah, which allowed me to find a JSON deserialization bug (fixed + quickcheck tested).
Note: it seems PluginTypes does NOT compile with aeson 0.10. To investigate and possibly report as a Aeson bug?

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -115,6 +115,7 @@ test-suite haskell-ide-test
                      , fast-logger
                      , haskell-ide-engine
                      , hie-apply-refact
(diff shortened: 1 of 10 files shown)


alanz:
I know there is an API change to aeson 0.10, and at the time I looked the one in stackage nightly was the earlier one, so we stuck with that at the time.

It may have changed by now, in which case moving forward may be an option

cocreature:
Thanks

PR #145, opened 2015-12-22
Support more commands

cocreature:
This is a bunch of different things but they built upon each other so splitting it up doesn’t really gain us anything.

In particular the generated functions now have docstrings, parameters and ask for the required paramaters when called interactively. Apart from that I also added support for the refactor type returned by hare-rename (and probably also a lot of the other hare commands, that was just what I tested). Currently I just overwrite the old buffer content. I know at least @alanz would like to have an ediff window so we should add an option like `hie-refactor-confirm` which when enabled causes an ediff window to pop up. There’s nothing particularly difficult about adding that, I simply haven’t gotten around to it yet. Imho that can be done in a separate commit as in the disucssions it seemed like most people just want the buffer to be overwritten.

### DIFF
--- a/elisp/hie.el
+++ b/elisp/hie.el
@@ -159,7 +159,9 @@ association lists and count on HIE to use default values there."
 (defun hie-handle-message (json)
   (-if-let ((&alist 'type_info type-info) json)
       (hie-handle-type-info type-info)
-    (message (format "%s" json))))
+    (-if-let ((&alist 'refactor refactor) json)
+        (message "refactoring")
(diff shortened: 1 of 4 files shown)


alanz:
@cocreature it may be enough to just show the patch in a buffer, as a preview. even after the fact, given undo capability

cocreature:
@alanz The patch is already written to the `*hie-refactor*` buffer. So if you just want to inspect it it’s easy.

alanz:
ok, sounds good.

PR #147, opened 2015-12-22
Add integration test for hie-hare-rename

cocreature:


cocreature:
If anybody has an idea why this fails on travis I would be very happy. I don’t really know where to look.

bergey:
I get the same error locally if I reduce the 2-second delay.  (The one after `hie-hare-rename`).  Maybe this pause needs to be longer on travis?

cocreature:
Thanks @bergey that was it

alanz:
Nice to see this

PR #148, opened 2015-12-23
Move to a plugin/command url scheme

cocreature:
This changes the format of url requests from
```
curl -v -H "Content-Type: application/json" -X POST -d '{"cmd": "lint", "params":{"file":{"file":"/home/moritz/code/haskell/playground/src/Main.hs"}}}' http://localhost:8001/req/ghcmod
```
to
```
curl -v -H "Content-Type: application/json" -X POST -d '{"file":{"file":"/home/moritz/code/haskell/playground/src/Main.hs"}}' http://localhost:8001/req/ghcmod/lint
```
Imho this is a more intuitive. The double `file` parameter comes from the fact that the first is the parameter name and the second is the type so that’s nothing specific to the servant protocol.

### DIFF
--- a/docs/Hacking.md
+++ b/docs/Hacking.md
@@ -57,13 +57,13 @@ This makes use of servant, and exposes the following API
 So it can be accesed by
 
 ```
-curl -v http://localhost:8081/req/base -X POST -H Content-Type:application/json -d '{"params":{},"cmd":"version"}'
+curl -v -H "Content-Type: application/json" -X POST -d '{"command":{"text": "version"},"plugin":{"text": "base"}}' http://localhost:8001/req/base/commandDetail
 ````
 
 This returns
 
 ```json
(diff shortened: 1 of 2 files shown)


alanz:
Looks like a good idea

alanz:
We need to come up with whatever scheme works best with swagger2, for documentation and tooling

cocreature:
My understanding is that swagger is a standardized way to specify your api but doesn’t require it to be in a specific way, however I’m happy to be corrected here.

Also swagger or at least servant-swagger doesn’t seem particularly useful for the way our servant api currently looks since it only contains one route and doesn’t know about the plugins or the commands that exists so anything that tries to generate documentation from the servant type can’t document plugins and commands. Again I’m happy to be corrected here.

For the json stdio I’m currently writing a documentation generator that outputs rst, but I’m happy to later merge that with the swagger docs if they turn out to be useful or to merge our http docs into that.

PR #149, opened 2015-12-28
Documentation generator

cocreature:
This is the start of a documentation generator for our json stdio protocol, but we can add any other protocols there as well.

The documentation is auto generated by travis and pushed to github pages (currently to https://cocreature.github.io/haskell-ide-engine/). I copied the steps for that pretty much exactly from this [gist](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd). I am pretty sure something will break when we try to do this on this repo, but it should be a matter of changing some user names from `cocreature` to `haskell` and maybe regenerating the key.

It’s still extremely basic, but it’s better than nothing and we can then build on what we have instead of trying to get it right in a single PR.

EDIT:
I don’t actually know sphinx so I might have done things terribly wrong.


hie-docs-generator/src/Options.hs
```diff
@@ -0,0 +1,32 @@
...
+  Config {plugins :: Plugins
+         ,prefix :: FilePath}
+
+config :: Parser Config
+config =
+  Config pluginList <$>
+  strOption (long "prefix" <> metavar "PREFIX" <>
+             help "directory where the docs are written to (defaults to current directory)")
+
+pluginList :: Plugins
```
alanz: We should try to make this common with the one in the app/Main.hs, else we have to change it in two places

cocreature: Good point, I’ll create a `hie-config` package containing it to avoid having the `hie` library depend on all plugins.

cocreature: For documentation purposes I’ll repeat what I said in IRC regarding this issue:

Splitting up the plugins into a separate `hie-config` package results in a circular dependency since it needs to depend on `haskell-ide-engine` for the base plugin. On the other hand `haskell-ide-engine` (the executable, but that doesn’t seem to make a difference) needs to depend on `hie-config`. The base plugin can’t be split of because it depends on internals from `haskell-ide-engine` for the version command. Since we’ll rework the config for #25 anyway, it’s probably not worth putting too much work into this right now.

PR #153, opened 2016-01-02
Just some textual improvements

cies:
I'd like to do some work on this project, but while delving into it I found I'm probably not up to the level of Haskell experience to be of any use within the limited time I have available. I did notice some thing in on the textual side that I could fix, so I did.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015, TBD
+Copyright (c) 2015-2016, TBD
 
 All rights reserved.
 
--- a/README.md
+++ b/README.md
@@ -11,39 +11,38 @@
 [license]: https://github.com/haskell/haskell-ide-engine/blob/master/LICENSE
 
 
-This project aims to be the __universal IDE interface__ to __all haskell tools__, proving a __full-featured and easy to query ide-backend__.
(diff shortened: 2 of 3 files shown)


cocreature:
Thanks it would be great if you could apply my comment, than I’ll happily merge it.

Regarding your wish to contribute a bit more:

- We have a [number of issues](https://github.com/haskell/haskell-ide-engine/issues?q=is%3Aopen+is%3Aissue+label%3A%22Well+Defined+Task+-+Awaiting+PR%22) tagged as being easy to start with, but to be honest, looking through them write now, we definitely need to do a better job there.
- We’re happy to lend you a hand if you get stuck (although a minimal level of haskell experience is probably required).
- If you don’t know haskell but know the language, the tooling for your editor of choice is written in, that’s a great way to help the project.

cocreature:
The build failure seems to be unrelated, so don’t worry about that. It’s most likely caused by an update to `Babel` which is required by `Sphinx`

cies:
Sorry for the typo. Fixed it.

Thanks for reaching out to lend a hand, but I think the time I need to get to each of these issues fixed is big. As I'm feverishly interested in having a stable AND powerful editor experience when using Haskell, I have done quite a bit of testing as an early adopter of `stack-ide`. Will surely do the same with HIE at some point.

cocreature:
Thanks again.

PR #162, opened 2016-01-05
Bump deps and adapt to new apis

cocreature:
There are two minor changes here:
- I adapted the ApplyRefactPlugin to the new hlint api.
- I slightly changed the encoding of `IdeError`. Previously we had a Maybe Value which causes issues with the new aeson (I haven’t looked into why we didn’t run into it with the old version) since both `Nothing` and `Just Null` are encoded to `Null` so we no longer have an isomorphism. Since we already have a `Null` value in `Value` and I don’t think we need the distinction here, this is now just a `Value`.

### DIFF
--- a/hie-apply-refact/Haskell/Ide/ApplyRefactPlugin.hs
+++ b/hie-apply-refact/Haskell/Ide/ApplyRefactPlugin.hs
@@ -2,7 +2,9 @@
 {-# LANGUAGE GADTs #-}
 module Haskell.Ide.ApplyRefactPlugin where
 
+import           Control.Arrow
 import           Control.Monad.IO.Class
+import           Data.Aeson
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import           Data.Vinyl
@@ -10,7 +12,6 @@ import           Haskell.Ide.Engine.MonadFunctions
 import           Haskell.Ide.Engine.PluginDescriptor
 import           Haskell.Ide.Engine.PluginUtils
 import           Haskell.Ide.Engine.SemanticTypes
-import           Language.Haskell.HLint
 import           Language.Haskell.HLint3
 import           Refact.Apply
 import qualified Refact.Types as R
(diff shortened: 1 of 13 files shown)


hie-apply-refact/Haskell/Ide/ApplyRefactPlugin.hs
```diff
@@ -77,39 +78,20 @@ applyAllCmd = CmdSync $ \_ctxs req -> do
...
-      opts = case mpos of
-        Nothing -> optsf
-        Just (r,c) -> optsf ++ " --pos " ++ show r ++ "," ++ show c
-    -- let hlintOpts = [file, "--quiet", "--refactor", "--refactor-options=" ++ opts ]
-    let hlintOpts = [file, "--quiet" ]
-    logm $ "applyHint=" ++ show hlintOpts
-    res <- catchException $ hlint hlintOpts
-    logm $ "applyHint:res=" ++ show res
-    -- res <- hlint hlintOpts
+    (flags,classify,hint) <- autoSettings
```
alanz: You should rather use argSettings, so we can eventually pass through any extra flags the user has set.

See https://github.com/alanz/haskell-ide-engine/commit/97ccf826cb9f551b0e1b82ebbcc67a9c7189a644#diff-7b1a90d13e6fd0644f04b15c6c20befeR106, from my branch updating apply-refact

cocreature: Thanks, updated. Sorry for duplicating your efforts, I didn’t see that you were already working on it.

stack.yaml
```diff
@@ -11,21 +9,13 @@ packages:
...
-    git: https://github.com/alanz/apply-refact.git
-    commit: ba98a2902e5333519e60d38803f30f82c44eaffc
+    git: https://github.com/mpickering/apply-refact.git
+    commit: 402458652844c1a0f42b15123e0ceff761919415
   extra-dep: true
 extra-deps:
- - HaRe-0.8.2.1
- - rosezipper-0.2
- - syz-0.2.0.0
+- hlint-1.9.26
```
alanz: I am pretty sure this version of hlint is in the current nightly, can be removed from this file

cocreature: I thought I checked but apparently I haven’t. Just pushed an update.

Issue #159, opened 2016-01-05
Fix warnings


cocreature:
We seem to have accumulated some warnings, most of which are probably completely unnecessary (like unnecessary imports), so it would be great if someone could go through them and fix them.

This sounds like a nice issue to get started with hie development.

One thing that’s annoying about this is that we can’t just use `--pedantic` since stack also enables the flags for extra-deps, but that is supposed to be fixed.

glittershark:
Any interest in setting -Werr on CI so this doesn't happen again?

cocreature:
Theoretically: absolutely, I think we even had that some point.

Particallly: Not an potion atm because stack will also set `-Werr` for all extra-deps which causes the build to fail since some of them don’t compile with `-Werr`. We should be able to get rid of the extra-deps rather soon and the stack guys also said they’re going to change this so when either of those happens, we should definitely do that again.

PR #165, opened 2016-01-09
Fix warnings

rasendubi:
PR for https://github.com/haskell/haskell-ide-engine/issues/159.

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE TemplateHaskell #-}
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -92,7 +92,7 @@ executable hie
                      , time
                      , transformers
                      , vinyl
-  ghc-options:         -threaded -rtsopts -with-rtsopts=-N -Wall
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N -Wall -Werror
   default-language:    Haskell2010
 
 test-suite haskell-ide-test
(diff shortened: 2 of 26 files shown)


cocreature:
Thanks! Could you please put the partial type signatures pack in and enable in those files `{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}`, the types don’t help here.

Also I’ll leave it to @alanz to decide whether we should enable `-Werror`. We should definitely remove it when we release but until then I like the idea of enabling it.

glittershark:
could have a cabal flag that's off-by-default to have it only enable on travis

cocreature:
A cabal flag sounds like a good idea.

DanielG:
+1 for having a default off flag for -Werror that's on on travis. It's that's very annoying during development.

mgsloan:
Agreed, and a cabal flag is not necessary.  This can be done by invoking `stack build` with `--pedantic`.

rasendubi:
@mgsloan `--pedantic` passes `-Werror` to all underlying libraries which eventually fails to build because of warnings in ghc-mod.

cocreature:
Maybe we just need to annoy @mgsloan enough so that he fixes that for us :)

alanz:
+1

mgsloan:
Ah, it isn't that it gets passed to underlying libraries in general.  Looks like there's a bug with packages that are marked `extra-dep: True`, where `ghc-options` are erroneously applied to them.

I didn't know about this bug, and I don't think the stack issue tracker did either.  I've opened https://github.com/commercialhaskell/stack/issues/1631

rasendubi:
@mgsloan see https://github.com/commercialhaskell/stack/issues/1292

mgsloan:
Gotcha.  I knew the issue sounded familiar!  Didn't find it since I was searching for ghc-options rather than pedantic

cocreature:
So what do we do here? It seems like the options are either adding a flag or removing the `-Werror`. I don’t really care what we decide to do, but I don’t want this PR to sit around waiting just because of that reason.

glittershark:
I'm ++ for the flag

alanz:
Perhaps remove -Werror, and then build with --pedantic once the bug is
fixed there

cocreature:
Since we are now back to having two options and two people each favoring one of those, let’s just go with a flag for now so we can merge this ( @rasendubi please add that). Once `--pedantic` is actually usable again, we can kill it.

alanz:
ok

cocreature:
Closing in favor of #171

Issue #161, opened 2016-01-05
HIE and current working directory


alanz:
This is a summary of a discussion with @DanielG on #ghc-mod last night

1. In general, given that GHC is non-reentrant, and we are running it in the same process as HIE via the GHC API, it will not be possible/easy to have more than one GHC session at the same time.

2. Certain plugins make use of the GHC API to typecheck and more, so the GHC version in HIE must match the GHC version for the project. So we are likely to end up with hie-7.10, hie-8.0, etc as executables

So the conclusion is that if an IDE is going to support multiple haskell projects simultaneously, it wiil need to run multiple instances of HIE, one per project.

An eventual future component may sit in front of the running HIE processes, to manage the communications from the IDE to them.

-----------

Furthermore, the monad we are using wraps a ghc-mod session, and this is started in the main programme, and launches the dispatcher. Up to 5.4 ghc-mod would throw an exception if it is not in the root directory of a file being loaded into a ghc session, and we used this to (transparently) change to that directory, run the command, and then change back to what it was, $HOME by convention.

The initial entry into `runGhcModT` sets up the environment, identifies the project root directory, and ensures that the cabal information is cached ready for use. This should happen once per project, when the ghc-mod / GHC session is initiated. Think of this as firing up ghci.

So, given than we are likely to run a separate HIE instance per project, it would make more sense to initiate it in the project directory, so that the initial ghc-mod session startup actually sets up the right thing, which is then cached for speedy operation of commands sent from the IDE.

We will have to make some change to manage this for the imminent release of ghc-mod 5.5, the question is whether we change to specifying a project root on startup (which ghc-mod can work out from any project file), or whether we try to panel beat what we have to work somehow, with all the pain of trying to get the ghc-mod caching/session management to work properly.

cocreature:
I’m fine with just using different instances of hie for separate projects. THere is not a lot to be gained by using the same instance since caching and similar performance improvements are also limited to a single project.

What I am not quite sure about is what you mean by "specifying a project root on startup". If ghc-mod can figure that out do we really need the ide to supply that? I don’t like the idea of forcing that burden on the ide. If you just mean that ghc-mod figures that out on startup, then I’m all for it.

alanz:
That is the part we need to figure out. The question is at what point does an IDE start hie? If it is in the context of a file in a project of interest, then it gets launched at that point and we are fine.

[...]

PR #168, opened 2016-01-11
Add ide-backend plugin

cocreature:
There are a few things here:
* It doesn’t work properly if you call it on 2 files from different projects. Given that #161 will limit hie to only deal with one project at a time anyway, I would prefer to leave it like this and instead spend the time on #161.
* I only added the type command, there is nothing stopping as from adding more commands, I just wanted to start somewhere.
* A lot of this stuff is very similar to the things in `ExamplePluginAsync`. It would probably make sense to create an `hie-plugin-utils` package containing things which are not directly relevant to `hie` itself but can be reused by plugin authors.

Also I just realized that I forgot the tests, I’ll hopefully add those later this evening.

Pinging @mgsloan since he seemed to be interested in an `ide-backend` plugin.

### DIFF
--- a/hie-ide-backend/Haskell/Ide/IdeBackend.hs
+++ b/hie-ide-backend/Haskell/Ide/IdeBackend.hs
@@ -1,0 +1,168 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+module Haskell.Ide.IdeBackend where
+
+import           Control.Concurrent
+import           Control.Concurrent.STM
+import           Control.Monad.IO.Class
+import           Data.Aeson
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
+import           Haskell.Ide.Engine.ExtensibleState
+import           Haskell.Ide.Engine.MonadFunctions
+import           Haskell.Ide.Engine.PluginDescriptor
+import           Haskell.Ide.Engine.PluginUtils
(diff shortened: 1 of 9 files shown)


glittershark:
Is ide-backend really the right name for a plugin that gets types of identifiers?

cocreature:
@glittershark Well it’s about as right as `ghc-mod` is for the `ghc-mod` plugin.

cocreature:
@glittershark Ah if it was confusing for you, [ide-backend](https://github.com/fpco/ide-backend) is a separate project upon which this plugin builds. So it’s not named that way because it aims to be a complete `ide-backend` but for the same reason the ghc-mod plugin is called ghc-mod.

glittershark:
oh! I understand now.

cocreature:
Added the tests.

cocreature:
Not sure why travis is complaining, the tests pass locally. I’ll take a look tomorrow.

cocreature:
Ah the error is because `ide-backend-server` is not found on travis. that should be an easy fix.

cocreature:
Travis is happy now.

hie-ide-backend/Haskell/Ide/IdeBackend.hs
```diff
@@ -0,0 +1,167 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+module Haskell.Ide.IdeBackend where
```
JPMoresmau: Some documentation on all the types and methods would be good...

cocreature: Good point, I’m terrible when it comes to documentation. I’ll add it later today or tomorrow.

cocreature:
Docs added.

PR #170, opened 2016-01-12
Ghc-dump-tree plugin

JPMoresmau:
Little plugin around the newly created ghc-dump-tree library to get the full AST

### DIFF
--- a/hie-base/Haskell/Ide/Engine/SemanticTypes.hs
+++ b/hie-base/Haskell/Ide/Engine/SemanticTypes.hs
@@ -62,6 +62,18 @@ data ModuleList = ModuleList {
   } deriving (Show,Read,Eq,Ord,Generic)
 
 -- ---------------------------------------------------------------------
+
+-- | GHC AST
+data AST = AST {
+    astModule :: T.Text
+  , astParsed      :: Value
+  , astRenamed     :: Value
+  , astTypechecked :: Value
+  , astExports     :: Value
+  } deriving (Eq,Show)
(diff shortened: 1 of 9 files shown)


hie-base/Haskell/Ide/Engine/SemanticTypes.hs
```diff
@@ -62,6 +62,18 @@ data ModuleList = ModuleList {
   } deriving (Show,Read,Eq,Ord,Generic)
 
 -- ---------------------------------------------------------------------
+
+-- | GHC AST
+data AST = AST {
+    astModule :: T.Text
+  , astParsed      :: Value
+  , astRenamed     :: Value
```
alanz: Shouldn't renamed and onward be `Maybe Value` as they may not exist?

cocreature: `Maybe Value`s are annoying since both `Nothing` and `Just Null` get serialized to `Null`. Also the `Null` in the `Value` makes the Nothing redundant at least as long as we don’t need two different null cases (I don’t know if that’s the case here).

alanz: ok, and I looked more closely, we are receiving it as JSON, so presumably
the absence is already encoded.

cocreature:
This should be rebased on master and the pedantic cabal flags I added to the other packages should be added here too and enabled in `stack_test.yaml`.

PR #174, opened 2016-01-15
Recreate buffers if they have been killed

### LINKED ISSUE #121: (emacs) "Selecting deleted buffer" after *hie* buffer killed
If you use `M-x kill-buffer` on the `*hie*` buffer, subsequent commands error out with "Selecting deleted buffer".  This is due to a use of `with-current-buffer (hie-buffer)`.  Maybe defining a utility like `with-hie-buffer` would be good, which also ensures that the buffer is initialized.

ankhers:
This should recreate the `*hie-log*` and `*hie-process*` buffers if they are
used after they have been killed.

Fixes #121

### DIFF
--- a/elisp/hie.el
+++ b/elisp/hie.el
@@ -47,7 +47,7 @@
 
 (defun hie-process-filter (process input)
   (let ((prev-buffer (current-buffer)))
-    (with-current-buffer hie-process-buffer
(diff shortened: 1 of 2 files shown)


cocreature:
Thanks!

PR #176, opened 2016-01-15
ghc-mod 5.5

cocreature:
Now the cwd is only changed at startup, which means that every cabal
project requires it’s own hie instance.

This is not finished, but I thought that it might be easier to fix the remaining issues if we have something to look at.

In particular HaRe needs to expose functions that work in `GhcModT` instead of trying to setup it’s own ghc-mod session since that no longer works with ghc-mod 5.5.

Also the ide bindings need to be updated. They simply need to pass in the directory with the cabal file or any subdirectory of that (ghc-mod figures the root out in that case). They can either do that by starting `hie` with the correct cwd or by using `-r projectdir`. On the emacs front we can probably simply abuse haskell-mode’s session management to have one `hie` instance per project.

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -18,7 +18,6 @@ import           Control.Exception
 import           Control.Monad
 import           Control.Monad.Logger
 import           Control.Monad.STM
-import           Control.Monad.Trans.Maybe
 import qualified Data.Map as Map
 import           Data.Proxy
 import qualified Data.Text as T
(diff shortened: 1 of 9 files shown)


cocreature:
Updated, currently requires a yet to be merged [HaRe pr](https://github.com/alanz/HaRe/pull/43).
Also we need to disable ghc-mod’s find command for now. It requires launching a separate executable which can not be found when the executable is not ghc-mod itself. This will be fixed in an upcoming release so there’s no good reason to put in a temporary hack.

cocreature:
Travis is finally happy (after running out of memory multiple times, because for some reason the cache got cleaned), so as soon as somebody has time to look over it we can merge this.

PR #179, opened 2016-01-18
Remove ide-backend

cocreature:
There are two reasons for this:
1. It causes our travis builds to run out of memory.
2. ide-backend doesn’t really provide anything which we don’t already
have and it doesn’t fit in very well (e.g. it can’t reuse ghc sessions).


Makefile
```diff
@@ -5,22 +5,27 @@ ifeq (,$(shell which $(emacs) 2> /dev/null))
...
-build-haskell:
-	@STACK_YAML=stack_test.yaml stack build --test --no-run-tests
-
 .PHONY: test-haskell
 test-haskell:
 # stack build --test --pedantic
 # stack complains about deprecations in ghc-mod as an extra dep
 # build with -Werror enabled but run the tests without it because ide-backend picks it up
-	stack build --test
+	stack build --test --flag haskell-ide-engine:pedantic \
```
alanz: This looks encouraging

alanz: Maybe we should call pedantic.sh here instead, so the flags only get set in one place

cocreature: You’re probably right, but it requires some bash to not pass `+RTS -N1 -RTS` here. I’ll look the syntax for that up tomorrow and change it.

mgsloan:
Seems reasonable to me.  This code is still interesting to reference, thanks for writing it!  My plan for the ide-backend based information was to have it be independent of the ide-backend package itself (just based on the code).

cocreature:
@mgsloan I’m fine with moving some code over, in the case of type info I am not sure it’s very useful to have a type info plugin based on ide-backend and one on ghc-mod, but if there are useful differences, I’m happy to include something like that. It would be great if you could make a list of the things you’d like to have from ide-backend, then we can see how we provide them in a nice way.

PR #184, opened 2016-01-22
TCP transport

cocreature:
Most of the stuff we use for stdio transport can be completely reused for the tcp transport, so I moved that to a separate module.

Regarding tests, I am aware that I didn’t add tests. I took a look at the stdio tests and they don’t actually test the transport but only test the encoding and decoding (which is shared between stdio and tcp). Ofc we could build hie, launch it as a separate process and then actually test the tcp transport and the stdio transport, but I wasn’t sure if it’s worth the trouble.


alanz:
Sorry, been slow to look here. I have to admit I was wondering why it was not in master

PR #200, opened 2016-05-02
Provide REST API explorer via swagger

### LINKED ISSUE #19: Provide REST API explorer
The fledgeling HIE has a JSON API. This includes some introspection as to what plugins are available, and what commands they provide, together with their parameters.

Provide an application that can interact with the API, and present the results in a meaningful way.

i.e.

```
$ stack build
$ stack exec hie -- --repl -d
```

Then access `http://localhost:8001/` as in

```
curl -v http://localhost:8081/req/base -X POST -H Content-Type:application/json --data-binary '{"ideParams":{},"ideCommand":"version","ideContext":{"ctxEndPos":null,"ctxCabal":null,"ctxStartPos":null,"ctxFile":null}}'
````

Possibly consider using http://swagger.io, especially as there is now [servant-swagger](http://hackage.haskell.org/package/servant-swagger)

Or any other appropriate technology. It needs to be able to run locally though.

cc @kRITZCREEK

### PR DESCRIPTION
**Pull request by alanz**
Closes #19 

How to use it

```
stack build
stack exec hie -- -d --http
```

And then point your browser at http://localhost:8001

### CODE REVIEW

File: app/MainHie.hs [hunk 1]
```diff
@@ -136,12 +143,20 @@ run opts = do
       Just err -> error (pdeErrorMsg err)
       Nothing -> return ()
 
+    when (optDumpSwagger opts) $ do
+      -- putStrLn "dumping swagger definition"
+      let swagger = hieSwagger plugins
+      -- let swagger = hieSwagger2 (recProxy taggedPlugins)
+      -- putStrLn (show swagger)
+      putStrLn (B.unpack $  encode swagger)
```
cocreature: Is this guaranteed to be ascii? Otherwise Char8 is evil and this should be properly decoded (or just dump the bytestring).

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -175,8 +179,51 @@ type PluginName = T.Text
 data IdePlugins = IdePlugins
   { ipMap :: Map.Map PluginId [UntaggedCommandDescriptor]
   } deriving (Show,Eq,Generic)
+instance ToSchema IdePlugins
+
+-- ---------------------------------------------------------------------
+
+-- | A position in a source file
+-- type Pos = (Line,Col)
+data Pos = Pos { line::Line, col:: Col} deriving (Generic,Show,Read,Eq,Ord)
```
cocreature: The type synonym should be removed and the fields should be strict to prevent accidental space leaks.

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -175,8 +179,51 @@ type PluginName = T.Text
...
+
+
+instance Bounded Line where
+  minBound = 0
+  maxBound = 1000000
+
+instance Bounded Col where
+  minBound = 0
+  maxBound = 100000
+
```
cocreature: Are these bounds just arbitrary or are they actually restrictions imposed by GHC? In the first case it seems better to just use the int max for that. Also I think the lower bounds in GHC are 1 for both rows and columns?

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -175,8 +179,51 @@ type PluginName = T.Text
...
+toPos (l,c) = Pos (Line l) (Col c)
+
+newtype Line = Line Int deriving (Generic,Show,Eq,Read,Ord,Enum,Real)
+instance ToSchema Line
+
+newtype Col  = Col  Int deriving (Generic,Show,Eq,Read,Ord,Enum,Real)
+instance ToSchema Col;
 
-type Pos = (Int,Int)
+unLine :: Line -> Int
```
cocreature: You can use record syntax in the newtype to get those for free.

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -389,15 +426,53 @@ instance FromJSON (ParamVal 'PtFile) where
   parseJSON = withObject "file parameter object" $ \v -> ParamFile <$> v.: "file"
 
 instance FromJSON (ParamVal 'PtPos) where
```
cocreature: Using withObject here gives us better errors messages so it probably makes sense to leave it in or at least move it to the instance of Pos.

hie-base/Haskell/Ide/Engine/SemanticTypes.hs
```diff
@@ -2,11 +2,13 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# LANGUAGE StandaloneDeriving #-}
```
cocreature: It looks like you are not using this?

hie-base/Haskell/Ide/Engine/SemanticTypes.hs
```diff
@@ -17,18 +19,21 @@ import           Haskell.Ide.Engine.PluginTypes
 -- | Type Information, from the most precise to the most generic
 data TypeInfo = TypeInfo { results :: [TypeResult] }
   deriving (Show,Read,Eq,Ord,Generic)
+instance ToSchema TypeInfo
 
 -- | One type result from ghc-mod
 data TypeResult = TypeResult
-    { trStart :: (Int,Int) -- ^ start line/column
-    , trEnd   :: (Int,Int) -- ^ end line/column
+    { trStart :: Pos -- ^ start line/column
```
cocreature: While we’re at it these fields should also be strict.

hie-base/Haskell/Ide/Engine/SemanticTypes.hs
```diff
@@ -82,16 +89,13 @@ instance ValidResponse TypeInfo where
...
+             , "end"   .= toJSON e
+             , "type"  .= t
              ]
 
 instance FromJSON TypeResult where
-  parseJSON (Object v) = TypeResult
-    <$> (jsonToPos =<< (v .: "start"))
-    <*> (jsonToPos =<< (v .: "end"))
-    <*> v .: "type"
+  parseJSON (Object v) = TypeResult <$> v .: "start" <*> v .: "end" <*> v .: "type"
```
cocreature: A withObject here would probably make for better error messages.

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -175,8 +179,51 @@ type PluginName = T.Text
...
+-- type Pos = (Line,Col)
+data Pos = Pos { line::Line, col:: Col} deriving (Generic,Show,Read,Eq,Ord)
+-- NOTE: Pos needs to be defined in record syntac otherwise the generically
+--       derived swagger schema does not work properly
+instance ToSchema Pos
+
+unPos :: Pos -> (Int,Int)
+unPos (Pos (Line l) (Col c)) = (l,c)
+
+toPos :: (Int,Int) -> Pos
```
cocreature: Is there a reason why this is uncurried?

src/Haskell/Ide/Engine/Swagger.hs
```diff
@@ -0,0 +1,128 @@
...
+
+mkParamsSchema :: Referenced Schema -> Referenced Schema -> [ParamDescription] -> Schema
+mkParamsSchema fileSchemaRef textSchemaRef allParams = s
+  where
+    pList = map mkParam allParams
+
+    requiredParams = concatMap isRequired allParams
+
+    isRequired :: ParamDescription -> [Data.Swagger.ParamName]
+    isRequired pd = if pRequired pd == Required
```
cocreature: You can just pattern match on ParamDescription.

src/Haskell/Ide/Engine/Swagger.hs
```diff
@@ -0,0 +1,128 @@
...
+    -- TODO: do we actually need the command and plugin here too?
+    s = mempty
+      & type_ .~ SwaggerObject
+      & properties .~ HM.fromList [("params",Inline p)]
+
+-- ---------------------------------------------------------------------
+
+-- | local type declaration for swagger, so that the generated Schema matches
+-- what HIE is expecting
+data File = File { file::T.Text} deriving Generic
```
cocreature: Can this and Text be a newtypes?

src/Haskell/Ide/Engine/Transport/JsonHttp.hs
```diff
@@ -27,29 +27,45 @@ import qualified Data.Map as Map
...
-  parseJSON (Object v) = fmap (TaggedMap . Map.fromList) (parseTaggedMap (Proxy :: Proxy tags) v)
+  parseJSON (Object v) = do
+    {-
+       The initial version of the hie API expected parameters in the POST body of the form
+
+         {"command":{"text": "version"},"plugin":{"text": "base"}}
+
+       The swagger version can only accept one parameter in the body, so this must be wrapped as
+
+         { "params:
```
cocreature: There is a missing " at the end of params.

app/MainHie.hs
alanz: Using plain `ByteString.Put` now

hie-base/Haskell/Ide/Engine/PluginTypes.hs
alanz: The swagger spec spits out bounds, which look horrible for lines and cols from -2^63 to +2^63-1.

But I still need to modify the swagger schema gen to actually use the bounds.

The upper bounds are to cope with some of the auto-gen'd source out there, which gets pretty extreme.

hie-base/Haskell/Ide/Engine/SemanticTypes.hs
alanz: Ok, added withObject wherever I could. Hope I got them all. Learned something new in the process.

hie-base/Haskell/Ide/Engine/PluginTypes.hs
alanz: Yes, it was a quick and dirty fix to make updating the tests easier. 

So no, not really, I guess.

src/Haskell/Ide/Engine/Swagger.hs
alanz: They exist purely for the ToSchema instances, so the swagger gen works properly.  What would a newtype do for us that is desirable?

src/Haskell/Ide/Engine/Transport/JsonHttp.hs
alanz: Once more a victim of pressing ", getting 2, deleting one and they both disappear.

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -175,8 +179,45 @@ type PluginName = T.Text
...
+toPos (l,c) = Pos (Line l) (Col c)
+
+-- newtype Line = Line Int deriving (Generic,Show,Eq,Read,Ord,Enum,Real)
```
cocreature: Leftover comment that should be removed

src/Haskell/Ide/Engine/Swagger.hs
```diff
@@ -81,9 +81,9 @@ mkParamsSchema fileSchemaRef textSchemaRef allParams = s
     requiredParams = concatMap isRequired allParams
 
     isRequired :: ParamDescription -> [Data.Swagger.ParamName]
-    isRequired pd = if pRequired pd == Required
-                       then [pName pd]
-                       else []
+    isRequired pd = case pRequired pd of
```
cocreature: Sorry for being pedantic, but imho it looks prettier to just pattern match on the left of `=` instead of introducing a case statement.

src/Haskell/Ide/Engine/Swagger.hs
cocreature: 1. It has no runtime overhead. While this may not be a big deal getting better performance for “free” is nice.
2. imho it makes it clearer that it’s just a wrapper around something so it makes it easier to see the purpose.

cocreature:
Apart from my minor nitpicky comments this looks good! Feel free to merge when you’re ready.

cocreature:
One thing that I’m not too happy with is just copying all of swagger-ui in this repo (makes updating kinda weird). Is there some easy solution, like use it as a submodule/subtree or is it possible to just let users install it separately? I imagine this is mostly useful for developers so having to install it separately doesn’t seem that bad.

alanz:
Thanks for the feedback.

In terms of the swagger-ui stuff, I am in two minds. On the one hand, having it as a submodule could make sense, but there has already been tailoring (to hardcode the hie swagger.json url), and I can imagine in the future it would get more tailoring. We almost need to have a fork of the original as a submodule, but I am not sure how to set that up.

rvion:
@cocreature my 2 cents: I think it's more practical to have it in the repo directly, even there is minor drawbacks

The ui is generic and we may want to tweak it in the future (eg: remove the top input allowing to choose a different swagger file, etc). I use servant-swagger for some project and having the ui in the same folder was practical to tweak small bits of the js code so that the generated swagger spec is loaded by default, to implement custom versionning related menu bar, etc.

(also, this is not the best arguments, but submodules are not very pleasant to work with)

cocreature:
Small modifications are precisely what I’m trying to avoid. The reason for this is that updating swagger-ui from upstream (for bugfixes, new features, …) is non trivial if there are modifications. This is even more so if it’s not a separate repo where you can clearly see the commits and just do a merge but a copy of part of the code where you mostly need to manually figure out what’s changed and resolve conflicts. Leaving aside the aesthetic reasons such as not messing up diffs and our commit history with swagger-ui changes pulled from upstream, this simply means a lot of work and we are not exactly drowning in contributors right now.

That said I have never worked before with swagger-ui before so I can’t comment on how useful and important it is to customize the interface. Hardcoding the URL and removing the option to select a different swagger file don’t seem like issues worth the trouble. On the other hand if it is trivial to modify it the temptation will probably be too big and we do a lot of changes which are not really necessary but maybe nice to have and thereby cause the update problem described above.

If I’m the only one that wants to have it separately, I’m fine with having it in the repo, but these issues should at least be considered.

alanz:
@bitemyapp  @dmjio I considered using `servant-swagger` but ended up not doing so because we basically have an internal structure describing the API, which we then expose in `servant`. It made more sense to generate the `swagger.json` from the API definition directly.

bitemyapp:
@alanz oh fair enough, but could still poke about. Project might interest 'im

Issue #199, opened 2016-04-11
hie-docs-generator doesn't expand tilde on Linux


pakanek:
`hie-docs-generator --prefix="~/abc"` will create `./~` directory.

PR #207, opened 2016-06-08
Handle ~ path naively

psibi:
A naive attempt making commands like "hie-docs-generator
--prefix="~/hiedocs"" work as intended. Addresses #199


cocreature:
Thanks! I’m slightly confused that this is not in `filepath` or `directory` because it seems like something that is useful for pretty much anything that deals with filepaths, but I couldn’t find it so let’s merge this.

PR #210, opened 2016-09-04
GHC 8.0.1 and cabal new-build

alanz:


haskell-ide-engine.cabal
```diff
@@ -62,12 +62,13 @@ library
                      , directory
                      , either
                      , fast-logger
-                     , ghc >= 7.10.2 && < 7.11
-                     , ghc-mod >= 5.4
+                     , ghc >= 7.10.2 && < 8.3
```
cocreature: If we add upperbounds, shouldn’t this be < 8.2 or <= 8.0? It seems unlikely that 8.2 will not break anything.

hie-base/Haskell/Ide/Engine/PluginTypes.hs
```diff
@@ -280,8 +282,15 @@ instance Eq ParamValP where
  (ParamPosP  p) == (ParamPosP p') = p == p'
  _ == _ = False
 
+-- pattern ParamTextP :: () =>
```
cocreature: I think this should be removed?

haskell-ide-engine.cabal
alanz: True. I just put something in at the time, without thinking through the details, I guess.

hie-ghc-mod/Haskell/Ide/GhcModPlugin.hs
```diff
@@ -130,9 +138,23 @@ typeCmd = CmdSync $ \_ctxs req ->
...
     Right (ParamFile fileName :& ParamPos (Pos (Line l) (Col c)) :& RNil) -> do
-      fmap (toTypeInfo . T.lines . T.pack) <$> runGhcModCommand (GM.types (T.unpack fileName) l c)
+      fmap (toTypeInfo . T.lines . T.pack) <$> runGhcModCommand (GM.types False (T.unpack fileName) l c)
+#if __GLASGOW_HASKELL__ <= 710
     Right _ -> return $ IdeResponseError (IdeError InternalError
       "GhcModPlugin.typesCmd: ghc’s exhaustiveness checker is broken" Null)
+#endif
+
+{-
+TODO: pass the bool through as a param
```
cocreature: It might make more sense to move this TODO directly to the call of types to make it easier to understand.

stack-lts.yaml
```diff
@@ -0,0 +1,17 @@
+resolver: lts-6.13
+
+# NOTE: HIE does not intend to support GHC 7.10.x, the tests fail here.
+packages:
```
cocreature: If we don’t intend to support GHC 7.10, wouldn’t it make more sense to just bump the base lower bound and remove all the CPP to make the exhaustiveness checker? Building with GHC 7.10 while not being able to run using 7.10 doesn’t seem very useful.

hie-apply-refact/Haskell/Ide/ApplyRefactPlugin.hs
```diff
@@ -1,9 +1,10 @@
 {-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
 
-{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE CPP                   #-}
```
cocreature: Is there a reason why you changed the alignment of a lot of pragmas? I don’t have any strong feelings about it but it should be consistent. One advantage of the previous style is that adding a pragma doesn’t potentially require modifications to all other pragmas which can clutter diffs. Also the previous style conforms to what hindent does which seems to become a popular style so it might make sense to not diverge to far from it.

cocreature:
Thanks, looks good!

hie-base/Haskell/Ide/Engine/PluginTypes.hs
alanz: It should actually be uncommented. I was getting warnings on missing signatures, which led to https://ghc.haskell.org/trac/ghc/ticket/12484

stack-lts.yaml
alanz: Agree, just checking with some possible users, will then take action, probably via a subsequent PR

hie-apply-refact/Haskell/Ide/ApplyRefactPlugin.hs
alanz: I like having them sorted. Then my OCD bit kicked in and I aligned the closing.  I don't feel strongly about the endings, do about the sorting.  And I guess I need to start considering hindent.

PR #212, opened 2016-09-12
Grab bag

alanz:
- Target GHC 8.0.x only, getting rid of broken exhaustiveness checker tweaks
- deal with "projects" not having a cabal file, in the elisp
- ghc-tree returns the tree(s) for the requested file only

### DIFF
--- a/hie-ghc-tree/Haskell/Ide/GhcTreePlugin.hs
+++ b/hie-ghc-tree/Haskell/Ide/GhcTreePlugin.hs
@@ -11,11 +11,14 @@ module Haskell.Ide.GhcTreePlugin where
 
 import           Data.Aeson
 import qualified Data.Text as T
+-- import           Haskell.Ide.Engine.MonadFunctions (logm)
 import           Haskell.Ide.Engine.PluginDescriptor
 import           Haskell.Ide.Engine.PluginUtils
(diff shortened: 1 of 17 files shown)


hie-ghc-tree/Haskell/Ide/GhcTreePlugin.hs
```diff
@@ -11,12 +10,15 @@ module Haskell.Ide.GhcTreePlugin where
 
 import           Data.Aeson
 import qualified Data.Text as T
+-- import           Haskell.Ide.Engine.MonadFunctions (logm)
```
cocreature: You might want to remove this and also the commented line below calling `logm`.

stack.yaml
```diff
@@ -10,6 +10,10 @@ packages:
 - hie-ghc-tree
 - hie-hare
 - hie-docs-generator
+- location:
```
cocreature: Why is this necessary? I don’t see any obvious changes to the HaRe-plugin that would require this.

alanz: Its for the "no-cabal" project case, needing https://github.com/alanz/HaRe/commit/48102f8899a4cd837fa0891620d5264bbf09d7fe

PR #219, opened 2017-04-22
Microsoft Language Server Protocol support

alanz:
Add support for [LSP](https://github.com/Microsoft/language-server-protocol)

This is still early stage, but can provide diagnostics via `hlint`, and apply the hints via `codeAction` requests.

It is being developed against

- vscode using https://github.com/alanz/vscode-hie-server
- emacs using https://github.com/alanz/lsp-haskell


alanz:
Travis sucks, time limit. 

It passed here: https://travis-ci.org/alanz/haskell-ide-engine/builds/224722639
and here: https://circleci.com/gh/alanz/haskell-ide-engine/32

And I propose we move to circleci, which has configurable time limits

PR #221, opened 2017-04-30
Adapt to hskell-lsp VFS

alanz:
And accept `text` params in `workspace/executeCommand` parameters


alanz:
Travis has been removed, its timeout is a problem. Using circleci only now

PR #224, opened 2017-05-02
Added hoogle plugin

wz1000:


alanz:
Can you modify this to use the library approach instead, now that [Hoogle 5.0.11](https://hackage.haskell.org/package/hoogle-5.0.11/docs/Hoogle.html) is out which provides `defaultDatabaseLocation`?

You will probably have to add it as an extra dependency at the end of  `stack.yaml`

hie-hoogle/Haskell/Ide/HooglePlugin.hs
```diff
@@ -0,0 +1,91 @@
...
+        db <- defaultDatabaseLocation
+        dbExists <- doesFileExist db
+        if dbExists then do
+            res <- searchHoogle db expr
+            if null res then
+                return $ IdeResponseOk "No results found"
+            else do
+                let Target{..} = head res
+                    packageModule = unwords $ map fst $ catMaybes [targetPackage, targetModule]
+                return $ IdeResponseOk $ T.pack $ unlines $ [unHTML targetItem, packageModule, unHTML targetDocs]
```
alanz: Is this different from `disp`? Perhaps use `disp` here instead.

hie-hoogle/Haskell/Ide/HooglePlugin.hs
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+-- ---------------------------------------------------------------------
+
+infoCmd :: CommandFunc T.Text
+infoCmd = CmdSync $ \_ctxs req -> do
+  case getParams (IdText "expr" :& RNil) req of
+    Left err -> return err
+    Right (ParamText expr :& RNil) -> liftIO $ do
+        db <- defaultDatabaseLocation
```
alanz: Both commands do the same process until they get their results.

Maybe put that process in a helper function?

hie-hoogle/Haskell/Ide/HooglePlugin.hs
wz1000: It is different from `disp`.

PR #244, opened 2017-06-14
Hlint pragmas

alanz:
When an hlint pragma disables a given hint it is still returned by hlint, but with severity `Ignore`.  Strip these out so they are actually ignored, from the user perspective.


alanz:
cc @wz1000 Please take a look

PR #247, opened 2017-06-17
Brittany plugin and support for textDocument/formatting, textDocument/rangeFormatting

### LINKED ISSUE #215: Integrate brittany
See https://github.com/lspitzner/brittany/issues/5

wz1000:
Range formatting doesn't take into account pragmas yet, so often fails.

~~Also, we don't take into account client supplied Formatting Options yet.~~

Closes #215 and lspitzner/brittany#5

### DIFF
--- a/hie-brittany/Haskell/Ide/BrittanyPlugin.hs
+++ b/hie-brittany/Haskell/Ide/BrittanyPlugin.hs
@@ -1,0 +1,75 @@
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE LambdaCase            #-}
+module Haskell.Ide.BrittanyPlugin where
+
+import           Data.Aeson
+import           Data.Monoid
+import qualified Data.Text as T
+import           Data.Vinyl
+import           Haskell.Ide.Engine.PluginDescriptor
+import           Haskell.Ide.Engine.PluginUtils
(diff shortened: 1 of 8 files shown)


stack.yaml
```diff
@@ -34,18 +35,28 @@ packages:
...
 - pipes-aeson-0.4.1.8
 - pipes-network-0.6.4.1
+- butcher-1.1.0.0
+- czipwith-1.0.0.0
+- data-tree-print-0.1.0.0
+- deque-0.2
+- monad-memo-0.4.1
 # - hspec-jenkins-0.1.1
 flags:
   haskell-ide-engine:
```
alanz: please add a pedantic flag down here for `hie-brittany` too. And we need to move the CI ti run in pedantic too.

wz1000: ok

hie-brittany/Haskell/Ide/BrittanyPlugin.hs
```diff
@@ -0,0 +1,92 @@
...
+import qualified GHC.LanguageExtensions.Type as GHC
+import qualified Data.Text.Lazy as TextL
+import qualified GhcMod.Utils as GM
+import qualified Language.Haskell.LSP.TH.DataTypesJSON as J
+import           Control.Lens
+
+
+brittanyCmd :: TextDocumentIdentifier -> Maybe Range -> IdeM (IdeResponse [J.TextEdit])
+brittanyCmd tdi range =
+  pluginGetFile "genapplicative: " (tdi ^. J.uri) $ \file -> do
```
alanz: There are a number of strings for `pluginGetFile` that do not actually match. Not sure how we can test for them though. Perhaps better to leave them out and let the exception location guide us?

alanz:
Made a few small comments. Please also refer to closing https://github.com/haskell/haskell-ide-engine/issues/215 in the commit message.

hie-brittany/Haskell/Ide/BrittanyPlugin.hs
wz1000: I'll go over all of them once.

Issue #248, opened 2017-06-17
Brittany plugin not taking GHC extensions into account when formatting fragments of a file.


wz1000:
Brittany also doesn't read pragmas enabled in .cabal files.

We can have a complete solution to this by using ghc-mod to get the DynFlags and passing them on to Brittany.

A simpler solution is enabling common, safe syntactic extensions by default. See lspitzner/brittany/issues/32

PR #256, opened 2017-07-05
brittany plugin: Enable syntactic extensions by default

lspitzner:
compiles, but otherwise untested (my editor doesn't support lsp sufficiently well yet, i fear).

This is not necessarily supposed to be a permanent solution; it would generally be preferable to have the extensions used be configurable by the user or to have them determined from the module/project environment (cabal file etc.).


wz1000:
Thanks.

Closes #248

PR #257, opened 2017-07-05
Add support for DocumentHighlight; improve FindDef, HoverReq; cleanup.

wz1000:


hie-hare/Haskell/Ide/HaRePlugin.hs
```diff
@@ -312,96 +315,137 @@ makeRefactorResult changedFiles = do
...
-                  Null)
+    Nothing -> hoistEither $ nonExistentCacheErr "getReferencesInDoc"
     Just cm -> do
       let tc = tcMod cm
           parsed = pm_parsed_source $ tm_parsed_module tc
           nameMap = initRdrNameMap tc
-      case locToRdrName (row, col) parsed of
-        Nothing -> return $ IdeResponseOk Nothing
-        Just pn@(L l _) -> pure $ IdeResponseOk (Just $ L l $ rdrName2NamePure nameMap pn)
+          invertedMap = invert nameMap
```
alanz: You should consider putting this inside the cached module, so it gets inverted once, rather than on every use. Or would it get memoised?

PR #262, opened 2017-07-17
Added hoogle instructions to README

wz1000:


README.md
```diff
@@ -73,6 +73,33 @@ npm install .
...
+To generate a global hoogle database for your user, run
+
+```bash
+$ hoogle generate
+```
+
+To generate a project specific database for a stack project, run
+
+```bash
+$ stack hoogle
```
alanz: It seems you have to do a `stack build` after this, otherwise `hie` can't load files in the project, the .stack-work has the pre-haddock build artefacts listed, which ghc-mod uses

PR #266, opened 2017-07-22
Some fixes/improvements

wz1000:


src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -331,10 +318,21 @@ reactor (DispatcherEnv cancelReqTVar wipTVar versionTVar) plugins lf st cin inp
...
 
+        rid <- nextLspReqId
         reactorSend $ fmServerRegisterCapabilityRequest rid registrations
 
+        lf <- ask
+        let hreq = PReq Nothing Nothing Nothing callback
+                     Hoogle.initializeHoogleDb
+            callback Nothing = flip runReaderT lf $
+              reactorSend $
+                fmServerShowMessageNotification J.MtWarning "No hoogle db found. Check the README for instructions to generate one"
```
alanz: Eventually we should give instructions as per the specific project identified. stack / cabal / nix etc
If that info is available from ghc-mod

PR #278, opened 2017-08-11
Rebase of @wz1000 cleanup branch

alanz:


hie-build-plugin/Haskell/Ide/BuildPlugin.hs
```diff
@@ -100,21 +111,91 @@ buildPluginDescriptor = PluginDescriptor
...
+        ,cpStack   :: Maybe String
+        ,cpFile    :: Uri
+    } deriving Generic
+
+instance FromJSON CommonParams where
+  parseJSON = J.genericParseJSON $ customOptions 2
+instance ToJSON CommonParams where
+  toJSON = J.genericToJSON $ customOptions 2
+
+incorrectParameter = undefined
```
arrowd: I would have to reimplement this somehow?

hie-build-plugin/Haskell/Ide/BuildPlugin.hs
```diff
@@ -216,16 +302,31 @@ flagToJSON f = object ["name" .= ((\(FlagName s) -> s) $ flagName f), "descripti
 
 -----------------------------------------------
 
-buildDirectory :: CommandFunc ()
-buildDirectory = CmdSync $ \ctx req -> withCommonArgs ctx req $ do
+data BuildParams = BP {
+         -- common params. horrible
```
arrowd: I didn't get why there are 3 things now: `CommonArgs`, `CommonParams`, and `BuildParams`.

hie-build-plugin/Haskell/Ide/BuildPlugin.hs
alanz: yes. I forgot about that, to be honest.

This is very much a "smallest effort to get it to compile" version

hie-build-plugin/Haskell/Ide/BuildPlugin.hs
alanz: Because of the way the plugin stuff works you need to specify a type that comes from the client, and is decoded from JSON.

The reason there are three types is that the function passed in to `withCommonArgs` expects the common arguments in a structure that is always populated, so cannot have `Maybe` values. But some common params are optional, so the `CommonParams` structure exists to manage this.

The `BuildParams` are an extension of the `CommonParams`, but we need to give a single type that can be serialised through JSON in a natural way.  Hence the duplication.

Thinking about it, we might be able to make completely off-the-wall JSON instances where `BuildParams` has a `CommonParams` field and then the extension, but it maps out to a flat JSON structure as expected on the wire.

@wz1000 thoughts?

wz1000: Another option could be using an anonymous record library like vinyl.

alanz: Perhaps we should merge it as is, and then clean it up?

The intent here was to get it into the new architecture.

hie-build-plugin/Haskell/Ide/BuildPlugin.hs
```diff
@@ -100,21 +111,91 @@ buildPluginDescriptor = PluginDescriptor
...
+instance FromJSON CommonParams where
+  parseJSON = J.genericParseJSON $ customOptions 2
+instance ToJSON CommonParams where
+  toJSON = J.genericToJSON $ customOptions 2
+
+incorrectParameter = undefined
+
+withCommonArgs (CommonParams mode0 mDistDir mCabalExe mStackExe fileUri) a = do
+      case readMode mode0 of
+        Nothing -> return $ incorrectParameter "mode" ["stack","cabal"] mode0
```
wz1000: I think a better approach would be

```
data Mode = Stack | Cabal

instance ToJSON/FromJSON Mode where 
  parseJSON (A.String "stack") = Stack
  parseJSON (A.String "cabal") = Cabal
  ...
```

This way all the parameter error handling is pushed into Aeson

PR #286, opened 2017-08-14
Fix JsonStdio transport not to exit as soon as it started.

arrowd:


src/Haskell/Ide/Engine/Transport/JsonStdio.hs
```diff
@@ -79,6 +79,7 @@ run dispatcherProc cin = flip E.catches handlers $ do
     _opid <- forkIO $ outWriter rout
     _rpid <- forkIO $ reactor rout
     dispatcherProc dispatcherEnv
+    forever $ threadDelay maxBound
```
wz1000: how about
```haskell
dispatcherProc dispatcherEnv
race_ (outWriter rout) (reactor rout)
```

wz1000:
See https://github.com/wz1000/haskell-ide-engine/commit/d2895ff34bc1947b50be80151875a6378f789387

arrowd:
Whatever you think is better.

wz1000:
take a look at #297

PR #292, opened 2017-08-16
Added URI decoding for building ghc-mod URIs

Friede80:
This is probably more an issue with ghc-mod, but `GhcMod.ModuleLoader.runActionWithContext` never returns if the URI is percent encoded. So I've added percent decoding before building `GhcMod.ModuleLoader.FileUri`s


alanz:
Not that I think of it, the ghc-mod FileUri is a clone of the Uri in haskell-lsp, which exists to not have to couple the two.  So perhaps this PR belongs on my ghc-mod branch, as referred to in the stack.yaml

PR #297, opened 2017-08-18
Fix JSONStdio

wz1000:
also cleanup stack.yaml

Responses are also separated by \STX now.

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/JsonStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/JsonStdio.hs
@@ -10,7 +10,6 @@ module Haskell.Ide.Engine.Transport.JsonStdio
     jsonStdioTransport
   ) where
 
-import           Control.Concurrent
 import           Control.Concurrent.Async
 import           Control.Concurrent.STM.TChan
 import           Control.Concurrent.STM.TVar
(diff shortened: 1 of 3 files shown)


wz1000:
supersedes #286

src/Haskell/Ide/Engine/Transport/JsonStdio.hs
```diff
@@ -76,9 +76,9 @@ run dispatcherProc cin = flip E.catches handlers $ do
           , docVersionTVar = versionTVar
           }
 
-    _opid <- forkIO $ outWriter rout
-    _rpid <- forkIO $ reactor rout
     dispatcherProc dispatcherEnv
+    race_ (outWriter rout) (reactor rout)
+
```
arrowd: `race` is `waitEither`, while we need `waitBoth` here, no?

wz1000: Since both are infinitely looping, I don't think it matters much in practice. But with `race_` , if one thread exits normally, the other is killed, which is behaviour we probably want.

I could make the type signatures `forall void. IO void` to make this explicit though. Then we could get rid of the unreachable `return 0`.

Issue #294, opened 2017-08-16
Not working with TypeApplications defined in cabal file


adelbertc:
Minimal reproduction:

Main.hs
```haskell
module Main where

x :: a -> a
x = id

y :: String
y = x @String "hello"

main :: IO ()
main = putStrLn "hello"
```

foo.cabal
```cabal
name: foo
version: 1.0.0
license: BSD3
license-file: LICENSE
cabal-version: >= 1.18
build-type: Simple

executable project0
    build-depends: base < 5
    main-is: Main.hs
    default-language: Haskell2010
    default-extensions: TypeApplications
```

Clone-able Git repo here (though I may delete it in the future, hence why I reproduced it above): https://github.com/adelbertc/typeapptest

`cabal build` is happy with it, but HIE reports

```
Parse error in expression: x@String.   y = x @String "hello".   . > main :: IO ().   main = putStrLn "hello". .
```

However putting the LANGUAGE pragma makes it happy.

I confirm this is HIE since when I remove HIE configuration from my editor (Neovim) and re-open the file the errors are no longer reported. I also confirm Cabal is doing the right thing since removing the default-extensions field and doing a cabal build fails.

This might possible be related to HLint if HIE uses HLint under the covers since until recently HLint had issues with TypeApplications: https://github.com/ndmitchell/hlint/issues/296 Looking at stack.yaml looks like HIE is on `nightly-2017-06-16` which may have too old of a version for [haskell-src-exts](https://www.stackage.org/nightly-2017-06-16/package/haskell-src-exts-1.18.2 ) ([HLint](https://www.stackage.org/nightly-2017-06-16/package/hlint-2.0.9) seems fine though)

**EDIT** Hm pinning the revision of haskell-src-exts to 1.19.1 (https://github.com/haskell-suite/haskell-src-exts/tree/1.19.1) still gives the same error:

```yaml
+- location:
+    git: https://github.com/haskell-suite/haskell-src-exts
+    commit: a095cd5b4a0159306426f3e15c2b98e96ca25018
+  extra-dep: true
```

wz1000:
fixed here: https://github.com/haskell/haskell-ide-engine/commit/f7dbdd9ddcca66af30d53c53ae29f520a19e30e5

adelbertc:
[...]

PR #298, opened 2017-08-18
 Enable TypeApplications in hlint by default, update for latest haskell-lsp

wz1000:


wz1000:
Fixes #294 and https://github.com/alanz/vscode-hie-server/issues/3

PR #335, opened 2017-10-17
Don't propagate error if no definition is found

AlexeyRaga:
## Changes

Current behavior: when for some reason `HIE` is unable to find a definition for a given symbol to navigate to, the error is returned to the client:
`hare:findDef:<no location found>`.
This includes all the symbols that are defined outside of the current project (`base` and libraries)

I find that error message unhelpful and noisy. There is no way client can compensate for it and no way user can benefit from it or fix it. 

With the suggested fix `HaRe` will return an empty successful response if it can't find a location to navigate to. `VSCode` deals with it correctly by not showing the possibility to jump to the definition.  

I suggest to use `?w=1` in github review (https://github.com/haskell/haskell-ide-engine/pull/335/files?w=1) to see the change without an indentation noise.

### DIFF
--- a/hie-hare/Haskell/Ide/HaRePlugin.hs
+++ b/hie-hare/Haskell/Ide/HaRePlugin.hs
@@ -609,53 +609,56 @@ findDef uri pos = pluginGetFile "findDef: " uri $ \file -> do
         Nothing -> return $ IdeResponseOk []
         Just pn -> do
           let n = snd pn
-          res <- srcSpan2Loc rfm $ nameSrcSpan n
-          case res of
-            Right l@(J.Location luri range) ->
-              case oldRangeToNew cm range of
-                Just r  -> return $ IdeResponseOk [J.Location luri r]
-                Nothing -> return $ IdeResponseOk [l]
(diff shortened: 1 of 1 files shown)


hie-hare/Haskell/Ide/HaRePlugin.hs
```diff
@@ -609,53 +609,56 @@ findDef uri pos = pluginGetFile "findDef: " uri $ \file -> do
...
-                              getNewNames n
-                            eithers <- mapM (srcSpan2Loc rfm' . nameSrcSpan) newNames
-                            case rights eithers of
-                              (l:_) -> return $ IdeResponseOk [l]
-                              []    -> failure
-                      Nothing -> failure
-                    else failure
-                Nothing -> failure
+          case nameSrcSpan n of
+            UnhelpfulSpan _ -> return $ IdeResponseOk []
```
AlexeyRaga: The real change is this (#612-#615), the rest is just indentation because of an extra `case`.

alanz:
Thanks

PR #347, opened 2017-11-01
Only give a hover result if we already have the cached module

alanz:
To speed things up.

@wz1000 I want to bring this in now, as it has a big responsiveness impact.

You can then rework the rest as per our new architecture discussion.

Note: The fancy hoogle lookup has had to be lost in the process. Tradeoff.


wz1000:
ok, looks good. I will try to put something together by the end of the weekend.

alanz:
@wz1000 there is no pressure for you to do anything timewise. I just want to get this out there in the meantime.

PR #364, opened 2017-11-24
Haskell lsp nightly

alanz:
Use `haskell-lsp-0.2.0.0` from the latest stackage nightly. For the GHC 8.2.1 build

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -38,7 +38,7 @@ library
                      , ghc >= 8.0.1
                      , ghc-mod-core
                      , gitrev >= 1.1
-                     , haskell-lsp
+                     , haskell-lsp >= 0.2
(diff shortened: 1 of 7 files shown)


Tehnix:
Does this bring GHC 8.2 support for HIE, or is it still blocking on ghc-mod?

alanz:
It is still blocked on ghc-mod /cabal-helper.

And I fear my next task is to dive into that swamp, as @DanielG does not seem to have the time to do it.

Tehnix:
Ah, damn :( 

Perhaps there should be more coordination between GHC releases and core tooling such as this, although I understand that the cabal-helper update was quite big, with backpack and all.

alanz:
Yes, backpack and new-build need to be digested. Both are *major* changes.

PR #373, opened 2017-11-27
Don't force ghc-mod to use -Wall switch

nponeccop:
It will pick up the options from .cabal

Closes #918 - Extra warnings were displayed
for people who used -Wall and -Wno-xxx flags.
An extra -Wall added by HIE just cancelled
the effect of -Wno switches.

This change breaks the behaviour for people who
omitted -Wall in their .cabal files and relied
on HIE to add the switch. They should just
add -Wall to ghc-options in .cabal.


nponeccop:
invalid issue no

PR #374, opened 2017-11-27
Don't force ghc-mod to use -Wall switch

### LINKED ISSUE #365: Respect ghc-options
If I have something like this in `.cabal` HIE outputs the suppressed warnings:

```cabal
  ghc-options: -Wall -Wno-name-shadowing
    -Wno-unused-matches -Wno-incomplete-patterns -Wno-missing-signatures -Wno-orphans
```
I.e. `stack build` shows 1 warning but HIE shows a dozen

nponeccop:
It will pick up the options from .cabal

Closes #365 - Extra warnings were displayed
for people who used -Wall and -Wno-xxx flags.
An extra -Wall added by HIE just cancelled
the effect of -Wno switches.

This change breaks the behaviour for people who
omitted -Wall in their .cabal files and relied
on HIE to add the switch. They should just
add -Wall to ghc-options in .cabal.

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -118,7 +118,7 @@ run opts = do
 
   let vomitOptions = GM.defaultOptions { GM.optOutput = oo { GM.ooptLogLevel = GM.GmVomit}}
       oo = GM.optOutput GM.defaultOptions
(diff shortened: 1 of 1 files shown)


alanz:
Looks good, thanks.

Issue #329, opened 2017-10-11
HIE reports invalid file uri on Windows


Anrock:
See https://github.com/autozimu/LanguageClient-neovim/issues/144
Looks like diagnostics from ghcmod are using invalid file path separator.

Anrock:
@alanz please take a look.

alanz:
Can you try again with current HIE master?

Anrock:
@alanz Master stopped building on Win since 8b9b623 because of `unix` dep in `hie-build-plugin-0.1.0.0`
Latest buildable commit: 7f77e7e, issue still present there.

alanz:
Where is the `unix` dep? Any idea what is bringing it in?

Anrock:
I have no idea right now. 
It looks like haskell tooling doesn't have something to list full dependency tree of package (omg, why?), so the only viable option i see now is to analyze diff of broken commit and find introduced dependency manually.

Anrock:
@alanz found it.
In 8b9b623 you switched from alanz/cabal-helper `94b7ad` to DanielG/cabal-helper `4af5dddb`
Previous commit of DanielG/cabal-helper `4af5dddb` - `4b3571a` introduces `unix` package to use instead of `time`.

Anrock:
@alanz so i've used local repo of cabal-helper with `4b357a` reverted, hie built fine.
Still file uri is wrong:
```
17:26:23 DEBUG   [RPC-haskell] <= {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///C:\\Users\\Anrock\\Desktop\\test.hs","diagnostics":[{"severity":1,"range":{"start":{"line":5,"character":8},"end":{"line":5,"character":12}},"source":"ghcmod","message":"Parse error: naked expression at top level\nPerhaps you intended to use TemplateHaskell"}]}}
```

alanz:
Ok, thanks.

Can you paste an edited version of what you would be expecting to see?

Anrock:
> Can you paste an edited version of what you would be expecting to see?
You mean edited server response? Honestly i have no idea how LSP protocol works.

As far as i understood explanations of @autozimu in autozimu/LanguageClient-neovim#144 and [reference](https://en.wikipedia.org/wiki/File_URI_scheme#Windows) [urls](https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/) he posted 
offending path is in `params.uri`
[...]

PR #385, opened 2017-12-04
Use newer cabal-helper to avoid choking on CFLib

nponeccop:
Probably we need to clean our cache of cabal-helpers regularly to catch such problems (it's in another place from other build products)

See #329


alanz:
There is also this branch: https://github.com/alanz/haskell-ide-engine/commit/6941e6908c3a1141909526030e8876e3fc905a42

It uses the one under dev, with cabal 2.0 support.

But there is still one tiny thing before the merge.

cc @DanielG

Issue #395, opened 2017-12-09
Cabal-helper uses older snapshots for its runtime part


nponeccop:
It appears that circleci builds use inconsistent snapshots. Below is an excerpt from build 100, `stack test -j 2 haskell-ide-engine` step. You can see that "runtime" cabal-helper and/or `testdata` package is built using `nightly-2017-06-16` while its `library` part uses `lts-9.14` (which is correct). 

```
functional spec
  consecutive plugin commands
Downloading nightly-2017-06-16 build plan ...
Downloaded nightly-2017-06-16 build plan.
mtl-2.2.1: using precompiled package
text-1.2.2.2: using precompiled package
parsec-3.1.11: using precompiled package
testdata-0.1.0.0: configure (exe)
Configuring testdata-0.1.0.0...
Completed 4 action(s).
[1 of 5] Compiling CabalHelper.Shared.Sandbox ( /tmp/cabal-helper-source30336/CabalHelper/Shared/Sandbox.hs, cabal-helper0.8.0.0-Cabal1.24.2.0.build/CabalHelper/Shared/Sandbox.o )
[2 of 5] Compiling CabalHelper.Shared.InterfaceTypes ( /tmp/cabal-helper-source30336/CabalHelper/Shared/InterfaceTypes.hs, cabal-helper0.8.0.0-Cabal1.24.2.0.build/CabalHelper/Shared/InterfaceTypes.o )
[3 of 5] Compiling CabalHelper.Shared.Common ( /tmp/cabal-helper-source30336/CabalHelper/Shared/Common.hs, cabal-helper0.8.0.0-Cabal1.24.2.0.build/CabalHelper/Shared/Common.o )
[4 of 5] Compiling CabalHelper.Runtime.Licenses ( /tmp/cabal-helper-source30336/CabalHelper/Runtime/Licenses.hs, cabal-helper0.8.0.0-Cabal1.24.2.0.build/CabalHelper/Runtime/Licenses.o )
```

alanz:
When you change some deps (like cabal-helper), you have to manually "run without cache", to restart the build clean.

nponeccop:
ok I've initiated build 101 to see

nponeccop:
**Update:** The inconsistency is still there in https://circleci.com/gh/haskell/haskell-ide-engine/101. It's a clean build and you can check that "no cache is used" during `Restoring cache` step.

```
functional spec
  consecutive plugin commands
[...]

PR #402, opened 2017-12-11
Read resolvers from stack*.yaml instead of hardcoding

nponeccop:
to avoid code duplication and inconsistencies such as
issue #395. See discussion there for rationale.

Now for each compiler the resolver is read from corresponding
stack.yaml using a YAML parser.


alanz:
Looks good, waiting for CI

alanz:
oops, supposed to be 8.2.2, for the second one

You are welcome to fix it, if you have time

nponeccop:
This is an unrelated issue, this PR doesn't change that file.

Issue #393, opened 2017-12-07
Hoogle lookups not working on local db


stellarhoof:
As per the README I've tried:
```
$ hoogle generate
$ stack exec hoogle generate
```
I've enabled logging but it (unhelpfully) shows the same information as in the popup in VSCode:
```
2017-12-07 18:35:02.397034 [ThreadId 12] - <--2--{"jsonrpc":"2.0","method":"window/showMessage","params":{"type":2,"message":"No hoogle db found. Check the README for instructions to generate one"}}
```
Then on trying
```
$ stack hoogle
$ stack build
```
I get 
```
2017-12-07 18:49:35.225028 [ThreadId 12] - <--2--{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":4,"message":"Using hoogle db at: /Users/berserk/Source/dhall-haskell/.stack-work/hoogle/x86_64-osx/lts-9.6/8.0.2/database.hoo"}}
```
which makes sense. It's using the local db. However on opening `https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall.hs` and trying to get the type for `Control.Applicative` doesn't work.

`hoogle --version` returns `5.0.14`. Is it possible that version isn't looked for? I've installed `5.0.13` as per #285 and it finds the database. However as with the local database it still doesn't find type definitions for `Control.Applicative`. Doing `hoogle Control.Applicative` on the cmdline works fine.

stellarhoof:
Also hovering on `atomically` on `src/Haskell/Ide/Engine` doesn't return documentation like in the gif in the README.

wz1000:
@alanz changed the hover request to immediately respond even if a recompile
is in progress. However this meant disabling documentation lookup due to
some technical issues. I am working on getting this re-enabled ASAP.

alanz:
Speed, accuracy. Choose one :)

AlexeyRaga:
> Speed, accuracy. Choose one :)

1. Make it work
2. Make it beautiful
3. Make it fast

;)

alanz:
Actually the  docs are always accurate, it is just where they come from, whether a bigger or smaller db is consulted.

wz1000:
I've got hoogle working with async on my branch now! 

https://github.com/wz1000/haskell-ide-engine

It will be merged into this repo in a couple of days.

stellarhoof:
[...]

PR #403, opened 2017-12-12
Async Architecture and reenabling hoogle/haddock

wz1000:
- Added `AsyncM`
- Re-enabled hoogle
- Got haddock-plugin working on ghc 8.2 and re-enabled it


wz1000:
This fixes #393

alanz:
BTW, we need to add `pedantic` to `hie-haddock`.

PR #407, opened 2017-12-13
Add EKG support

alanz:
Launch with the `--ekg` option to `hie`, and point a browser at
http://localhost:8000

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -2,6 +2,8 @@
 {-# LANGUAGE TemplateHaskell     #-}
 module Main where
 
+
+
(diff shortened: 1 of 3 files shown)


soylens:
Is this the TCP/IP interface the Atom client(win) mentions?

alanz:
No, this is a mini web server giving a view into the running HIW process, to see memory, cpu etc usage.

See https://hackage.haskell.org/package/ekg

soylens:
There should probably be a command line option to specify the port in case localhost:8000 is occupied

alanz:
I agree.

PR #408, opened 2017-12-13
Make the default stack.yaml be for GHC 8.2.2

alanz:


circle.yml
```diff
@@ -14,6 +14,8 @@ dependencies:
     - stack upgrade
```
nponeccop: Why do we do `stack setup` 2 times (see the log).

circle.yml
```diff
@@ -32,16 +34,16 @@ dependencies:
         timeout: 7200
 
     - stack clean
-    - stack --stack-yaml=stack-8.2.2.yaml setup
-    - stack --stack-yaml=stack-8.2.2.yaml build  -j 2 --test --only-dependencies :
+    - stack --stack-yaml=stack-8.0.2.yaml setup
+    - stack --stack-yaml=stack-8.0.2.yaml build  -j 2 --test --only-dependencies :
         timeout: 7200
-    - stack --stack-yaml=stack-8.2.2.yaml build -j 2 :
+    - stack --stack-yaml=stack-8.0.2.yaml build -j 2 :
```
nponeccop: This line doesn't make sense to me. Basically we build HIE just to clean it later and rebuild (and make 2x build time)

circle.yml
```diff
@@ -50,6 +52,6 @@ test:
     - stack --stack-yaml=stack-8.2.1.yaml test -j 2  haskell-ide-engine :
         timeout: 14400
 
-    - stack --stack-yaml=stack-8.2.2.yaml clean
-    - stack --stack-yaml=stack-8.2.2.yaml test -j 2  haskell-ide-engine :
+    - stack --stack-yaml=stack-8.0.2.yaml clean
```
nponeccop: This one should be 8.2.1 - we clean the previous build

circle.yml
alanz: The initial build is to get the deps installed, I think.

And I was getting builds stomping on each other, so perhaps went overboard on this stuff.

circle.yml
alanz: ok

alanz:
BTW, technically this PR is intended to just be a renaming of files, to bring a different default into scope for `stack.yaml`.

So perhaps we should keep it as that, and then clean it up.

nponeccop:
> So perhaps we should keep it as that, and then clean it up.

I agree

PR #422, opened 2017-12-20
Undelete type cmd

### LINKED ISSUE #420: ghcModPlugin "type" command removed
Hey,

I'm not completely familiar with how this whole repo fits together but I notice that the `showType` command from https://github.com/alanz/vscode-hie-server no longer works - seems to be due to these changes from #403 

https://github.com/haskell/haskell-ide-engine/pull/403/files#diff-7b73da39d2cdccd6267a96bbb00d7fd6L56

I'm just wondering - is there another way that we should try to request the same functionality from now on or is it an oversight that `newTypeCmd` or something similar are not exposed to the client or are there plans for something else?

wz1000:
Fixes #420

### DIFF
--- a/hie-ghc-mod/Haskell/Ide/GhcModPlugin.hs
+++ b/hie-ghc-mod/Haskell/Ide/GhcModPlugin.hs
@@ -57,18 +57,6 @@ ghcmodDescriptor = PluginDescriptor
       ]
   }
 
-ghcmodTestDescriptor :: PluginDescriptor
-ghcmodTestDescriptor = PluginDescriptor
-  {
-    pluginName = "ghc-mod(test)"
-  , pluginDesc = "ghc-mod is a backend program to enrich Haskell programming "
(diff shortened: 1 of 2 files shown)


hie-ghc-mod/Haskell/Ide/GhcModPlugin.hs
```diff
@@ -53,6 +53,19 @@ ghcmodDescriptor = PluginDescriptor
...
       , PluginCommand "lint" "Check files using `hlint'" lintCmd
       , PluginCommand "info" "Look up an identifier in the context of FILE (like ghci's `:info')" infoCmd
+      , PluginCommand "type" "Get the type of the expression under (LINE,COL)" typeCmd
+      ]
+  }
+
+ghcmodTestDescriptor :: PluginDescriptor
+ghcmodTestDescriptor = PluginDescriptor
+  {
+    pluginName = "ghc-mod(test)"
```
alanz: Why is this suffixed "(test)"?

Issue #391, opened 2017-12-06
Giant popup in Sublime


nponeccop:

nponeccop:
@xldenis wrote:

So far no luck! I've removed the cabal helpers and nothing changed. I'm currently rebuilding from _latest_ master as my previous version was:
```
Version 0.1.0.0, Git revision 04af8885a89402168cf464b119697d2539df1e61 (1071 commits) x86_64
```

I'm building with `stack build --stack-yaml=stack-8.2.1.yaml` and that produces a binary:

```
Version 0.1.0.0, Git revision 358c6489eb1724b5d29698fe28fc389c12cdc10f (1096 commits) x86_64
```

Wow! it doesn't crash now but leads to an almost _worse_ error... 
![image](https://user-images.githubusercontent.com/451373/33590258-c0cbdcb8-d943-11e7-8e60-e480dc7dc8d8.png)

Pops up everytime I save! Strictly speaking that error message is _correct_ they aren't in `other-modules` they are in `exposed-modules` I don't know what could be causing this though.

Initial research shows that this could be a problem for cabal files that specify multiple targets (ie: app + lib) ?

nponeccop:
From messages like 

```
ue,"completionProvider":{"resolveProvideirn"f:ot:r uSeT,A"CtKr_iEggXeEr Csheatr,a cptreerfse"r:r[i"n.g" ]S}t,a"cdke fpirnoijteicotn:
```
it seems we have a racing condition. Hopefully it's just the logger.

The popups come from these lines:

```
2017-12-04 22:42:24.254393 [ThreadId 12] - <--2--{"jsonrpc":"2.0","method":"window/showMessage","params":{"type":1,"message":"Got error while processing diagnostics: These modules are needed for compilation but not listed in your .cabal file's other-modules: Control.Monad.Fresh\n                                                                                             Control.Monad.Unify\n                                                                                             Ill.Syntax.Core\n                                                                                             Ill.Syntax.Kind\n                                                                                             Ill.Syntax.Literal\n                                                                                             Ill.Syntax.Name\n                                                                                             Ill.Syntax.Pretty\n                                                                                             Ill.Syntax.Type"}}:
```

xldenis:
My suspicion is that it's improperly building the executable as if it's the only targget...

xldenis:
I build my project iwht `stack build` and the error never happens.

FongHou:
[...]

PR #436, opened 2018-01-04
Temporary workaround for #391

jchia:
Although it hides the real issue, this change allows HIE to continue to be used by getting rid of the incessant popups.


wz1000:
Thanks

PR #448, opened 2018-01-24
Convert all LICENSE files into copies

domenkozar:
This is an unfortunate consequence of Nix packages
being restricted filesystem access to only root of
the package, so ../LICENSE doesn't exist.


alanz:
And they should probably be in favour of a contributors list, rather than me

Issue #435, opened 2018-01-03
Building cabal-helper fails on Windows 10


capitanbatata:
While trying to build HIE on windows 10 I get he following error:

```text
haddock-library-1.4.4: build
haddock-library-1.4.4: copy/register
Progress: 2/16
--  While building custom Setup.hs for package cabal-helper-0.8.0.0 using:
      C:\Users\nadalesagutde\Documents\GitHub\haskell-ide-engine\.stack-work\downloaded\XcrjBfxqjaxS\.stack-work\dist\5c8418a7\setup\s
etup --builddir=.stack-work\dist\5c8418a7 build lib:cabal-helper exe:cabal-helper-wrapper --ghc-options " -ddump-hi -ddump-to-file -fd
iagnostics-color=always"
    Process exited with code: ExitFailure 1
    Logs have been written to: C:\Users\nadalesagutde\Documents\GitHub\haskell-ide-engine\.stack-work\logs\cabal-helper-0.8.0.0.log

    Preprocessing library for cabal-helper-0.8.0.0..
    Building library for cabal-helper-0.8.0.0..
    Preprocessing executable 'cabal-helper-wrapper' for cabal-helper-0.8.0.0..
    Building executable 'cabal-helper-wrapper' for cabal-helper-0.8.0.0..
    [ 9 of 12] Compiling CabalHelper.Compiletime.Data ( src\CabalHelper\Compiletime\Data.hs, .stack-work\dist\5c8418a7\build\cabal-hel
per-wrapper\cabal-helper-wrapper-tmp\CabalHelper\Compiletime\Data.o )
    [10 of 12] Compiling Paths_cabal_helper ( .stack-work\dist\5c8418a7\build\cabal-helper-wrapper\autogen\Paths_cabal_helper.hs, .stack-work\dist\5c8418a7\build\cabal-helper-wrapper\cabal-helper-wrapper-tmp\Paths_cabal_helper.o )
    .stack-work\dist\5c8418a7\build\cabal-helper-wrapper\cabal-helper-wrapper-tmp\.stack-work\dist\5c8418a7\build\cabal-helper-wrapper
```

I don't know whether windows is a supported platform, but this might be related to long path names.

nponeccop:
Windows is a supported platform.

capitanbatata:
Maybe I can help with setting up an appveyor CI then, and updating the build instructions on windows.

nponeccop:
[...]

PR #455, opened 2018-01-30
Add installation instructions for windows users

ShrykeWindgrace:
Add installation instructions for windows users who are facing the `path too long` issue on windows. Addresses #435

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,15 @@ stack --stack-yaml=stack-8.2.1.yaml install
 stack --stack-yaml=stack-8.0.2.yaml install
 ```
 
+#### Installation on Windows
+
+In order to avoid problems with long paths you can do the following:
+
+1. Edit the group policy: set "Enable Win32 long paths" to "Enabled". Works only only for Windows 10
+
+2. Clone the `haskell-ide-engine` to the root of your logical drive (e.g. to `E:\hie`)
+
+
 ### Installation with Nix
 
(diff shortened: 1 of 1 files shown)


alanz:
Thanks. Every bit helps

nponeccop:
@alanz apparently it's not #435 but an unrelated issue. And now we have an instruction suggesting fairly terrible things to everyone on Windows. While in fact it's not for everyone but for those using " haskell-ide-engine somewhere deep in folder hierarchy" as @ShrykeWindgrace put it in #435.

The practice of creating anything out of home folder is as terrible on Windows as it is on Linux (although historically nobody cares - the tradition does back all the way to DOS and it's predecessors). "Normal" folders complying with Windows analog of Filesystem Hierarchy Standard such as `c:\users\username\Documents\dev\foo\bar\quux\haskell-ide-engine` work just fine, the path just cannot be too long. So the suggestion should explicitly mean that it's a workaround for a rarely encountered issue.

alanz:
Ok, so what to do?

nponeccop:
> the installation instruction should explicitly mention that it's a workaround for a rarely encountered issue.

Either I or @ShrykeWindgrace will change the instruction text when we have time. Hopefully tomorrow.

ShrykeWindgrace:
On a second thought, I'd say it's better to revert the merge altogether. The workaround is windows-centric and has little to do with `hie` itself.

PR #457, opened 2018-02-01
Fix build error

vimuel:
This failed to build (with GHC 8.2.2):
```
error: \u used with no following hex digits; treating as '\' followed by identifier [-Werror,-Wunicode]
```
Adding a space resolved it on my machine.


alanz:
Weird, but makes sense. Thanks.

vimuel:
Weird indeed.

PR #484, opened 2018-03-01
Make completions fuzzy, utilize cachedData and prevent it from being overwritten

dpren:


alanz:
I am happy to merge this, then we can try to understand current issues and work from there.

The module caching in master is already broken, this does make it somewhat better.

I am still seeing issues on hover though, which seems not to be async anymore.. Or is async, but does not use the stored module.

wz1000:
See https://github.com/dpren/haskell-ide-engine/pull/1

wz1000:
Another thing - I don't seem to get completion results for qualified items imported without explicit qualification

For example

    import Data.Map
    Map.

will not give me completions for all the stuff inside Map. It would be nice if it could handle this case.

dpren:
Ok, I think I'm covering all the import cases now.
https://wiki.haskell.org/Import

PR #485, opened 2018-03-03
Remove `stack hoogle` support, update README, improve haddock rendering

wz1000:


alanz:
LGTM, still getting docs, and less admin.

Issue #462, opened 2018-02-05
No other features work while there is a compilation error


lorenzo:
Making errors while coding is quite common, and a good IDE should help fixing them. Currently, when a compilation error happens in a project (syntax error, type mismatch or otherwise), no other feature in the IDE engine work.

Would it be possible to retain certain features that would make fixing the error easier, like hovering on a symbol, or finding definitions?

nponeccop:
There is some support for that in the form of caching the last good state of each module, but there are few issues: #460 , #416

Full support is very hard to implement, and it can be split roughly into three parts:
- robust parsing
- robust name resolution
- robust type checking

Issue #460, opened 2018-02-04
Manage initial module loading UI process


alanz:
At the moment, when a file is opened, we request diagnostics on it as usual, and return a "Module not Loaded" error message.

Once the compilation is complete, if it succeeds, then we start getting hover info etc.
If the compilation fails, we continue to get the "Module not Loaded" warning. 

We need to manage UI messaging better.

1. Initial "Loading module" message rather than an error
2. Return a message when the loading process completes, with the outcome
3. If the module load fails due to errors in the current module, send those as diagnostics
4. What if the load fails for errors in another module?

nponeccop:
See also the monotonicity considerations of #416. Essentially is

5. What if module compilation fails, but we have stale compilation results for it?

Especially in the case 4.

alanz:
There is also case 0, hie GHC version does not match project GHC version

alanz:
We need to revisit this, based on real world feedback.

cc @reactormonk

Issue #416, opened 2017-12-19
"Didn't get typechecked module" message


nponeccop:
From what I understand the cache of typechecked modules is supposed to be "monotonic": once completion works for a module, it never stops working.

But apparently it's not the case. I load a haskell file in neovim and completion works. But after some time it stops working, and when I try to complete something there is the following message in `hie.log`:

```
 setTypecheckedModule: Didn't get typechecked module for: "/home/foo/bar/Main.hs"
```

The message is produced by hie-ghc-mod:

https://github.com/haskell/haskell-ide-engine/blob/4001c3e9fea3ad36ec7eddcb3ddbb5c08b4dbd1e/hie-ghc-mod/Haskell/Ide/GhcModPlugin.hs#L150-L159

meck:
I've been looking at this as well.  I have the same messages, but I think its related to [Language Client](https://github.com/autozimu/LanguageClient-neovim), it seems to correspond to the errors there:
 `Some arguments are missing, requesting from vim. Keys: ["buftype", "filename", "line"]. Exps: ["&buftype", "s:Expand(\'%:p\')", "line(\'.\') - 1"]` 

 I am thinking the rust wrapper maybe failing to expand those properly and confusing ghc-mod, but I've not had time to look at it further. Perhaps @autozimu has a insight? The completion was working before the new rust based version

autozimu:
No, that’s not an error and that message is totally fine. The message basically says those information are not included in the request, languageclient needs to ask vim to fulfill those info though RPC call.

nponeccop:
@autozimu You definitely need to have it at `INFO` log level so people don't get misleaded

PR #488, opened 2018-03-07
Do not return error message for hover and symbols req

alanz:
IF the module is not currently loaded, simply return an empty result
for hover request and document symbols request.

Affects #416, #460, #462

I am not sure that it closes any of them though.

### DIFF
--- a/src/Haskell/Ide/Engine/Dispatcher.hs
+++ b/src/Haskell/Ide/Engine/Dispatcher.hs
@@ -53,7 +53,7 @@ ideDispatcher env pin = forever $ do
     res <- action
     liftIO $ callback res
 
-ghcDispatcher :: foall void. DispatcherEnv -> TChan GhcRequest -> IdeGhcM void
+ghcDispatcher :: forall void. DispatcherEnv -> TChan GhcRequest -> IdeGhcM void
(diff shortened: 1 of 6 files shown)


src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -23,7 +23,7 @@ data DispatcherEnv = DispatcherEnv
   , docVersionTVar     :: !(TVar (Map.Map Uri Int))
   }
 
-dispatcherP :: forall void. DispatcherEnv -> TChan PluginRequest -> IdeGhcM void
+dispatcherP :: DispatcherEnv -> TChan PluginRequest -> IdeGhcM ()
```
wz1000: Why is this needed? The old type is more descriptive, as it states the the function never "returns", i.e. loops forever.

The old type is also strictly more general than the new type(IdeGhcM () is a subtype of forall void. IdeGhcM void)

wz1000:
This seems good for now, but we really need a ModuleStatus flag, that tracks whether the module compiled successfully or not, whether the typechecked module we have is fresh or not etc.

alanz:
I agree on the module status, it ties into the life cycle being explicit as for #460 .

But this sorts out an initial annoyance, paving the way for that.

I also want to move all that Hover processing into the `HieExtras` module.

Issue #439, opened 2018-01-11
Matching hie GHC version to the project GHC version


alanz:
Come up with a process to launch the right hie, to match the compiler. And this has to work for stack and cabal, so stack install is not an option, necessarily.

I think there are two alternative approaches

1.    Make a concierge process that does the initial handshake until the initialize message with the project root is given, then detect the GHC version and spawn the appropriate hie version. The trick will be to replace the one process with the other, or to delegate stdio to the spawned hie. With the current state of initialization.

2.    Expose a helper that the client can invoke when determining the project root that also returns the hie instance to use.

The second option may be simpler, and can fall back to just using a setting in the client. These things tend to be pretty stable.

alanz:
Also, report an error if we discover that a project is using a mismatched GHC, via an LSP error message.

Tehnix:
As a temp fix, currently I'm using a wrapper script around HIE to make it start the right hie version,

```bash
# If a .stack-work exists, assume we are using stack.
DEBUG=0
function debug {
  if [[ $DEBUG == 1 ]]; then
    echo $@
  fi
}

GHCBIN='ghc'
if [ -d ".stack-work" ]; then
  GHCBIN='stack ghc --'
fi
versionNumber=`$GHCBIN --version`
debug $versionNumber

HIEBIN='hie'
BACKUP_HIEBIN='hie'
# Match the version number with a HIE version, and provide a fallback without
# the patch number.
if [[ $versionNumber = *"8.0.1"* ]]; then
  debug "It's was GHC 8.0.1"
  HIEBIN='hie-8.0.1'
  BACKUP_HIEBIN='hie-8.0'
elif [[ $versionNumber = *"8.0.2"* ]]; then
  debug "It's was GHC 8.0.2"
  HIEBIN='hie-8.0.2'
  BACKUP_HIEBIN='hie-8.0'
elif [[ $versionNumber = *"8.0"* ]]; then
  debug "It's was GHC 8.0.*"
  HIEBIN='hie-8.0'
elif [[ $versionNumber = *"8.2.1"* ]]; then
  debug "It's was GHC 8.2.1"
  HIEBIN='hie-8.2.1'
  BACKUP_HIEBIN='hie-8.2'
elif [[ $versionNumber = *"8.2.2"* ]]; then
  debug "It's was GHC 8.2.2"
  HIEBIN='hie-8.2.2'
  BACKUP_HIEBIN='hie-8.2'
[...]

PR #491, opened 2018-03-09
Use current directory as default when project root is not specified

ktonga:
Is using `$HOME` as default root dir the desired behavior? It doesn't seem to me like a sensible default to use. Even the CLI docs say the default is current directory

```
  -r,--project-root PROJECTROOT
                           Root directory of project, defaults to cwd
```

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -97,12 +97,7 @@ run opts = do
 
   origDir <- getCurrentDirectory
 
-  case projectRoot opts of
-    Nothing -> do
-      h <- getHomeDirectory
-      setCurrentDirectory h
-      logm $ "Setting home directory:" ++ h
-    Just root -> setCurrentDirectory root
+  maybe (pure ()) setCurrentDirectory $ projectRoot opts
 
   logm $  "run entered for HIE " ++ version
   d <- getCurrentDirectory


wz1000:
This seems fine, but hie doesn't really care about this directory, since it changes it anyway before running anything. 

https://github.com/haskell/haskell-ide-engine/blob/master/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs#L48

ktonga:
Ok. But I recall having to pass `-r` from vim in order for HIE to work.

ktonga:
I'm currently working on #439 and I'll need the root dir to be accurate before even starting ghc-mod in order to find project's ghc version using cabal-helper

alanz:
The determination of the GHC version needs to be done before the client launches the server.

So I would see this being useful by being able to do

`hie --project-root /home/foo/project1 --get-ghc-version`

Which would then not launch the LSP server, but instead find out what the version of GHC in use is, and return it.

This could then be called from the client.

We have a bit of a chicken and egg situation, where according to the LSP spec, the client sends the project root only in the `initialize` message, which can only be sent *after* the server is launched.

So at the moment the best thing seems to be to be able to ask hie via a specific command what GHC version is in use, so the client can choose an appropriate exe, and/or store that in a project-specific config.

And once we know how to determine the GHC version, we should send an error message back to the client when we get the `initialize` if there is a mismatch.

alanz:
And I will merge this, as it is a step in the right direction.

ktonga:
@alanz thanks for merging.

PR #498, opened 2018-03-14
Tentative in-document textDocument/references based on textDocument/documentHighlight

MaskRay:


alanz:
@MaskRay does this behave the way you expect it to? If so it looks good to me.

MaskRay:
It works. With the following hydra, I can jump to prev/next reference with `xk` `xj` (I unbind evil normal-state `x` to make it a cross reference prefix key)

```elisp
    (defhydra hydra/ref (evil-normal-state-map "x")
      "reference"
      ("d" lsp-ui-peek-find-definitions "next" :bind nil)
      ("k" (-let [(i . n) (lsp-ui-find-prev-reference)]
             (if (> n 0) (message "%d/%d" i n))) "prev")
      ("j" (-let [(i . n) (lsp-ui-find-next-reference)]
             (if (> n 0) (message "%d/%d" i n))) "next")
    )
```

alanz:
Thanks, you will see I merged it.  And added you as a contributor.

Now to find out what hydra is.

PR #501, opened 2018-03-22
Circle 2.0

alanz:
Use the CircleCI 2.0 API, allowing builds to run in parallel.

See https://circleci.com/workflow-run/3cc34799-f1af-4977-8549-e04d3331f1e9


nponeccop:
Note that there is an older attempt by me:

https://github.com/nponeccop/haskell-ide-engine/tree/master/.circleci

I'll review it and check if there are any features in my code worth integrating

PR #508, opened 2018-03-27
Read Brittany configuration

AlexeyRaga:
## Changes
- Use latest `Brittany`
- Respect `Brittany`'s config files (both global and local)

With this change users can update their "global" Brittany's config files (`~/.config/brittany/config.yaml`) and/or "local" (project level) config files (`brittany.yaml`) and this configuration will be taken into account.


alanz:
LGTM

PR #509, opened 2018-03-27
DRY CCI 2.0

AlexeyRaga:
Hi @alanz 

I see that you are working on CCI 2.0. Have a look, maybe you'll like this config?


alanz:
Thanks, I was hoping someone with more knowledge would do this for us.

PR #525, opened 2018-04-13
Fix extra newline being added when formatting document

### LINKED ISSUE #521: Formatting code adds newline at the end of the file
As the title reads, newlines are inserted at the end of the file each time I format the code. This problem occurs in VS Code and Emacs at least.

Bubba:
Fixes #521

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/Brittany.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Brittany.hs
@@ -46,8 +46,10 @@ brittanyCmd tabSize uri range =
                 l = length textLines
                 c = T.length $ last textLines
                 textLines = T.lines text
-                newTextWithoutNewline = T.init newText -- get rid of extraneous \n
-                textEdit = J.TextEdit (Range startPos endPos) $ T.init newTextWithoutNewline
+                truncatedNewText = if T.last newText == '\n'
+                  then T.init newText -- get rid of extraneous \n
+                  else newText
+                textEdit = J.TextEdit (Range startPos endPos) truncatedNewText
             return $ IdeResponseOk [textEdit]
 
 extractRange :: Range -> Text -> Text
--- a/src/Haskell/Ide/Engine/Plugin/Brittany.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Brittany.hs
@@ -42,7 +42,7 @@ brittanyCmd tabSize uri range =
                       (T.pack $ "brittanyCmd: " ++ unlines (map showErr err)) Null)
           Right newText -> do
             let startPos = Position 0 0
-                endPos = toPos (lastLine,1)
+                endPos = Position lastLine 0
                 {-
                 In order to replace everything including newline characters,
                 the end range should extend below the last line. From the specification:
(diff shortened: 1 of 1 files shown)


src/Haskell/Ide/Engine/Plugin/Brittany.hs
```diff
@@ -44,7 +46,8 @@ brittanyCmd tabSize uri range =
                 l = length textLines
                 c = T.length $ last textLines
                 textLines = T.lines text
-                textEdit = J.TextEdit (Range startPos endPos) newText
+                newTextWithoutNewline = T.init newText -- get rid of extraneous \n
```
wz1000: Wouldn't it be a better idea to remove all trailing newlines specifically instead of blindly truncating the text?

Bubba: Probably! Brittany seems to only remove the first trailing newline though, so should we try to match this?

lspitzner:
The use of `T.lines` above looks suspicious to me, btw. Afaict, `T.lines` interprets "newline" as purely "\n" and leaves "\r" in place. I have not considered/tested the exact consequences of this, but it seems not all unlikely to be behind this newline bug and also potentially other newline insertion problems that were reported for HIE.

@Bubba Unrelated: Under what circumstances does brittany remove trailing newlines?

alanz:
@lspitzner what do we do with this PR?

lspitzner:
I cannot reproduce the issue with `readFile`:

~~~~
> :!hexdump -C test.txt
00000000  66 69 72 73 74 0d 0a 73  65 63 6f 6e 64           |first..second|
0000000d
> TIO.readFile "test.txt"
"first\r\nsecond"
~~~~

I also point out that it is really hard to tell from

> ~~~~
> > cat test
> hello world
> this is a test with no newline
> ~~~~

whether the file actually terminates with a newline. It really should look like this:

~~~~
> cat test
hello world
this is a test with no newline> [cursor here]
~~~~

lspitzner:
As this might be system/locale dependent, I should add that I have tested on linux only, with "\n" and "\r\n". @Bubba what environment are you seeing this in?

Bubba:
Sorry, disregard what I said about `T.readFile`! I was naïve to think that echoing to a file wouldn't add a newline. I'm on macOS 10.13. I will try to figure out more about what is exactly causing this.

Bubba:
What I'm now convinced was happening is that the range of text being replaced didn't include the last newline character, which meant that it stayed around after the formatted text was inserted, adding an extra line. [From the specification:](https://microsoft.github.io/language-server-protocol/specification#range)
> If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line.

Although I found that the end range column needs to be past zero, contradictory to above. This was tested on VS Code.

wz1000:
The Position type is zero indexed(as specified by LSP), but the tuple (Int,Int) is assumed to always be one indexed in HIE(as GHC reports positions as one indexed tuples). `toPos :: (Int, Int) -> Position` converts between these two representations. So

    toPos (1,1) == Position 0 0

The mixing of types here is unfortunate(probably my fault, and how this bug originated). It would be better to use one of these position types consistently here.

Otherwise it looks good.

PR #526, opened 2018-04-16
Switch to using submodules for external repos

alanz:
These can be set up once, and used for all builds, instead of needing
to be set per stack file, and in the cabal.project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,10 @@ defaults: &defaults
   steps:
     - checkout
     - run:
+        command: git submodule sync --recursive
+    - run:
+        command: git submodule update --recursive --init
+    - run:
         name: Write provided stack.yaml with predictable name
         command: cp ${STACK_FILE} stack-build.txt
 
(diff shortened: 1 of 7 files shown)


domenkozar:
Ugh, this will make hie-nix harder to setup, but it's not impossible :)

alanz:
I'm sort of in two minds about this, but keeping hashes lined up between `n` different stack.yaml files and a Makefile (for cabal new-build) gets a bit tiresome after a while.

And submodules are a known thing, the alternative is to make a custom something to do basically the same.

But this PR is more of a discussion point than anything else, as it is quite a big change in some ways.

domenkozar:
Maybe just a comment in both places would suffice so commits are in sync? How often are those files updated?

alanz:
Pretty much all the time at the moment, as we try to stabilise cabal-helper and ghc-mod.

alanz:
@domenkozar How bad would it be for nix?

alanz:
I am once again suffering from slow pipes, and finding if frustrating to try and sort out repo versions, with a fresh 40MB clone of HaRe each time for stack.  So I am going to merge this.

PR #530, opened 2018-04-20
Implement case slitting through ghc-mod

### LINKED ISSUE #259: Features request : case split, make definition.
I really like idris language server. 

Especially the *case split* feature.

```
map2 :: (a -> b -> c) -> [a] -> [b] -> [c]
map2 f x y = undefined
```

Right click  x and choose *case split*.

```
map2 f [] y = undefined
map2 f (x:xs) y = undefined
```

*Now lets forget about \_|\_ / bottom / undefined for a moment.*

What can we do with the function `f`, if we don't have any variable of type `a`? Nothing... lets get rid of that.

```
map2 _ [] _ = []
map2 f (x:xs) y = undefined
```

----------

And also the *named hole* feature.

```
sum :: [Int] -> Int
-- we do not get compile error here, ?foo acts like undefined
-- you can hover over ?foo and see its type
sum = ?foo (+) 0 
```

Right click `?foo` and select *make definition*.

```
sum :: [Int] -> Int
sum = foo (+) 0

foo :: (Int -> Int -> Int) -> Int -> [Int] -> Int
foo a b c = undefined
```

txsmith:
This implements a command that calls ghc-mod's [splits](https://hackage.haskell.org/package/ghc-mod-5.8.0.0/docs/GhcMod-Exe-CaseSplit.html) and parses the output into a `WorkspaceEdit`.

~~There's quite some code involved in parsing the output from ghc-mod. An example of the ouput:
`11 0 11 13 "f (Just x) = \"Hello\"\NULf Nothing = \"Bye\""`
This indicates the range of text (start line, start column, end line and end column) to replace and the new text to replace it with. Newlines are indicated by `\NUL` as per the default ghc-mod settings.
If there's a better way to parse this and turn it into a `WorkspaceEdit` with less code I'd love to know!~~
Now uses the new ghc-mod api! 

I've added tests similar to those for "insert type". I'll also open a PR on [vscode-hie-server](https://github.com/alanz/vscode-hie-server) so you can try it there for real.

Let me know what you think! I'm happy to follow up with improvements if needed.

Fixes #259


wz1000:
Thanks¸ this looks good.

However, the way ghc-mods splits function works, it forces reparsing and retypechecking of the module. But HIE has a module cache with the TypecheckedModule already on hand. 

If [`getSrcSpanTypeForFnSplit`](https://github.com/alanz/ghc-mod/blob/47e200a728a575f407ee6f9893d9a1e77b1b5325/GhcMod/Exe/CaseSplit.hs#L88) is modified to accept a typechecked module, we could hand it the one we already have in HIE, making the whole process much, much faster. 

Also, it would eliminate the need for parsing as you would get the SplitInfo data structure directly.

txsmith:
Thanks, good point. Seems doable so I'll see if I can get something going there.

alanz:
I am about to merge this, there is a minor fixup required for master, the name of one of the tests changed.

alanz:
@txsmith  I am getting the following when testing with GHC 8.4.2 (the default stack.yaml), in my branch https://github.com/alanz/haskell-ide-engine/tree/case-split-259

```
      runs the lint command
      runs the info command
      runs the type command
      runs the type command with an absolute path from another folder, correct params
EXCEPTION: splits:     Prelude.last: empty list
EXCEPTION: splits:     Prelude.last: empty list
      runs the casesplit command FAILED [1]
EXCEPTION: splits:     Prelude.last: empty list
EXCEPTION: splits:     Prelude.last: empty list
      runs the casesplit command with an absolute path from another folder, correct params FAILED [2]
HaRePlugin
```

So I am going to hand this back, please fix it. Sorry.

txsmith:
I'll take a look. I was also still updating this to use the new `split'` in your ghc-mod fork so this is not really ready to be merged just jet ;)

alanz:
Ok, thanks. I felt bad about forcing you to rebase the whole time, thought I would try a quick fixup and land.

txsmith:
I probably made a mistake somewhere. Will investigate more tomorrow.

txsmith:
See alanz/ghc-mod#10

wolverian:
Did https://github.com/alanz/ghc-mod/pull/11 fix the issue?

alanz:
I have not really paid attention to what this PR does, so I do not know.

I did see a trivial error, which was fixed in ghc-mod, but it seems there may be other issues too?

wolverian:
Right, I meant @txsmith, sorry.

txsmith:
I believe I fixed the issue in alanz/ghc-mod#10. The problem was that case splitting never worked in ghc-mod for newer GHC versions in the first place.
Then there was also the trivial mistake that was fixed in alanz/ghc-mod#11

alanz:
Ok, So I will merge https://github.com/alanz/ghc-mod/pull/10 and we can see about getting this in.

Sorry to mess you around so much.

alanz:
@txsmith  Current hie uses ghc-mod with your patch, so you can get this one finished,

txsmith:
I think this is ready for another review now!

src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -290,6 +293,66 @@ cmp a b
 isSubRangeOf :: Range -> Range -> Bool
 isSubRangeOf (Range sa ea) (Range sb eb) = sb <= sa && eb >= ea
 
+
+
+data DocumentPosition =
```
alanz: Shouldn't this be an instance of `TextDocumentPositionParams`

```typescript
interface TextDocumentPositionParams {
	/**
	 * The text document.
	 */
	textDocument: TextDocumentIdentifier;

	/**
	 * The position inside the text document.
	 */
	position: Position;
}
```

as per https://microsoft.github.io/language-server-protocol/specification

This is made available via `haskell-lsp-types`, see https://github.com/alanz/haskell-lsp/blob/406b9e0e795b5e9c2ae4a2ab29b2339c29fa0354/haskell-lsp-types/src/Language/Haskell/LSP/TH/DataTypesJSON.hs#L1008

src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -290,6 +293,66 @@ cmp a b
...
+  toJSON = genericToJSON $ defaultOptions { fieldLabelModifier = camelTo2 '_' . drop 3}
+
+splitCaseCmd ::CommandFunc DocumentPosition WorkspaceEdit
+splitCaseCmd = CmdSync $ \(DocPos uri pos) -> do
+  splitCaseCmd' uri pos
+
+splitCaseCmd' :: Uri -> Position -> IdeGhcM (IdeResult WorkspaceEdit)
+splitCaseCmd' uri position =
+  pluginGetFile "splitCaseCmd: " uri $ \path -> do
+    origText <- GM.withMappedFile path $ liftIO . T.readFile
```
wz1000: This version of the file may differ from the version we have a TypecheckedModule for. Use `newPosToOld` and `oldPosToNew` to correlate positions across versions of the file.

src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -290,6 +293,66 @@ cmp a b
...
+    cachedMod <- getCachedModule path
+    case cachedMod of
+      ModuleCached checkedModule _ -> runGhcModCommand $ do
+        Just splitResult <- GM.splits' path (tcMod checkedModule) line column
```
wz1000: A pattern match failure may cause HIE to crash

src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -290,6 +293,66 @@ cmp a b
...
+        return $ splitResultToWorkspaceEdit origText splitResult
+      ModuleFailed errText -> return $ IdeResultFail $ IdeError PluginError (T.append "hie-ghc-mod: " errText) Null
+      ModuleLoading -> return $ IdeResultOk mempty
```
wz1000: You should use `withCachedModule`, as that will automatically defer the action if the module is still loading.

Alternatively, you can defer it manually over here.

txsmith: I've looked at this but could not find a nice way to defer the action while keeping it in `IdeGhcM`. To defer an `IdeGhcM` action we'd need to use `IdeResponse'`, both of which do not work with `withCachedModule`. Furthermore, in the end all plugins seem to use `CommandFunc` which expects to return an `IdeResult`. I can't really find a sensible way to make these work together without creating a mess. Also note that other plugins like `typeCmd` don't defer either. Am I missing something?

txsmith:
Thanks for your feedback! Finally had some time to get it fixed. Please take a look :)

wz1000:
Looks good. Thanks a lot.

PR #535, opened 2018-04-25
Update README.md

liang-ting-chen:
Add option "--recursive" in the installation guide due to the use of submodules


wz1000:
Thanks

PR #536, opened 2018-04-25
GHC 8.4 support

alanz:
This builds, and passes the tests we have.

I have not tried it on an actual project yet.

Use at your own risk, and provide feedback please.


alanz:
This is ready to land.

A side-effect of cabal-helper is that it can no longer support projects for cabal < 2.0 (via the `cabal-version` field in the relevant cabal file.

This means we cannot use stack to build it for GHC 8.0.2, although `cabal new-build` using cabal 2.2 does work.  But in that case `cabal-helper` needs to be installed via plain `cabal install` too, else the `libexec` dir is not resolved.

patrickt:
Very happy to see this! My team can upgrade to GHC 8.4 now 😁

PR #547, opened 2018-05-04
Fix type in hie-8.4.2 build target

Anrock:


Anrock:
Oh irony. Typo in commit message about fixing typo.

PR #548, opened 2018-05-06
Don't provide code actions for diagnostics that hlint doesn't have refactorings for

Bubba:
This changes the code action request reactor to send a request to hlint.
It can use this information to determine if any of the diagnostics coming from the LSP client are actionable, and filter out the ones that aren't.

Fixes alanz/vscode-hie-server#61


wz1000:
Looks good to me

Issue #551, opened 2018-05-08
Incorrect file uri


Leonti:
Hi!
First of all, thanks for the great IDE engine, I believe this will be the future of Haskell IDE tooling.  
I have a problem though, my project compiles and works fine, but I get errors coming from `/private/var/folders/...` folder, for example:  
```
2018-05-0813:56:23.784628[
   ThreadId 5
]- <--2--{
   "jsonrpc":"2.0",
   "method":"textDocument/publishDiagnostics",
   "params":{
      "uri":"file:///private/var/folders/hm/3t_b50tj3ng1fj3g4p9zgcn02hg99c/T/ghc-mod12335/PayslipCalculatorSpec12334-0.hs",
      "diagnostics":[
         {
            "severity":1,
            "range":{
               "start":{
                  "line":4,
                  "character":0
               },
               "end":{
                  "line":4,
                  "character":33
               }
            },
            "source":"ghcmod",
            "message":"Could not find module ‘PayslipArbitrary’\nUse -v to see a list of the files searched for."
         },
         {
            "severity":1,
            "range":{
               "start":{
                  "line":6,
                  "character":0
               },
               "end":{
                  "line":6,
                  "character":42
               }
            },
            "source":"ghcmod",
            "message":"Could not find module ‘PayslipCalculatorFixtures’\nUse -v to see a list of the files searched for."
         },
         {
            "severity":1,
            "range":{
               "start":{
                  "line":7,
                  "character":0
               },
               "end":{
                  "line":7,
                  "character":27
               }
            },
            "source":"ghcmod",
            "message":"Could not find module ‘Test.Hspec’\nUse -v to see a list of the files searched for."
         },
         {
            "severity":1,
            "range":{
               "start":{
                  "line":8,
                  "character":0
               },
               "end":{
                  "line":8,
                  "character":38
               }
            },
            "source":"ghcmod",
            "message":"Could not find module ‘Test.Hspec.QuickCheck’\nUse -v to see a list of the files searched for."
         },
         {
            "severity":1,
            "range":{
               "start":{
                  "line":9,
                  "character":0
               },
               "end":{
                  "line":9,
                  "character":32
               }
            },
            "source":"ghcmod",
            "message":"Could not find module ‘Test.QuickCheck’\nUse -v to see a list of the files searched for."
         }
      ]
   }
}
```  

[...]

PR #554, opened 2018-05-09
Start canonicalizing mapped files again

wz1000:
Should fix https://github.com/haskell/haskell-ide-engine/issues/551 and https://github.com/alanz/vscode-hie-server/issues/80


alanz:
Thanks.   Sorry for causing this regression.

PR #564, opened 2018-05-12
Make symbol requests check the module if no cache is available

Bubba:
Currently if you make a document symbol request as soon as the server starts, you get an empty list as a response, usually since there is no cached module available.

This was causing problems with [noughtmare/haskell-lsp-client](https://github.com/noughtmare/haskell-lsp-client/tree/233eec4515844e3ec22ff2ec5cb28d565061dfc5), so this PR forces the module to be checked whenever there is no cache available to use during the request.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -100,91 +100,100 @@ instance ModuleCache NameMapData where
 
 -- ---------------------------------------------------------------------
 
-getSymbols :: Uri -> IdeM (IdeResponse [J.SymbolInformation])
+getSymbols :: Uri -> IdeGhcM (IdeResponse [J.SymbolInformation])
 getSymbols uri = pluginGetFile "getSymbols: " uri $ \file -> do
     mcm <- getCachedModule file
     case mcm of
-      Nothing -> return $ IdeResponseOk []
-      Just cm -> do
-          let tm = tcMod cm
-              rfm = revMap cm
-              hsMod = unLoc $ pm_parsed_source $ tm_parsed_module tm
-              imports = hsmodImports hsMod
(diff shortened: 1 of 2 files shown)


wz1000:
This doesn't look right. If we don't have a cached module when a symbol request is made, either the module is still loading and the IdeGhcM dispatcher thread is busy compiling it, or the module couldn't be loaded/failed to compile. In both cases, it doesn't make sense to try loading it again.

We definitely don't want `getSymbols :: Uri -> IdeGhcM`. The whole point of the `IdeM`/`IdeGhcM` split is so that HIE can still respond to requests when the dispatcher for `IdeGhcM` is busy compiling stuff(the most expensive and slow operation that HIE executes). Changing `getSymbols` to `IdeGhcM` means that HIE will now block and not answer symbol requests until it has finished compiling stuff.

wz1000:
If I am guessing correctly, you are trying to test symbol request, but HIE is returning an empty response early, so the test is failing.

As an alternative, consider adding a mechanism to HIE by which it can signal module loading status, and whether a module has been loaded/is loading/failed to load. Then you can wait for HIE to signal its status before trying to send a symbol request

I am in favour of closing this pull request as it removes functionality from HIE and causes a performance regression.

Bubba:
Ah, I see it might make more sense to not try loading the module a second time and just wait for the IdeGhcM thread to finish what it's doing.
But would there be much of a difference returning nothing and then making the client remake the request, rather than blocking the symbol request until the module is checked before returning the result? As long as this only blocks the symbol requests and not anything else

Bubba:
Right, I see the issue now. Could `findDef`/definition requests also be edited to work in the same way?

wz1000:
In principle, findDef could work this way, but it doesn't currently because it needs to make some ghc api requests(which are only possible in IdeGhcM). It would be nice if someone could figure out how to make even findDef non-blocking.

PR #568, opened 2018-05-13
[WIP] Add a queue for module caches

Bubba:
This allows for requests that depend on a cached module to receive a callback whenever the cached module is ready. As an example, if a document symbols request is made right as HIE is started, it will no longer immediately return an empty list but will now return the correct response as soon as the module has been loaded.

It adds an extra field to `IdeState` to keep track of which modules have which actions queued, and changes getCachedModule to return more information about if it's stale, loading or failed. 

- [ ] Figure out how to tell if module loading failed?
- [ ] Handle JSON of IdeDeferredResponse
- [ ] Bring back discarded ExceptT transformers in the form of IdeResponseT
- [ ] Go through for `error "impossible"`s and add matches for IdeDeferredResponse

### DIFF
--- a/src/Haskell/Ide/Engine/Dispatcher.hs
+++ b/src/Haskell/Ide/Engine/Dispatcher.hs
@@ -49,9 +49,28 @@ ideDispatcher env pin = forever $ do
   (IdeRequest lid callback action) <- liftIO $ atomically $ readTChan pin
   debugm $ "ideDispatcher:got request with id: " ++ show lid
   cancelled <- liftIO $ atomically $ isCancelled env lid
+  res <- action
   unless cancelled $ do
-    res <- action
-    liftIO $ callback res
+    case res of
+      IdeResponseDeferred uri cacheCb -> do
+        queueActionForModule uri \cm -> do
+          cacheRes <- cacheCb cm
+          liftIO $ callback cacheRes
+      _ -> liftIO $ callback res
+
+
+-- | Queues an aciton to be run after the module has been loaded
+queueActionForModule :: FilePath -> (CachedModule -> IdeM ()) -> IdeM ()
+queueActionForModule fp action = do
+  fp' <- liftIO $ canonicalizePath fp
+  modifyMTS $ \s ->
+    let oldQueue = actionQueue s
+    in  s
+          { actionQueue = if Map.member fp' oldQueue
+                            then Map.update (Just . (action :)) fp oldQueue
+                            else Map.insert fp [action] oldQueue
+          }
+    
 
 ghcDispatcher :: forall void. DispatcherEnv -> TChan GhcRequest -> IdeGhcM void
 ghcDispatcher env@DispatcherEnv{docVersionTVar} pin = forever $ do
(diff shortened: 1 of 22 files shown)


src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -100,91 +98,100 @@ instance ModuleCache NameMapData where
...
-              declsToSymbolInf :: (J.SymbolKind, Located T.Text, Maybe T.Text)
-                               -> IdeM (Either T.Text J.SymbolInformation)
-              declsToSymbolInf (kind, L l nameText, cnt) = do
-                eloc <- srcSpan2Loc rfm l
-                case eloc of
-                  Left x -> return $ Left x
-                  Right loc -> return $ Right $ J.SymbolInformation nameText kind loc cnt
-          symInfs <- mapM declsToSymbolInf (imps ++ decls)
-          return $ IdeResponseOk $ rights symInfs
+getSymbols :: Uri -> (IdeResponse [J.SymbolInformation] -> IdeM ()) -> IdeM ()
```
Bubba: @wz1000 I'm not the biggest fan of this callback part. Is there a better place/way to do this?

wz1000: working on something

wz1000:
For the monad instance

    IdeResponseT ≅FreeT IdeResponseF
    data IdeResponseF a = IdeError Error
                        | IdeResponseDeferred FilePath (CachedModule -> a)

https://hackage.haskell.org/package/free-5.0.1/docs/Control-Monad-Trans-Free.html#t:FreeT

alanz:
What is the status of this?

Bubba:
@alanz Still need to add the transformer in @wz1000's comment, and there's a couple of problems I need to address:

- [ ] `IdeResponse`'s are usually serialised, so will need to make a wrapper or find a way to prevent `IdeResponseDeferred` from being serialized
- [ ] Need to double check areas where `runExceptT` were used will still work

I think I will return to it this week since I'm happy with the progress I've made with haskell-lsp-test.

src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -49,9 +51,30 @@ ideDispatcher env pin = forever $ do
   (IdeRequest lid callback action) <- liftIO $ atomically $ readTChan pin
   debugm $ "ideDispatcher:got request with id: " ++ show lid
   cancelled <- liftIO $ atomically $ isCancelled env lid
-  unless cancelled $ do
-    res <- action
-    liftIO $ callback res
+  res <- action
```
wz1000: This looks wrong- we definitely should not be running actions that have been cancelled.

src/Haskell/Ide/Engine/Types.hs
```diff
@@ -32,6 +32,6 @@ data GhcRequest = forall a. GhcRequest
 
 data IdeRequest = forall a. IdeRequest
   { pureReqId :: J.LspId
-  , pureReqCallback :: a -> IO ()
-  , pureReq :: IdeM a
+  , pureReqCallback :: IdeResponse a -> IO ()
```
wz1000: I think it would be a good idea to have the type of the callback be `a -> IO ()` and we let the dispatcher handle error/deferred responses.

Issue #574, opened 2018-05-18
Makefile broken on macos


newhoggy:
This change from `2018-05-16` broke the `Makefile` my macos environment:

```
$ git blame Makefile | head -n 3
5370b013 (Christian Kjaer Laustsen 2018-01-24 11:02:08 +0100  1) BASEDIR=$(CURDIR)
9691675e (Vladislav Shtepin        2018-05-16 01:02:42 +0300  2) STACKLOCALBINDIR != stack path --local-bin
5370b013 (Christian Kjaer Laustsen 2018-01-24 11:02:08 +0100  3)
```

This is the error I get:

```
cp: /hie: No such file or directory
```

After patching this line, the build works again:

```
diff --git a/Makefile b/Makefile
index 85f4159..b73266c 100644
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BASEDIR=$(CURDIR)
-STACKLOCALBINDIR != stack path --local-bin
+STACKLOCALBINDIR:=$(shell stack path --local-bin)

 build: hie-8.2.1 hie-8.2.2 hie-8.4.2
 # build: hie-8.0.2 hie-8.2.1 hie-8.2.2 hie-8.4.2
```

Tehnix:
Just tried the `make build` (I usually do `make build-copy-compiler-tool`), and it seems https://github.com/haskell/haskell-ide-engine/commit/9691675e6537c8d8e56fc9abeb5baa711ee422f8#diff-b67911656ef5d18c4ae36cb6741b7965 broke support for macOS at least (don't have linux at hand right now).

The problem is that `STACKLOCALBINDIR != stack path --local-bin` does not return anything on macOS, but the patched command fixes this `STACKLOCALBINDIR:=$(shell stack path --local-bin)`.

Does `STACKLOCALBINDIR:=$(shell stack path --local-bin)` also work on Linux? Cause then I would say we should switch to this.

Anrock:
I can confirm that `STACKLOCALBINDIR:=$(shell stack path --local-bin)` works at Windows atleast. 
Sorry for breaking things for MacOs.

Tehnix:
No worries at all! I think macOS is on an older version of make or something.

@Anrock would you be up for making a PR with this change, if you have time? :)

Anrock:
@Tehnix 
> No worries at all! I think macOS is on an older version of make or something.
[...]

PR #576, opened 2018-05-18
call make in cci

AlexeyRaga:
## Changes
- Call `make` in CCI to ensure that the makefile works.

This is to address #574 but as far as I can see Linux builds are not affected while MacOS builds are.

Unfortunately, CircleCI doesn't give us MacOS builds on free plans :(

I can implement Travis builds to run MacOS build(s), but it will happen not earlier than in 1 or 2 weeks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,26 +74,34 @@ defaults: &defaults
           - ~/build/.stack-work
           - ~/build/hie-plugin-api/.stack-work
 
+    - run:
+        name: Making sure that 'make' works
+        command: make ${HIE_VER}
+
 version: 2
 jobs:
   ghc-8.0.2:
     environment:
       - STACK_FILE: "stack-8.0.2.yaml"
+      - HIE_VER: "hie-8.0.2"
     <<: *defaults
 
   ghc-8.2.1:
     environment:
       - STACK_FILE: "stack-8.2.1.yaml"
+      - HIE_VER: "hie-8.2.1"
     <<: *defaults
 
   ghc-8.2.2:
     environment:
       - STACK_FILE: "stack-8.2.2.yaml"
+      - HIE_VER: "hie-8.2.2"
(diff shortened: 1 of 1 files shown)


newhoggy:
It might be possible to CCI against an older version of `gmake`

```
$ gmake --version
GNU Make 4.2.1
Built for x86_64-apple-darwin17.0.0
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
 ~/wrk/arbor-networks/haskell-ide-engine   master 
$ make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

AlexeyRaga:
Oh, if a brewed `gmake` works then perhaps updating README suggesting `brew install make` on MacOS  would be enough?

Or/And is it possible to check min `make` version in the Makefile itself?

PR for any of these @newhoggy ? ;)

alanz:
@AlexeyRaga @newhoggy ping

AlexeyRaga:
@alanz I've totally forgotten about this issue. I'll check it out in a couple of days when I return back to work.

newhoggy:
Looks like a good start.  Is there a way to fix the make version at this?

```
$ make --version
GNU Make 3.81
```

newhoggy:
Otherwise a note in the `README` telling users to upgrade `make` with `brew` is also good.

AlexeyRaga:
Not to me. I don't have any problems with making HIE now.

PR #588, opened 2018-05-21
Canonicalize uri for dummy diags

### LINKED ISSUE #560: [Win10] Diagnostics are shown only for a split-second
[Full hie log](https://bpaste.net/show/e468b717796f)
`HIE Version 0.2.0.0, Git revision 4e900d53c401c0d3b6f69c6bbddc9ab6c6834d0d (dirty) (1338 commits) x86_64 ghc-8.2.2`

I guess here is the issue:
```
2018-05-10 23:13:36.6195978 [ThreadId 4] - <--2--{"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///C%3A/Users/Anrock/Nextcloud/Projects/nand2tetris/projects/hack-translator/Main.hs","diagnostics":[{"severity":1,"range":{"start":{"line":14,"character":61},"end":{"line":14,"character":68}},"source":"ghcmod","message":"Variable not in scope: outName :: FilePath"}]}}
2018-05-10 23:13:36.6195978 [ThreadId 4] - <--2--{"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///C:/Users/Anrock/Nextcloud/Projects/nand2tetris/projects/hack-translator/Main.hs","diagnostics":[]}}
```

Diagnostics are shown due to first message and then almost immediately cleared by second message. Not sure if there any more info needed?

Anrock:
`getTypecheckedModuleGhc'` returns diags with canonicalized URIs so dummy diags with un-canonicalized URI is always inserted to diags map.
Which leads to LSP client overwriting actual diags with dummy ones because both URI are treated the same.

Fixes #560

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
+++ b/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
@@ -157,7 +157,8 @@ setTypecheckedModule uri =
     debugm $ "setTypecheckedModule: file mapping state is: " ++ show fileMap
     rfm <- GM.mkRevRedirMapFunc
     ((diags', errs), mtm) <- GM.getTypecheckedModuleGhc' (myLogger rfm) fp
-    let diags = Map.insertWith Set.union uri Set.empty diags'
+    canonUri <- canonicalizeUri uri
+    let diags = Map.insertWith Set.union canonUri Set.empty diags'
     case mtm of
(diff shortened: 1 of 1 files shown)


wz1000:
Looks good.

PR #590, opened 2018-05-21
Remove dead code from findDef and make it async

wz1000:
Can anyone try this out and confirm that findDef doesn't fail on any cases it handled before this commit?


alanz:
Wow, that's a lot of dead code going. Great.

PR #594, opened 2018-05-24
Add deferred responses and move error/callback handling into dispatcher

Bubba:
Currently, if a document symbol request is made just as the server is started, it returns an empty list, since the module is not loaded yet.
We would like it to wait until the module is loaded before sending a response. To do this, an `IdeResponseDeferred` type has been added.

This type allows the request callback to be deferred until a module at a specific file path has been loaded. A queue is kept that keeps track of it.

```haskell
data IdeState = IdeState
  { moduleCache :: GhcModuleCache
  -- | A queue of requests to be performed once a module is loaded
  , requestQueue :: Map.Map FilePath [(LspId, CachedModule -> IdeM ())]
  , idePlugins  :: IdePlugins
  , extensibleState :: !(Map.Map TypeRep Dynamic)
  , ghcSession  :: Maybe (IORef HscEnv)
  }
```

`getCachedModule` also returns more information now:
```haskell
-- | The possible states the cache can be in
-- along with the cache or error if present
data CachedModuleResult = ModuleLoading
                        | ModuleFailed IdeError
                        | ModuleCached CachedModule IsStale
type IsStale = Bool

-- | looks up a CachedModule for a given URI
getCachedModule :: (GM.MonadIO m, HasGhcModuleCache m)
                => FilePath -> m CachedModuleResult
```

`IdeResponse` has also been split into two separate types:
- `IdeResponse` can either hold an `IdeResult` or a deferred response. This is meant to be used by `IReqs`, i.e. requests that go through `IdeM`, since they might be deferred if they depend on a module that's being loaded in `IdeGhcM`.
- `IdeResult` is either a success or failure. `IdeGhcM` usually returns this since it will never need to be deferred.

```haskell
data IdeResult a = IdeResultOk a
                 | IdeResultFail IdeError
data IdeResponse m a = IdeResponseDeferred FilePath (CachedModule -> m (IdeResponse m a))
                     | IdeResponseResult (IdeResult a)
```

The dispatcher has also been changed so that it now handles the unwrapping of `IdeResponse` and `IdeResults`. It now takes in a handler for errors and callbacks so that `LspStdio` and `JsonStdio` can deal with them separately. This means that the reactor doesn't need to perform error handling anymore, and handlers can be written without `hieResponseHelper`:
```haskell
Core.ReqDocumentFormatting req -> do
        let params = req ^. J.params
            doc = params ^. J.textDocument . J.uri
            tabSize = params ^. J.options . J.tabSize
            callback textEdit = do
              let rspMsg = Core.makeResponseMessage req $ J.List textEdit
              reactorSend rspMsg
        let hreq = GReq (Just doc) Nothing (Just $ req ^. J.id) callback
                     $ Brittany.brittanyCmd tabSize doc Nothing
        makeRequest hreq
```

## Todo
- [x] Handle cancellation properly


src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -43,18 +66,39 @@ mainDispatcher inChan ghcChan ideChan = forever $ do
...
+          handleResponse lid actualCb cacheResponse
+    
+        queueAction :: J.LspId -> FilePath -> (CachedModule -> IdeM ()) -> IdeM ()
+        queueAction lid fp action = do
+          fp' <- liftIO $ canonicalizePath fp
+          modifyMTState $ \s ->
+            let newReq   = (lid, action)
+                oldQueue = requestQueue s
+                -- add to existing queue if possible
+                newQueue = if Map.member fp' oldQueue
```
alanz: I am sure this if member stuff can be simplified, by using Map.adjust, or one of the other Map update options

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -124,28 +132,34 @@ withCachedModuleAndData uri noCache callback = do
...
-               modifyCache (\s -> s {uriCaches = Map.insert uri' (UriCache cm dat')
+               modifyCache (\s -> s {uriCaches = Map.insert uri' (UriCache cm dat' False)
                                                                  (uriCaches s)})
                return val
-             Just x -> do
+             Just x ->
                case fromDynamic x of
                  Just val -> return val
                  Nothing  -> error "impossible"
       callback cm a
```
wz1000: Shouldn't this mirror the changes in withCachedModule?

Bubba: Updated this to remove noCache and used `IdeResponseDeferred` instead

src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -72,14 +72,20 @@ ideDispatcher env errorHandler callbackHandler pin = forever $ do
...
       
-  where handleResponse lid callback response = case response of
-          IdeResponseResult (IdeResultOk x) -> liftIO $ callbackHandler callback x
-          IdeResponseResult (IdeResultFail err) -> liftIO $ errorHandler lid err 
-          IdeResponseDeferred fp cacheCb -> handleDeferred lid fp cacheCb callback
+  where handleResponse lid callback response = do
+          cancelled <- liftIO $ atomically $ isCancelled env lid
+          if cancelled
+            then liftIO $ errorHandler lid J.RequestCancelled ""
+          else case response of
```
alanz: This if statement is just re-doing the logic above, wrt cancelling. This function should just be the else leg

stack.yaml
```diff
@@ -28,8 +28,13 @@ extra-deps:
...
 - haddock-library-1.6.0
-- haskell-lsp-0.2.2.0
-- haskell-lsp-types-0.2.2.0
+# - haskell-lsp-0.2.2.0
+# - haskell-lsp-types-0.2.2.0
+- github: alanz/haskell-lsp
+  commit: 82a5128e66ea8e4a82fa96826eca2d94f729d8c5
+  subdirs:
+  - .
+  - haskell-lsp-types
```
alanz: Interesting, did not realise this syntax was a thing

alanz: Except we are preferring to use submodules for this, else we have to update every stack file, and the cabal build does not work.

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -77,75 +74,92 @@ getCradle fp = do
...
-                          (UriCache cm Map.empty)
+                          (UriCache cm Map.empty False)
                           (uriCaches gmc)
           }
     )
-  where
+
+  -- execute any queued actions for the module
+  actions <- fmap (fromMaybe [] . Map.lookup uri') (requestQueue <$> readMTS)
+  liftToGhc $ forM_ actions (\(_, a) -> a cm)
```
wz1000: Shouldn't we be checking if the action has been cancelled before executing it?

src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -43,59 +65,92 @@ mainDispatcher inChan ghcChan ideChan = forever $ do
...
+checkCancelled :: MonadIO m => DispatcherEnv -> J.LspId -> ErrorHandler -> m () -> m ()
+checkCancelled env lid errorHandler callback = do
+  cancelled <- liftIO $ atomically isCancelled
+  if cancelled
+    then liftIO $ do
+      -- remove from cancelled and wip list
+      atomically $ modifyTVar' (cancelReqsTVar env) (S.delete lid)
+      completedReq env lid
+      errorHandler lid J.RequestCancelled ""
+    else callback
```
wz1000: If it is not cancelled, we still need to remove it from wipReqs.

Bubba: Non-cancelled requests get removed from wipReqs outside of `checkCancelled`, since we check for cancellation before executing the request and again after executing it
https://github.com/Bubba/haskell-ide-engine/blob/edc5abe06dde60eee11a23b7008d17288ec4bd4f/src/Haskell/Ide/Engine/Dispatcher.hs#L81-L86

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
Bubba: I believe this is checked in `Dispatcher.hs` https://github.com/Bubba/haskell-ide-engine/blob/edc5abe06dde60eee11a23b7008d17288ec4bd4f/src/Haskell/Ide/Engine/Dispatcher.hs#L80

wz1000: Then we don't need to have the LspId in the deferred request queue, right?

wz1000:
Looks good to me. We can merge if @Bubba is happy with it.

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
Bubba: Would seem so! Removing now

PR #600, opened 2018-05-28
Wrap exception handler around module loading.

alanz:
So that instead of a silent hang, we get an actuall error message
back, and operation continues.


alanz:
Note, @DanielG informs me that we should not be catching `SomeException`. So I guess we need to put a subset list in there instead.

alanz:
See https://hackage.haskell.org/package/base-4.11.1.0/docs/Control-Exception.html#g:4

wz1000:
`ghc` throws `SourceError`, `GhcApiError` and `IOEnvFailure`. `ghc-mod-core` throws `GhcModError`.

alanz:
FYI GHC also throws `GhcException`, which is the one that I have been triggering.

PR #612, opened 2018-06-02
Add session capturing

Bubba:
Adds the option to capture a session for testing later by passing the `-c` flag along with a file path:

`hie -c session.log`

Also begins migrating the functional tests to use `haskell-lsp-test`

Mirrors https://github.com/alanz/haskell-lsp/pull/87, and closes https://github.com/Bubba/haskell-lsp-test/issues/5.

### DIFF
--- a/stack-8.2.1.yaml
+++ b/stack-8.2.1.yaml
@@ -43,6 +43,7 @@ extra-deps:
 - github: Bubba/haskell-lsp-test
   commit: ac2553b38ed2228839cb72747f8c48b7c2fab488
 - conduit-parse-0.2.1.0
+- conduit-1.3.0.2
 
 flags:
   haskell-ide-engine:
--- a/stack-8.2.1.yaml
+++ b/stack-8.2.1.yaml
@@ -25,6 +25,10 @@ packages:
     - .
     - haskell-lsp-types
 
+- location:
+    ./submodules/haskell-lsp-test
+  extra-dep: true
+
 extra-deps:
 - brittany-0.11.0.0
 - butcher-1.3.1.1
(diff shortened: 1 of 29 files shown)


stack-8.2.1.yaml
```diff
@@ -40,6 +40,8 @@ extra-deps:
 - sorted-list-0.2.1.0
 - syz-0.2.0.0
 - yi-rope-0.11
+- github: Bubba/haskell-lsp-test
```
alanz: Please use a submodule, like all the other dependencies.  It makes life a lot easier for those of us with constrained bandwidth at times.

test/Functional.hs
```diff
@@ -1,226 +1,56 @@
...
+import FunctionalDispatch
 
 main :: IO ()
 main = do
-  setupStackFiles
-  let logfile = "./test-functional.log"
-  exists <- doesFileExist logfile
-  when exists $ removeFile logfile
-  withFileLogging logfile $ cdAndDo "./test/testdata"  $ hspec spec
+  hspec spec
```
alanz: `dispatcherSpec` calls `setupStackFiles`, which makes sure the test projects are compatible with the version of hie we are running.

It is probably best to keep all that logic here, as it was before.

test/Functional.hs
```diff
@@ -1,226 +1,56 @@
...
+  dispatchSpec
 
 spec :: Spec
-spec = do
-  describe "functional spec" functionalSpec
-
--- ---------------------------------------------------------------------
+spec =
+  describe "deferred responses" $ do
+    it "do not affect hover requests" $ runSession "test/testdata" $ do
```
alanz: I think it would be good to use a version of `runSession` that takes the path to the hie exe to be used, which we can then get from the appropriate query to stack.  I do not think that the test should require a `stack install`, as this can potentially clobber a working hie when developing on a branch.

PR #617, opened 2018-06-03
Add a global message tracking number, for logging

alanz:
As each request arrives in the reactor, it is given a tracking number.

This is passed into the dispatcher, and logged as the message is
processed.

This makes it easier to understand/tie back the message processing,
when tracking down bugs.


Anrock:
@alanz is it possible to also log how many and what requests are pending? I'm having an issue now that looks like too much completion requests are put into queue and hie appears freezed - logging list of pending requests would help a lot to find a repro steps.

alanz:
@Anrock with this PR the request number is logged as it is received by the client, and then when processing starts in the various parts of the dispatcher.

That should be enough to track down what is happening to a message.

PR #629, opened 2018-06-13
Use UUIDs for commands

Bubba:
Currently when starting multiple instances of HIE for the same client, it an errors due to registering the same commands multiple times in the initialise response that we send. This happens when creating a multi-root/project workspace in VS Code.

This was discussed [here](https://github.com/Microsoft/vscode-languageserver-node/issues/333) and the suggested solution was to use UUIDs to register commands. 
This PR changes the server to register random UUIDs, and maps back and forth between them in code action and execute command requests. You can still send an execute request for "hare:demote" as well as the new random UUID for it, and have the same command executed.

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/LspStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/LspStdio.hs
@@ -134,10 +134,10 @@ run dispatcherProc cin _origDir captureFp = flip E.catches handlers $ do
   ioExcept (e :: E.IOException) = print e >> return 1
   someExcept (e :: E.SomeException) = print e >> return 1
   getCommandUUIDs = do
-    gen <- getStdGen
-    let commands = ["hare:demote", "applyrefact:applyOne"]
-        uuids :: [UUID]
-        uuids = randoms gen
+    uuid <- toText <$> randomIO
+    let commands :: [T.Text]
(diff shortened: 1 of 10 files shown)


src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -120,12 +127,18 @@ run dispatcherProc cin _origDir captureFp = flip E.catches handlers $ do
...
   handlers  = [E.Handler ioExcept, E.Handler someExcept]
   finalProc = L.removeAllHandlers
   ioExcept (e :: E.IOException) = print e >> return 1
   someExcept (e :: E.SomeException) = print e >> return 1
+  getCommandUUIDs = do
+    gen <- getStdGen
+    let commands = ["hare:demote", "applyrefact:applyOne"]
+        uuids :: [UUID]
+        uuids = randoms gen
+    return $ BM.fromList (zip commands uuids)
```
alanz: I think it might be good to include the original command too, so if the UUID is `abcdef`, then use a command string of `hare:demote:abcdef`.

It will make interpreting logs etc much easier.

alanz: And then we only need one UUID per session too, come to think of it.

PR #632, opened 2018-06-15
Use just plain process IDs instead of UUIDs to unique commands

Bubba:
Realised that there's no need for UUIDs if we're basing it off of process IDs


alanz:
Good idea, I was wondering what we could use as a unique base, this is a great one.

PR #634, opened 2018-06-16
Add another feature example

pythonissam:


wz1000:
I think this is a vscode-hie-server feature, not a haskell-ide-engine feature.

alanz:
I think we need to note that this feature is (currently) unique to the vscode plugin, as it is implemented completely there.

pythonissam:
@alanz @wz1000 So, is this accepted with noting that?

alanz:
Yes, please add that note.

PR #640, opened 2018-06-18
Add config options to readme

### LINKED ISSUE #635: How to control linters?
I haven't found a way to completely disable hlint. I find most of its suggestions either outright wrong (it suggests changing semantics) or just annoying. So I don't want to use a `.hlint.yaml` for every single project.

`hie --help` doesn't show anything useful either.

It should somehow be possible to control the behavior of the linters, which are shown [here](https://github.com/haskell/haskell-ide-engine/blob/1051cc69d3434f4ccf67e03846410ec463135db9/docs/Tools.md#linters--refactoring).

Bubba:
Closes #635


alanz:
Thanks.

PR #642, opened 2018-06-19
Fall back to a default config when there is none specified in settings.json

Bubba:
This was a problem in LanguageClient-neovim since it would send a didChangeConfiguration notification with an empty settings file. Now if the `languageServerHaskell` field isn't defined in the object it will fall back, but still throw an error if the parsing fails in another way

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/LspStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/LspStdio.hs
@@ -25,7 +25,7 @@ import           Control.Monad.IO.Class
 import           Control.Monad.STM
 import           Control.Monad.Reader
 import qualified Data.Aeson as J
-import           Data.Aeson ( (.=), (.:), (.:?), (.!=) )
+import           Data.Aeson ( (.=), (.:?), (.!=) )
 import qualified Data.Bimap as BM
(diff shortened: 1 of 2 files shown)


alanz:
There is already a default config, that is set up when the server is started.

The problem is really what to do if you get a malformed config request, and the options are to either silently ignore, or send back some kind of meaningful error message.

There is no need to change the config, as returning a `Left err` will cause `haskell-lsp` to continue using the config values from before.

So this patch adds complexity, in my opinion, without addressing the actual issue.

Bubba:
@alanz I agree, this already duplicates the default config options, but I think the issue is that it's treating an empty configuration file that doesn't contain a "languageServerHaskell" field as malformed when it seems like a valid option. All this PR really does is get rid off the error message that it generates. 

Would it make more sense to change `getConfig` to return a `Config` instead of an `Either` and move the error handling from `haskell-lsp` into hie, so we can ignore empty configs but still log other errors? And then we can remove `configVal`/`defVal`:
https://github.com/haskell/haskell-ide-engine/blob/374da0753b4c126eb71175284f0b6e551674abb6/src/Haskell/Ide/Engine/Transport/LspStdio.hs#L180-L184

Issue #414, opened 2017-12-18
Error out on GHC mismatch


domenkozar:
Given the recent GHC bump to 8.2.x, many people will get mismatches of GHC and currently hie doesn't report this. I suggest it errors out with a clear message.

Anrock:
iirc hie reports ghc mismatches pretty clearly on start.

PR #643, opened 2018-06-19
Check GHC version on launch

### LINKED ISSUE #459: Detect and report GHC version mismatch
If `hie` has been compiled with one GHC, and is being used on a project with a different one, report this to the user via an LSP error message.

At the moment it silently fails to load modules.

Bubba:
Logs to the hie log, logs to the lsp log and shows a message to the user if the HIE ghc version doesn't match up with the project ghc version, should hopefully close #459 and #414, and many more support requests


alanz:
This looks like it needs to be updated for the Semigroup / Monoid change that GHC 8.4 introduced.

PR #653, opened 2018-06-22
Update haskell-lsp-test to no longer require an extra-dep on process-1.6.3

Bubba:
Should hopefully fix window builds on 8.2.1/8.2.2


alanz:
LGTM, let's see what CI does

Anrock:
Let me check this on win.

Anrock:
LGTM.

PR #654, opened 2018-06-22
Fix stack command typo

Bubba:


alanz:
LGTM, merge when ci passes.

I guess we need tests for this stuff.

Bubba:
Superseeded by #656

alanz:
I think we should merge this now, while we consider whether it is good to use stack as a lib

PR #659, opened 2018-06-26
[WIP] ghc-mod code actions

Bubba:
This does some rudimentary string manipulation to read the warnings given out by ghc-mod, and provide contextual code actions for them. In the future we should probably add API in ghc-mod to directly give us the information to create these code actions. 

So far it adds these types of code actions:
- Renaming code actions for when something is misspelt (Perhaps you meant?)
- Importing missing modules by searching up symbols in Hoogle with Hsimport (based off of https://github.com/alanz/vscode-hie-server/pull/83)
- Add missing dependencies to the .cabal files (also based off of https://github.com/alanz/vscode-hie-server/pull/83 !)

It also adds support for the new code action types in LSP v3.8, so you can use quick fix and refactor keybindings and such inside VS Code. 

# Todo
- [ ] Add package.yaml support
- [ ] Add more code actions
- [ ] Add functional tests for all of this
- [ ] Possibly look at moving the `makeRequest` functions inside the reactor into functions in the `R` monad

# Also
This moves out the code action code (the logic that glues LSP and the various Haskell tools) into a separate module since it was getting pretty big, so some reactor types have also been moved out. 

It also introduces `Cabal` as a dependency to mechanically edit the .cabal file, and I'm not sure how much this affects build time. 
Its used to work out which library/executable should have the dependencies added to based on the module that the quick fix is being applied on.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
@@ -9,6 +9,7 @@ module Haskell.Ide.Engine.PluginUtils
     mapEithers
   , pluginGetFile
   , pluginGetFileResponse
+  , makeDiffResult
   , diffText
   , srcSpan2Range
   , srcSpan2Loc
@@ -32,6 +33,8 @@ import           Data.Algorithm.DiffOutput
 import qualified Data.HashMap.Strict                   as H
 import           Data.Monoid
 import qualified Data.Text                             as T
+import qualified Data.Text.IO                          as T
+import qualified GhcMod.Utils                          as GM
 import           FastString
 import           Haskell.Ide.Engine.MonadTypes
 import           Haskell.Ide.Engine.MonadFunctions
@@ -133,6 +136,14 @@ mapEithers f (x:xs) = case mapEithers f xs of
 mapEithers _ _ = Right []
 
 -- ---------------------------------------------------------------------
+
+-- | Generate a 'WorkspaceEdit' value from an original file and text to replace it.
+makeDiffResult :: FilePath -> T.Text -> (FilePath -> FilePath) -> IO WorkspaceEdit
+makeDiffResult orig new fileMap = do
+  origText <- T.readFile orig
+  let fp' = fileMap orig
+  fp <- GM.makeAbsolute' fp'
+  return $ diffText (filePathToUri fp,origText) new
 
 -- |Generate a 'WorkspaceEdit' value from a pair of source Text
 diffText :: (Uri,T.Text) -> T.Text -> WorkspaceEdit
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
@@ -10,6 +10,7 @@ module Haskell.Ide.Engine.PluginUtils
   , pluginGetFile
   , pluginGetFileResponse
   , makeDiffResult
+  , makeAdditiveDiffResult
   , diffText
   , srcSpan2Range
   , srcSpan2Loc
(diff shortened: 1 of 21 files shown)


Bubba:
Needs https://github.com/alanz/haskell-lsp/pull/94 https://github.com/alanz/haskell-lsp/pull/96 https://github.com/alanz/haskell-lsp/pull/97

hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
```diff
@@ -134,25 +144,70 @@ mapEithers _ _ = Right []
...
 -- ---------------------------------------------------------------------
 
--- |Generate a 'WorkspaceEdit' value from a pair of source Text
-diffText :: (Uri,T.Text) -> T.Text -> WorkspaceEdit
-diffText (f,fText) f2Text = WorkspaceEdit (Just h) Nothing
+data WithDeletions = IncludeDeletions | SkipDeletions
+  deriving Eq
+
+-- | Generate a 'WorkspaceEdit' value from an original file and text to replace it.
+makeDiffResult :: FilePath -> T.Text -> (FilePath -> FilePath) -> IdeM WorkspaceEdit
```
alanz: Should this functionality move to haskell-lsp, some time?

src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -47,17 +48,31 @@ dispatcherP :: forall m. TChan (PluginRequest m)
...
+      gchan <- liftIO $ do
+        ghcChan <- newTChanIO
+        ideChan <- newTChanIO
+
+        _ <- forkIO $ mainDispatcher inChan ghcChan ideChan
+        _ <- forkIO $ flip runReaderT stateVar $
+                        flip runReaderT caps $ 
+                          ideDispatcher env errorHandler callbackHandler ideChan
+
+        -- TODO: Causes STM deadlock?
```
alanz: What is going on here with this comment?

Bubba: I tried to get the ideDispatcher thread to throw its exceptions onto the main thread instead of dying silently, but it caused other exceptions due to me not doing the async properly. We should probably revisit this some time. I can remove the comment and create a todo for this?

src/Haskell/Ide/Engine/Plugin/Package.hs
```diff
@@ -0,0 +1,221 @@
...
+    ensureObject :: (Object -> Object) -> (Value -> Value)
+    ensureObject f v = case v of
+      Object o -> Object (f o)
+      _ -> error "Not an object!"
+
+    addDep (Array deps) = Array $ fromList (String pkgName:GHC.Exts.toList deps)
+    addDep _            = error "Not an array in addDep"
+
+
+-- | Takes a cabal file and a path to a module in the dependency you want to edit.
```
alanz: And it will lose comments and formatting?

hie-plugin-api/Haskell/Ide/Engine/PluginUtils.hs
Bubba: Probably, I imagine this would be useful for both clients and servers.

I'm also thinking we should probably also pass back the lsp funcs/lsp config back to haskell-lsp, so it can handle formatting the workspace edit according to the specification.

src/Haskell/Ide/Engine/Plugin/Package.hs
Bubba: With cabal files it *shouldn't*. The printed cabal module after editing doesn't contain any formatting, but it tries to only apply the additions in the diff.
For package.yaml I couldn't get makeDiffResult to work with it, so it loses formatting. Interestingly enough there's another GSOC project for format preserving yaml editing: https://github.com/wisn/format-preserving-yaml/

PR #661, opened 2018-06-27
Fixes #660 by adding windows specific extension

### LINKED ISSUE #660: hie-wrapper script not working on windows
The rewritten hie-wrapper script doesn't work on windows. This is because executable files in Windows have the extension .exe, and thus `findExecutable` fails to find it when looking for `hie-8.2.2`. On windows it should probably add the `.exe` part when searching.

Supernerd11:
This pull request fixes #660 by adding the required extension on windows.

### DIFF
--- a/app/HieWrapper.hs
+++ b/app/HieWrapper.hs
@@ -97,7 +97,7 @@ run opts = do
       args <- getArgs
       logm $ "args:" ++ show args
       logm "launching ....\n\n\n"
-      callProcess e args
+      callProcess e ("--lsp" : args)
       logm "done"
 
 -- ---------------------------------------------------------------------
--- a/app/HieWrapper.hs
+++ b/app/HieWrapper.hs
@@ -7,7 +7,7 @@ module Main where
 import           Data.Semigroup
 #endif
 import           Data.List
-import           Data.Maybe
+import           Data.Foldable
 import           Data.Version                          (showVersion)
 import           Haskell.Ide.Engine.MonadFunctions
 import           Haskell.Ide.Engine.Options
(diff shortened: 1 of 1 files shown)


Bubba:
Should we also add this check for`getProjectGhcVersion` in Base.hs?

Supernerd11:
That would probably be a good idea. Basically anywhere `findExecutable` is used.

Supernerd11:
@Bubba We don't actually. findExecutable seems to find an executable when used without extension. The bug was that it read the version number as it's extension, eg `hie-8.2.2` would assume that `.2` was the extension.

app/HieWrapper.hs
```diff
@@ -90,9 +90,9 @@ run opts = do
       args <- getArgs
       logm $ "args:" ++ show args
       logm "launching ....\n\n\n"
-      callProcess e args
+      callProcess e ("--lsp" : args)
```
Bubba: Not sure if we should do this by default, there does exist a non-lsp JSON mode that should be able to use the wrapper too

Supernerd11: This was mostly because of the way both the vscode and atom plugins handled custom wrappers. It could probably be removed.

Bubba: Looks like https://github.com/alanz/vscode-hie-server/pull/93 will add it for us, but its not merged yet

Supernerd11: I know. A better workaround for now would be to make a local batch/shell script to launch the wrapper with the lsp flag.

Supernerd11: Or even easier, change default executable to hie-wrapper.

Issue #490, opened 2018-03-08
Don't mark HLint parse errors as errors


LeanderK:
I have some `hlint` parse errors that happen because `haskell-source-exts` can't handle my code, but it compiles fine. I think only ghc-mod parse errors should handled as errors.

EDIT: I use an Extension that is not robustly handled by `haskell-source-exts` and nearly every file is marked red, which is a bit annoying.

alanz:
It is possible to turn off the hlint check, in vscode, by setting `languageServerHaskell.hlintOn` to `false`.

I am not sure if this configuration setting is enabled for other clients at the moment, make an issue on the relevant client if not.

I will also look at turning off hlint parse errors.

Can you give a short  example of a file that has this problem?

LeanderK:
Here https://github.com/ndmitchell/hlint/issues/450 is one and I opened another on haskell-src-exts https://github.com/haskell-suite/haskell-src-exts/issues/403.

Issue #443, opened 2018-01-17
Suppress hlint hints when there are GHC errors


alanz:

nponeccop:
Note the hints are already of lower priority after #375 / #378

Anrock:
At least in Neovim + LanguageClient-neovim lints and warnings aren't displayed if errors are present.
@alanz i think this can be closed.

PR #664, opened 2018-06-28
Ignore hlint errors instead of flushing all hlints 

Bubba:
Whenever ghc-mod returns an error it currently flushes all hlint diagnostics, even if hlint didn't necessarily have a parse error. This changes it to not flush the diagnostics and instead filters out the hlint errors in the first place. Should still keep #490

- [x] Add a config option

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
+++ b/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
@@ -247,10 +247,7 @@ filterIdeas (OneHint (Position l c) title) ideas =
   let
     title' = T.unpack title
     ideaPos = (srcSpanStartLine &&& srcSpanStartColumn) . ideaSpan
-    satisfiesAll preds i = foldl (\acc p -> acc && p i) False preds
-    predicates = [ \i -> ideaHint i == title'
-                 , \i -> ideaPos i == (l+1, c+1)]
-  in filter (satisfiesAll predicates) ideas
+  in filter (\i -> ideaHint i == title' && ideaPos i == (l+1, c+1)) ideas
 
 hlintOpts :: FilePath -> Maybe Position -> [String]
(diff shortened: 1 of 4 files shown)


Anrock:
@Bubba why closed?

Bubba:
@Anrock after discussing with @alanz, this is actually the intended behaviour, https://github.com/haskell/haskell-ide-engine/issues/443
We could also look at making this an configurable setting?

Anrock:
Hmm, this is confusing.

I see a reason in suppressing **errors** - there is no sense in having "parse error" from ghc and hlint at the same place, for example. 
However suppressing **hints** is a strange thing to do. 
Suppose, i'm writing a pattern match cases and despite parse error in current line (i didn't put a closing bracket yet) i still want to see a hint from hlint that pattern match is incomplete with unmatched pattern example. That's the real case and a reason i wanted this PR merged.

IMO, langserv should provide all info it has and client as a presentation layer should decide what to show or not.

alanz:
@Anrock I agree, but I also think it may be useful for hie to have a configuration setting for this, to make it easier for client authors.

Especially as it is a pretty simple thing to do.

Anrock:
@alanz i've checked lsp spec and it states that diagnostics are owned by server and client doesn't (even mustn't) do much with them, so this config option is actually pretty sane to be serverside. So i agree here.

Bubba:
@alanz @Anrock is the consensus to add a config option for this then?

Anrock:
@Bubba yep.

Bubba:
@Anrock Then the real debate now is what should the default be 😛

Anrock:
@Bubba while @alanz is away i vote for not flushing by default >_>

alanz:
I would propose we keep the existing behaviour, unless the setting changes it.

This gives the least surprise to existing users.

PR #668, opened 2018-07-04
Add version of icu-macos-fix with no brew install

Bubba:
Will be useful for making a brew formula where we can list icu4c as a
dependency there


alanz:
I am not on a mac, is this thing ok to land?

Anrock:
I'm away from mac, can check only in 12+ hours

Anrock:
Didn't have time to check today. But it's pretty mechanical change and everything looks fine.

PR #669, opened 2018-07-07
Add Windows batch equivalent of `make build-all`

mpilgrem:
The command `make build-all` does not work on 'out of the box' Windows. There are work-arounds, for example the Cygwin installer has the option of including a `make` command. This pull request suggests taking the approach of providing a Windows batch file that replicates the function of `make build-all`.  

`README.md` is updated, accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ we talk to clients.__
    ![Renaming](https://i.imgur.com/z03G2a5.gif)
 
  - Adding chosen import and the package dependencies (vscode only)
- 
+
    ![Adding import & deps](https://user-images.githubusercontent.com/1387653/40287051-b6f987fe-5c5f-11e8-980f-ed7bfa1b2aec.gif)
 
 ## Installation
@@ -169,6 +169,9 @@ stack --stack-yaml=stack-8.0.2.yaml install
 ```
 
 #### Installation on Windows
+
+The Windows batch file `make-build-all.bat` can substitute for `make build-all` on
+systems without the `make` command.
 
 In order to avoid problems with long paths you can do the following:
 
--- a/make-build-all.bat
+++ b/make-build-all.bat
@@ -1,0 +1,61 @@
+@REM Windows batch file script that replicates the function of 'make build-all'.
+
+@ECHO OFF
+
+SETLOCAL
+FOR /F "delims=" %%F IN ('stack path --local-bin') DO (
+SET STACKLOCALBINDIR=%%F
+)
+
+REM build-all
+CALL :build
+CALL :build-docs
+EXIT /B
+
+:build
+CALL :hie-8.2.1
+CALL :hie-8.2.2
+CALL :hie-8.4.2
+CALL :hie-8.4.3
+EXIT /B
+
+:hie-8.2.1
+CALL :submodules
+stack --stack-yaml=stack-8.2.1.yaml install happy
+stack --stack-yaml=stack-8.2.1.yaml install
+COPY "%STACKLOCALBINDIR%\hie.exe" "%STACKLOCALBINDIR%\hie-8.2.1.exe"
+COPY "%STACKLOCALBINDIR%\hie-8.2.1.exe" "%STACKLOCALBINDIR%\hie-8.2.exe"
+EXIT /B
+
+:hie-8.2.2
+CALL :submodules
+stack --stack-yaml=stack-8.2.2.yaml install happy
+stack --stack-yaml=stack-8.2.2.yaml install
+COPY "%STACKLOCALBINDIR%\hie.exe" "%STACKLOCALBINDIR%%\hie-8.2.2.exe"
+COPY "%STACKLOCALBINDIR%\hie-8.2.2.exe" "%STACKLOCALBINDIR%\hie-8.2.exe"
+EXIT /B
+
+:hie-8.4.2
+CALL :submodules
+stack --stack-yaml=stack-8.4.2.yaml install
+COPY "%STACKLOCALBINDIR%\hie.exe" "%STACKLOCALBINDIR%\hie-8.4.2.exe"
+COPY "%STACKLOCALBINDIR%\hie-8.4.2.exe" "%STACKLOCALBINDIR%\hie-8.4.exe"
+EXIT /B
+
+:hie-8.4.3
+CALL :submodules
+stack --stack-yaml=stack.yaml install
+COPY "%STACKLOCALBINDIR%\hie.exe" "%STACKLOCALBINDIR%\hie-8.4.3.exe"
+COPY "%STACKLOCALBINDIR%\hie-8.4.3.exe" "%STACKLOCALBINDIR%\hie-8.4.exe"
+EXIT /B
+
+:submodules
+git submodule update --init
+EXIT /B
+
+:build-docs
+stack --stack-yaml=stack-8.2.1.yaml exec hoogle generate
+stack --stack-yaml=stack-8.2.2.yaml exec hoogle generate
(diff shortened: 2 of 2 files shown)


Anrock:
Wait wat, you can't build hie without stack and stack on windows brings msys with make.
So it's impossible to be unable to build because of make absent. Right?

mpilgrem:
@Anrock, if `stack` comes with and provides `make` - and I am not sure that it does - I don't know to access it.

For example, the command `stack exec where -- gcc` reveals that `stack exec` has added to the environment a path to `gcc.exe` at (for example) `%LOCALAPPDATA%\Programs\stack\x86_64-windows\ghc-8.4.3\mingw\bin\gcc.exe`. However, the corresponding command `stack exec where -- make` draws a blank (`INFO: Could not find files for the given pattern(s).`).

Anrock:
Yup, just reinstalled stack and indeed make is not bundled by default. 

However, is it really necessary to have .bat-copy of Makefile if make is trivially installed with single `stack exec pacman -- -S make`? I don't think any of regular maintainers are able to maintain it.

mpilgrem:
The answers to the following questions are for others, but I think relevant questions are 'What sort of potential users does the Haskell IDE Engine want to help?' and 'How best to support them?'.

I have some experience of Windows and coding (as a hobbyist) and relatively little experience of unix-like operating systems. Having researched what `make build-all` did and then what the native Windows equivalent would be, my hypothesis is that an experienced coder maintaining the `.bat` file would be less 'costly' than a relatively inexperienced Windows user having to learn unix-like tools exposed through `stack exec` to set things up to use Haskell with the Haskell IDE Engine. The `.bat` file and `Makefile` are actually very similar in their structure, once you see both side by side, and so 'translation' between the two is not very challenging.

I could propose a further pull request to: (1) add comments to the `.bat` file, to make it easier for unix -only coders to understand how its contents relate to the `Makefile`; and/or (2) add further content to `README.md` to explain how, in the alternative, Windows users can use `stack exec` to obtain the `make` tool and then run `make build-all`.

Anrock:
> my hypothesis is that an experienced coder maintaining the .bat file would be less 'costly' than a relatively inexperienced Windows user having to learn unix-like tools exposed through stack exec to set things up to use Haskell with the Haskell IDE Engine.

1) AFAIK, HIE doesn't have experienced contributor that can maintain Windows-specific pieces. @alanz and @Bubba don't have Windows machines, therefore they can't even run and test .bat files. I do have Windows machine, but i don't want to learn Windows batch language. 
So either someone would be that experienced coder maintaining it or it will quickly fall out of sync with Makefile and become effectively deadcode. That's what i'm afraid.
2) IMO, to copypaste and run `make build-all` or even `stack exec pacman -- -S make && make build-all` you don't have to learn unix tools. Anyway, almost any dev activity on Windows ends up with installing msys\mingw\cygwin and using unix tools, so learning now is just as good as any other time.

> The .bat file and Makefile are actually very similar in their structure, once you see both side by side, and so 'translation' between the two is not very challenging.

Yeah, that's exactly my concern here - why create & maintain essentially second copy of Makefile if first one already there, works on all platforms with little effort, and can be maintained and tested by anyone on any platform?

> I could propose a further pull request to: (1) add comments to the .bat file, to make it easier for unix -only coders to understand how its contents relate to the Makefile;
This wouldn't help much - unix-only devs would still have to edit it blindly without means to test. If they even remember to update it along with Makefile

> and/or (2) add further content to README.md to explain how, in the alternative, Windows users can use stack exec to obtain the make tool and then run make build-all.

Please do this.

wz1000:
Can't the commands in the make file be integrated into `hie-wrapper` or something? Then we have a cross platform solution(since everyone who needs to run `hie` needs to have some way of running Haskell programs).

We can have commands like

    hie-wrapper --install build-all

nponeccop:
> unix-only devs would still have to edit it blindly without any means to test

I'll volunteer on getting AppVeyor Windows CI running when I have time

Anrock:
@wz1000 not sure how this would work. We can use Shake as a haskell make replacement, sure, however make rules are the least significant thing in building hie. 
At the end, to build hie, hie-wrapper would have to incorporate also git to download sources and stack\ghc to compile them.

Anrock:
@nponeccop having precompiled binaries is the best solution here. 
User won't have issues compiling HIE if he won't compile it.

nponeccop:
Well, then CI will create the binaries for us. We may even want to ship some version of make with us, Perl distributions shipped various nmake/amake/gmake variants which both work well on windows and don't have much dep.

wz1000:
@Anrock our make file doesn't use any Make features(i.e. it doesn't do any dependency resolution for building), it is just a way to get a cross platform shell script. Shake is overkill for our task. We use cabal/stack as our build tool, not Make.

All we need to do it use the `process` library to run the necessary commands.

wz1000:
For example, if we just put the following code in hie-wrapper

    if args = ["--install", "hie-8.2.1"]
    then do
      callProcess "stack" ["--stack-yaml=stack-8.2.1.yaml", "install", "happy"]
      ...

Anrock:
@wz1000 oh, i see. That's cool, but kinda strange.
1) You would still have to get sources somehow
2) You would still have to get hie wrapper somehow
3) You lose agility of plain-text script files

Also, i've just noticed that stack isn't shipped with git also. And you can't replace git with bat file, so this a dead horse. Except precompiled binaries.

wz1000:
1. You have to get the sources to use a .bat file or a Make file too
2. Hie wrapper can be a simple script with very few dependencies, so you just need to do a `stack/cabal install hie-wrapper` to get it. If you need HIE, you will have a way to compile and use Haskell programs.
3. Makefiles/batch files are as plain text as .hs files

Anrock:
@wz1000 
1) I mean having Makefile or having its equivalent in hie-wrapper doesn't make a difference here. But we already have Makefile
2) Yeah, ok. But anyway you would have to get sources somehow, so you would also have to install git and if you need to install git why not also install make? And maintaining makefile which specifically designed to build and install software is anyway easier than general purpose haskell, right?
3) But you don't have to compile them to run. Though you need to have make or cmd to run them anyway... Ok, i see where this is going.

I'll try to make a (opionated) summary here. 

- We had only a makefile, which was fine in terms of maintaining, but had `make` dependency, which isn't installed by default on Windows.
- Now we also have a .bat file, which makes `make` dependency optional, but requires maintaining Windows-only piece of code and nobody have ability\will to do it.

Now proposed options are:
- Do nothing and don't forget to sync Makefile with .bat and pray it still works.
- Add one line to README on how to install `make` (and git, btw) on Windows, remove .bat file. `make` dependency still will be there, but Windows-specific code and duplication are eliminated.
- As Makefile is overkill and actually almost flat sequence of commands needed - put Makefile equivalent right into hie-wrapper, put hie-wrapper to hackage/stackage and tell people to run `stack install hie-wrapper` and then `git clone` hie repo and run `stack exec hie-wrapper -- --build-all`. Removes both `make` and Win issues but adds some complexity to hie-wrapper and hie-wrapper has to be moved separate project otherwise it won't make to stackage/hackage, i assume.
- When @nponeccop or someone manages to setup CI so we can have prebuilt binaries it would be possible to make `hie-wrapper` standalone and make it just download prebuilt binaries depending on ghc version of current project.

And it looks like those aren't options but more like incremental steps.

mpilgrem:
I've made a pull request (#672) proposing further content for `README.md`.

alanz:
We could perhaps look at what @ndmitchell is doing for hlint, which has binaries available for installation on various platforms.

I think these are generated as part of the CI process.

PR #676, opened 2018-07-10
Add lower bound on yaml version

larskuhtz:
Module `Haskell.Ide.Engine.Plugin.Build` uses `decodeThrow` which was added to the yaml package in version 0.8.31.

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -77,7 +77,7 @@ library
                      , text
                      , transformers
                      , vector
-                     , yaml
+                     , yaml >= 0.8.31
                      , yi-rope
   if os(windows)
     build-depends:     Win32


Bubba:
This has been fixed in the import-code-actions branch albeit without the dependency bound, just changing decode to decodeThrows. It might not get merged for a while though

larskuhtz:
The fix in the import-code-actions branch seems to address only stack builds. So far, I didn't have any luck in getting stack builds of haskell-die-engine to work with plain cabal projects. Plain cabal builds have worked fine for me. It would therefore be nice if haskell-ide-engine would continue to be usable/buildable with just cabal.

Bubba:
I agree, I will try to get a version of haskell-lsp-test uploaded to hackage at some point so we can start to test with cabal too. And then maybe even get cabal on CircleCI

PR #696, opened 2018-07-21
Bump to 0.2.2.0

Bubba:
Add changelog details


alanz:
Should the hie-plugin-api version track this one? Probably not.

Bubba:
Just did a diff, looks like the API did change. Should it match with hie's or just follow PVP?

alanz:
If it changes, may as well match hie

PR #698, opened 2018-07-22
Add a cabal CircleCI job

Bubba:


.circleci/config.yml
```diff
@@ -57,7 +57,7 @@ defaults: &defaults
 
     - run:
         name: Test
-        command: stack -j 1 --stack-yaml=${STACK_FILE} test --dump-logs
+        command: stack -j 2 --stack-yaml=${STACK_FILE} test --dump-logs
```
alanz: I explicitly set this to use one core only, as the tests make use of the same test data, and running more than one ghc-mod/cabal-helper instance against the same project at the same time can cause problems. And did in the past.

Issue #701, opened 2018-07-23
Refactor Code Actions


Bubba:
## Code actions and plugins
At the moment, we have one big module `CodeActions.hs` that does all the logic for taking in diagnostics and spitting out code actions. Ideally, all of this logic should be done inside the plugins so that `CodeActions.hs` can gather them up without having to deal the specifics of each plugin.
As a rough idea, the `PluginDescriptor` could take on another field to help produce code actions:
```haskell
data PluginDescriptor =
  PluginDescriptor { pluginName :: T.Text
                   , pluginDesc :: T.Text
                   , pluginCommands :: [PluginCommand]
                   , pluginCodeActions :: [Diagnostic] -> IdeM [CodeAction]
                   }
```
## Command identifiers
Currently command identifiers are passed around as strings, and have a lot of dodgy logic for converting to and from their unique, PID prepended counterparts. This should be refactored so that they are strongly typed, and the PID-ifying is done behind the scenes. 

## Todo

- [x] Add new field to `PluginDescriptor`
- [x] Migrate existing plugins over
- [x] Refactor command identifiers

alanz:
@Bubba I just noticed that the docs for a `CodeAction` say

> If a code action provides an edit and a command, first the edit is executed and then the command.

The current `wrapCodeAction` behaviour ignores the command if there is an edit, when there is no `literalSupport`.

I am not sure what the best way is to group this so that if the user in the client chooses the action, both things happen, in the required order.  Perhaps we should receive a specific wrapper command, that returns the workspaceEdit and then executes the command.

Bubba:
@alanz yeah we should probably just bundle up both the workspace edit and command, into the fallback command, good eye

PR #715, opened 2018-07-25
Improve backwards compatible code actions, simplify command mapping

Bubba:
Does part of #701.
Both workspace edits and commands are now included in the `fallbackCodeAction` shim command.
Also gets rid of the command map which is now no longer needed, and automatically sends all plugin commands in the initialise request

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/LspStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/LspStdio.hs
@@ -26,7 +26,6 @@ import           Control.Monad.STM
 import           Control.Monad.Reader
 import qualified Data.Aeson as J
 import           Data.Aeson ( (.=) )
-import qualified Data.Bimap as BM
 import qualified Data.ByteString.Lazy as BL
 import           Data.Char (isUpper, isAlphaNum)
 import           Data.Default
@@ -84,10 +83,11 @@ lspStdioTransport
   :: (DispatcherEnv -> ErrorHandler -> CallbackHandler R -> ClientCapabilities -> IO ())
   -> TChan (PluginRequest R)
   -> FilePath
+  -> IdePlugins
   -> Maybe FilePath
   -> IO ()
-lspStdioTransport hieDispatcherProc cin origDir captureFp = do
-  run hieDispatcherProc cin origDir captureFp >>= \case
+lspStdioTransport hieDispatcherProc cin origDir plugins captureFp = do
+  run hieDispatcherProc cin origDir plugins captureFp >>= \case
     0 -> exitSuccess
     c -> exitWith . ExitFailure $ c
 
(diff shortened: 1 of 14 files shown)


alanz:
LGTM

Do we have tests for the same plugin codeActions being used in a client with both the old and the new style of code actions?

alanz:
@gbaz It looks like appveyor is auto building here, so it works. Thanks.

Bubba:
Yes,
https://github.com/Bubba/haskell-ide-engine/blob/ad1fae71ad0c41eecdc0fb47529be623f8e30416/test/functional/CodeActionsSpec.hs#L57-L64
has the new code action capability, and
https://github.com/Bubba/haskell-ide-engine/blob/ad1fae71ad0c41eecdc0fb47529be623f8e30416/test/functional/CodeActionsSpec.hs#L21-L33
doesn't declare it, which is why it receives a command and not a code action

alanz:
@Bubba I would expect there to be 2 variations of this test, with differences in [this line](https://github.com/Bubba/haskell-ide-engine/blob/ad1fae71ad0c41eecdc0fb47529be623f8e30416/test/functional/CodeActionsSpec.hs#L21)

```haskell
 it "provides hlint suggestions" $ runSession hieCommand "test/testdata" $ do
```

to parameterise the client to honour the old vs the new code action behaviour.

Is that currently possible? Or do I misunderstand?

Bubba:
Can you hold off on merging this? I'm getting some issues with this in vs code

alanz:
That reminds me.  The Plugin interface should return possible commands the plugin can register, for initialisation to vscode, else it won't use the command.  Rather than that hardcoded map in LspStdIO.hs

And it needs to manage the commandMap stuff.

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -523,20 +532,25 @@ reactor inp commandMap = do
...
-              -- if this is a UUID then use the mapping for it
-              command = fromMaybe command' (BM.lookupR command' commandMap)
+              command = stripCmdPrefix (params ^. J.command)
+
+              -- | strips the PID prefix from the command id if it has one
+              stripCmdPrefix :: T.Text -> T.Text
+              stripCmdPrefix x
+                | T.count ":" x >= 2 = T.tail $ T.dropWhile (/= ':') x
+                | otherwise = x
+
```
alanz: Maybe just split the command on ":" (using `Data.Text.splitOn`), and if you get three chunks discard the first. Then you will already have broken the command into its plugin and command within the plugin.

alanz: Or use something specific for the start of the prefix, and just drop if it is there.

So do something like "eg2:addPragma" becomes "(12345)eg2:addPragma", and if the first char is not an opening paren you keep it as is, else strip off until the closing paren.

Because it is only the vscode client that cares about the prefix, we can toss it when the message comes in.

Bubba:
Tidied that up a bit. It's still quite a big chunk in the reactor though, should we split that out into another module?

PR #719, opened 2018-07-29
update Readme.md, neovim: add --lsp argument

srghma:


Bubba:
The `--lsp` argument is no longer needed since its now the default: https://github.com/haskell/haskell-ide-engine/pull/706

srghma:
Ups, sorry, seems like I'm using old hie

PR #723, opened 2018-07-30
diagnostic providers

alanz:
Adds the ability for a given plugin to generate diagnostics.

These can be triggered from any set of `DidOpen`, `DidChange` or `DidSave` notifications.

Still need to add a config option to enable the example plugin, else normal users will get spammed with junk.


Bubba:
You could probably expose an option to enable the example plugins in the LSP config options, which I think lives in the `Config` type in `LSPStdio.hs`

alanz:
@Bubba that is what I was thinking too. Default off, with the JSON instance written so it does not care if the field is missing from anything that comes in.

PR #731, opened 2018-07-31
Fix JsonStdio transport.

arrowd:
The `getNextReq` function was calling `exitProcess` causing whole application to exit too early.


alanz:
LGTM, will merge when CI passes

PR #732, opened 2018-07-31
Liquid plugin part one, diagnostics and types on hover

alanz:
Work in progress on liquid haskell plugin.

Currently extracts errors and presents them as diagnostics.

Next step is the type info on hover.


haskell-ide-engine.cabal
```diff
@@ -17,6 +17,10 @@ flag pedantic
   Description: Enable -Werror
   Default:     False
 
+flag dev
+  Description: Enable development, where unit tests include src rather than lib
+  Default:     False
```
Bubba: Is this a workaround for stack rebuilding the extra-deps?

alanz: It's to allow ghci to reload all the files when you are working on a test, and change something in the core lib.

But it does not seem to work for `stack repl` or for `cabal repl`.

alanz: And it did not work because I was testing against pedantic, not dev. Oops.

PR #735, opened 2018-08-01
Command ID refactor redux

Bubba:
Instead of creating new types this time, add a function `mkLspCommand` that converts HIE commands to LSP Commands. It grabs the process ID every time though so we would probably want to store this at the start. The only thing is that this function needs to be available in both the reactor and IdeM monads.

Will hopefully be enough to close #735


alanz:
This is a step in the right direction.

I think the final piece will be to have registration in `MainHie.hs` look something like 

```haskell
      ,("package"    , packageDescriptor "package")
```
And the parameter to the `packageDescriptor` is then used instead of the hard-coded one in

```haskell
 cmd <- mkLspCommand "package" "add" title (Just cmdParams)
```

alanz:
@Bubba, I think this branch has to be reworked against current master.  Once that is done it can land.

PR #760, opened 2018-08-15
HaRe code actions

Bubba:
Follows #759 
Allows other clients than vs code to access HaRe's refactoring!


alanz:
At the moment HaRe does some precondition checks before actually doing a refactoring.

Given that we potentially have the type checked module already, these checks can be cheap.

So we can call the check, and if it passes add the action.

It will require some changes to HaRe though.

alanz:
Merging, appveyor failing for artifact store size limit exceeded.

PR #765, opened 2018-08-17
Disable appveyor artifacts for now

Bubba:
Since we're running out of space

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,6 +18,8 @@ cache:
 build_script:
 - cmd: stack --stack-yaml=stack-%GHCVER%.yaml install
 test: off
-artifacts:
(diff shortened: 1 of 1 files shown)


nponeccop:
Note that we could 7z them.

Bubba:
Ideally we could just delete old artefacts, but people are asking [here](https://help.appveyor.com/discussions/problems/10253-maximum-allowed-artifact-storage-size-of-1000-mb-will-be-exceeded) and getting their storage sized bump. I think we would quickly run through the extra 20GB though

nponeccop:
I think the solution is:

- 7z the artifacts (GHC binaries tend to compress 10x)
- produce artifacts nightly (appveyor if I remember correctly allows nightly builds)

PR #771, opened 2018-08-19
Improved the suggestions for typed holes actions

### LINKED ISSUE #767: Consider showing local bindings for type hole substitutions in 8.4.3
Since GHC does not show local bindings as valid substitutions in 8.4.3, it would be cool to append the `Relevant bindings include` section in the bottom of the list.

 One heuristic could be that if the only valid substitution is `undefined` or the `error*` functions, then appending the local bindings for help may be a good idea.

Additionally, I think it would be helpful to show the type of the binding as part of the substitution. For instance, it could read as follows:

```
Substitute hole (ShellCheck.AST.Token) with allFlags (ShellCheck.AST.Token -> [String])
Substitute hole (ShellCheck.AST.Token) with hasYesOption (ShellCheck.AST.Token -> Bool)
Substitute hole (ShellCheck.AST.Token) with undefined
```

lorenzo:
This implements what is described in #767

I took the liberty of renaming one test file, so I could run the tests
from inside ghci. Without the rename, it would complain about duplicate
module names

closes #767

### DIFF
--- a/test/functional/FunctionalCodeActionsSpec.hs
+++ b/test/functional/FunctionalCodeActionsSpec.hs
@@ -17,7 +17,7 @@ import qualified Language.Haskell.LSP.Types as LSP
 import qualified Language.Haskell.LSP.Types.Capabilities as C
 import Test.Hspec
 import TestUtils
-import Data.Semigroup
+import Data.Monoid ((<>))
 
 spec :: Spec
 spec = describe "code actions" $ do
--- a/test/functional/FunctionalCodeActionsSpec.hs
+++ b/test/functional/FunctionalCodeActionsSpec.hs
@@ -17,6 +17,7 @@ import qualified Language.Haskell.LSP.Types as LSP
 import qualified Language.Haskell.LSP.Types.Capabilities as C
 import Test.Hspec
 import TestUtils
+import Data.Semigroup
 
 spec :: Spec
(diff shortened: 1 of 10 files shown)


test/functional/FunctionalCodeActionsSpec.hs
```diff
@@ -233,6 +233,31 @@ spec = describe "code actions" $ do
             \foo :: [Int] -> Int\n\
             \foo x = maxBound"
 
+      it "shows more suggestions" $ when ghc84 $
```
alanz: Can we add parsing for any of the other GHC versions too?

lorenzo: Sure, I can look into that

lorenzo:
@alanz done

test/functional/FunctionalCodeActionsSpec.hs
```diff
@@ -17,6 +17,7 @@ import qualified Language.Haskell.LSP.Types as LSP
 import qualified Language.Haskell.LSP.Types.Capabilities as C
 import Test.Hspec
 import TestUtils
+import Data.Semigroup
```
Bubba: You can use import `Data.Monoid ((<>))` which works on both 8.4 and 8.2

lorenzo:
@Bubba Fixed

PR #777, opened 2018-08-20
Remove IdeResponse, integrate it into IdeM as FreeT IdeDefer

Bubba:
This removes the whole `IdeResponse`/`pluginGetFileResponse` headache. Use `IdeResult` now instead.
It also removes the versions of `withCachedModule` that didn't take a default. Now they must provide a default fallback if there is no module (which most of them did anyway), removing the `NoModuleAvailable` error code.
The dispatcher now also handles all the logic for deferred responses now, whereas previously half of it was in`ModuleCache.hs` as well.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
@@ -9,6 +10,8 @@
 {-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 -- | IdeGhcM and associated types
 module Haskell.Ide.Engine.PluginsIdeMonads
@@ -31,18 +34,15 @@ module Haskell.Ide.Engine.PluginsIdeMonads
   , SymbolProvider
   , IdePlugins(..)
   -- * The IDE monad
-  , IdeGhcM
   , IdeState(..)
+  , IdeGhcM
   , IdeM
+  , iterT
   , LiftsToGhc(..)
-  -- * IdeResult and IdeResponse
+  -- * IdeResult
   , IdeResult(..)
   , IdeResultT(..)
-  , pattern IdeResponseOk
-  , pattern IdeResponseFail
-  , IdeResponse
-  , IdeResponse'(..)
-  , IdeResponseT(..)
+  , IdeDefer(..)
   , IdeError(..)
   , IdeErrorCode(..)
   -- * LSP types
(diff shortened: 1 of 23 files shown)


hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -198,22 +199,46 @@ instance ToJSON IdePlugins where
...
+type IdeGhcM = GM.GhcModT IdeBase
 
 instance MonadMTState IdeState IdeGhcM where
   readMTS = lift $ lift $ lift readMTS
-  modifyMTS f = lift $ lift $ lift $ modifyMTS f
+  modifyMTS = lift . lift . lift . modifyMTS
+
+-- | A computation that is deferred until the module is cached.
+-- Note that the module may not typecheck, in which case 'UriCacheFailed' is passed
+data IdeDefer a = IdeDefer FilePath (UriCache -> a) deriving Functor
```
alanz: This is probably not the right time to do it, but at some point we should let the module cache be more lenient, in the sense that if we get anything out, we call the function.

Here anything is `ParsedSource`, `RenamedSource`, `TypeCheckedSource`.

src/Haskell/Ide/Engine/Dispatcher.hs
```diff
@@ -81,46 +80,33 @@ mainDispatcher inChan ghcChan ideChan = forever $ do
     Left r ->
       atomically $ writeTChan ideChan r
 
-ideDispatcher :: forall void m. DispatcherEnv -> ErrorHandler -> CallbackHandler m -> TChan (IdeRequest m) -> IdeM void
-ideDispatcher env errorHandler callbackHandler pin = forever $ do
+ideDispatcher :: forall void m. TVar IdeState -> J.ClientCapabilities -> DispatcherEnv -> ErrorHandler -> CallbackHandler m -> TChan (IdeRequest m) -> IO void
+ideDispatcher stateVar caps env errorHandler callbackHandler pin = flip runReaderT stateVar $ flip runReaderT caps $ forever $ do
```
alanz: Long lines again, difficult to review

alanz:
Nothing obvious jumps out at me.

alanz:
@Gurkenglas, @Bubba 

Are we ok with this?  Should I merge?

@wz1000 any comment?

wz1000:
@alanz Seems reasonable to me.

alanz:
@Gurkenglas 229?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -198,22 +198,42 @@ instance ToJSON IdePlugins where
 
 -- ---------------------------------------------------------------------
 
-type IdeGhcM = GM.GhcModT IdeM
+type IdeGhcM = GM.GhcModT IdeBase
```
alanz: This definition should be grouped with the other three, just below.

Bubba:
@Gurkenglas @alanz Instead of throwing an error/having an invariant, how would you feel about making `IdeM` `IdeDefer` and making `IdeBase` `IdeM`? This way we can be explicit in the type signatures of when and where you can defer. On the other hand, its added complexity

alanz:
@Bubba I do not have strong feelings either way, except I prefer simplicity in general.

And we need to merge this thing some time.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
Bubba: I wrote this review 5 days ago, but didn't realise you had to submit it for you to see it...

Issue #667, opened 2018-07-04
`hGetContents: invalid argument (invalid byte sequence)` on Japanese Windows


igrep:
# My environment

OS: Windows 10 Pro ver. 1709 build 16299.431
Editor: NVIM v0.2.2
LSP Client: https://github.com/autozimu/LanguageClient-neovim ver. 0.1.93

HIE (extracted from the log file):

```
hie-wrapper(hie-wrapper) Version 0.2.0.0, Git revision 2a8bad55422cb436b18660dd9238c315b1d9bf82 (1545 commits) x86_64 ghc-8.2.2
HIE(hie-8.2.2.exe) Version 0.2.0.0, Git revision 2a8bad55422cb436b18660dd9238c315b1d9bf82 (1545 commits) x86_64 ghc-8.2.2
```

# Problem

An error is reported when opening a file in a project where some files contains Japanese characters.
From the debug log:

```
2018-07-04 10:30:51.3777965 [ThreadId 5] - <--2--{"jsonrpc":"2.0","params":{"type":1,"message":"Got error while processing diagnostics: C:\\Users\\yuji-yamamoto\\AppData\\Local\\Temp\\ghc-mod32077\\Experiment32076-0.hs: hGetContents: invalid argument (invalid byte sequence)"},"method":"window/showMessage"}
```

# Suggestion and question

I know this is a very popular error on Japanese Windows related to character encoding.
I've once fixed very similar one in haddock: https://github.com/haskell/haddock/pull/566.
So the solution is perhaps to add `hSetEncoding handle utf8` before reading a source file.

Then, where should I fix?
According to the next line of the error in the debug log, hlint is successfully executed:

```
[...]

PR #783, opened 2018-08-25
Use Brittany with hsimport when needed

### LINKED ISSUE #772: Run Brittany after adding an import via hsimport
Just on the line that was added to make it aligned with the rest of the imports

bubba:
This does a spartan check to see if the imports are formatted the way Brittany formats it, and if so then applies a coat of Brittany to the import code action.
Closes #772 

# Todo
- [x] Add LSP config option to disable this


Gurkenglas:
The obvious way to check if it's formatted with britanny is to format it with brittany and see if that changes anything.

src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -67,8 +74,34 @@ importModule uri modName =
...
+              newDocChanges <- forM mDocChanges $ mapM $ \(J.TextDocumentEdit vDocId tes) -> do
+                ftes <- forM tes (formatTextEdit confFile)
+                return (J.TextDocumentEdit vDocId ftes)
+    
+              return $ IdeResultOk (J.WorkspaceEdit newChanges newDocChanges)
+            else
+              return $ IdeResultOk (J.WorkspaceEdit mChanges mDocChanges)
+
+  where hasFormattedImports fp = do
+          ls <- T.lines <$> liftIO (T.readFile fp)
```
lorenzo: Unrelated to this PR, can we consider not using the read function from the Text package? I’ve had problem in the past with locale settings messing up the results. 

An alternative is to used ByteString and do a lenient decode to utf8

Bubba: Could this same issue be causing #667?

lorenzo: Yes, that's exactly how the problem manifests. Should we create a ticket to get this addressed?

Bubba: @Lorenzo We can probably just reuse that ticket, and add some comments explaining that it could be related to `Data.Text.IO.readFile`. I just did a quick search and hie does use it in quite a few locations

Bubba:
Updated to simply always apply Brittany formatting

src/Haskell/Ide/Engine/Plugin/HsImport.hs
kRITZCREEK: I ran into similar issues when working on the PureScript IDE integration, if it's any help we're using these functions to read source files currently: https://github.com/purescript/purescript/blob/b07042fb7b4f6a68c0fdc68b7e51076b7106cfba/src/System/IO/UTF8.hs

alanz:
I think this PR needs an update.

alanz:
Then we can probably merge it.

bubba:
Not sure what I did differently but managed to add `LspFuncs` into `IdeM`, which now gives all plugins access to the Config so we can add an LSP option to disable this. This also turned out to have some great benefits and I was able to refactor `Scheduler` so that `ideDispatcher` and `ghcDispatcher` are symmetrical.
I plan on revising #861 so we can move VFS access straight into IdeM too!

bubba:
Ready for re-review

hie-plugin-api/Haskell/Ide/Engine/Config.hs
```diff
@@ -46,19 +58,20 @@ instance FromJSON Config where
...
                  , "completionSnippetsOn" .= c
+                 , "formatOnImportOn"     .= f
                  ]
 
 -- ---------------------------------------------------------------------
 
 -- | For the diagnostic providers in the config, return a map of
 -- current enabled state, indexed by the plugin id.
-getDiagnosticProvidersConfig :: Config -> Map.Map PluginId Bool
+getDiagnosticProvidersConfig :: Config -> Map.Map T.Text Bool
```
lorenzo: Why is this different?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -220,12 +227,57 @@ instance ToJSON IdePlugins where
...
+    Nothing -> return Nothing
+
+getRootPath :: IdeM (Maybe FilePath)
+getRootPath = do
+  IdeEnv mlf <- ask
+  case mlf of
+    Just lf -> return (Core.rootPath lf)
+    Nothing -> return Nothing
+
+data IdeEnv = IdeEnv (Maybe (Core.LspFuncs Config))
```
lorenzo: Can you make this a `newtype` instead?

src/Haskell/Ide/Engine/Scheduler.hs
```diff
@@ -271,10 +259,10 @@ ideDispatcher
      -- ^ Callback to run for handling the request.
   -> Channel.OutChan (IdeRequest m)
      -- ^ Reading end of the channel where the requests are sent to this process.
-  -> IO void
-ideDispatcher stateVar caps env errorHandler callbackHandler pin =
+  -> IdeM void
+ideDispatcher env errorHandler callbackHandler pin =
   -- TODO: AZ run a single ReaderT, with a composite R.
```
lorenzo: This TODO should be moved to the place where the code was moved to

hie-plugin-api/Haskell/Ide/Engine/Config.hs
bubba: `PluginId` is defined inside `PluginsIdeMonads`, which now imports `Config` so its to avoid a circular dependency. This is probably a sign that we should split the module up a bit

lorenzo: Ah, makes sense

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
bubba: I meant to make this more than a wrapper for LspFuncs, for things that shouldn't change during the execution of IdeM/IdeGhcM/IdeDeferM. Updated it to include the PID and plugins

src/Haskell/Ide/Engine/Scheduler.hs
bubba: Should be fixed

bubba:
Looks like chocolately is down, AppVeyor build should be fine otherwise

hie-plugin-api/Haskell/Ide/Engine/Config.hs
alanz: Should we not keep it as a PluginId, just define it in a different place? I am not all that happy with it sudennly being `Text`, we are losing coherence.

bubba: Can this function be moved into `LspStdio.hs`? to avoid the circular dependency? It feels like it belongs in `Config.hs` though. I couldn't really find a logical grouping of Plugin related types to move out `PluginsIdeMonads` since they all reference IdeGhcM etc.
Another possibility is that `PluginDescriptor` gets moved into `PluginsIdeMonads` where the actual PluginDescriptor type lives

bubba:
@alanz @lorenzo is this ready to merge?

lorenzo:
I’m ok with merging this

alanz:
LGTM, merge when CI passes

alanz:
BTW, I have seen intermittent failures with that test on various branches / ghc versions recently. I think we should try to make it more robust.

PR #786, opened 2018-08-26
Import autocomplete

### LINKED ISSUE #746: Imports autocomplete
When specifying imports `import Data.<cursor>` no modules are suggested. Forcing completions using `ctrl+space` gives only local suggestions.

Is this not supported?

lorenzo:
closes #746

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE TypeFamilies        #-}
-{-# LANGUAGE FlexibleContexts    #-}
 module Haskell.Ide.Engine.Plugin.HieExtras
   ( getDynFlags
   , getCompletions
@@ -56,7 +55,10 @@ import           Var
 getDynFlags :: Uri -> IdeM (IdeResponse DynFlags)
 getDynFlags uri =
   pluginGetFileResponse "getDynFlags: " uri $ \fp ->
-    withCachedModule fp (return . IdeResponseOk . ms_hspp_opts . pm_mod_summary . tm_parsed_module . tcMod)
+    withCachedModule fp (return . IdeResponseOk . getDynFlagsPure)
+
+getDynFlagsPure :: CachedModule -> DynFlags
+getDynFlagsPure = ms_hspp_opts . pm_mod_summary . tm_parsed_module . tcMod
 
 -- ---------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -141,16 +159,15 @@ instance ModuleCache CachedCompletions where
...
-#else
-        asNamespace :: ImportDecl name -> ModuleName
-        asNamespace imp = fromMaybe (iDeclToModName imp) (ideclAs imp)
-#endif
+
         -- Full canonical names of imported modules
         importDeclerations = map unLoc limports
 
+        -- The list of all importable Modules from all packages
+        moduleNames = map showModName (GM.listVisibleModuleNames GHC.unsafeGlobalDynFlags)
```
lorenzo: Is getting global dynamic flags ok here?

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -245,33 +262,48 @@ instance ModuleCache CachedCompletions where
...
-        filtCompls = Fuzzy.filterBy label ident compls
+        filtCompls = Fuzzy.filterBy label prefixText compls
           where
-            compls = if T.null qualifier
+           compls = if T.null prefixModule
               then unqualCompls
-              else Map.findWithDefault [] qualifier qualCompls
+              else Map.findWithDefault [] prefixModule qualCompls
+
+        insertWithoutPrefix orig = J.detail ?~ orig
```
lorenzo: I'm not sure why this line requires `FlexibleContexts`

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
Bubba: Is it possible to use `getDynFlags` in `PluginUtils.hs` instead? https://github.com/haskell/haskell-ide-engine/blob/60d6bd40c295cbd3a7443fa7735d4f47dce8b0cd/src/Haskell/Ide/Engine/Plugin/HieExtras.hs#L55-L58

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
Bubba: Would adding a type signature to `insertWithoutPrefix` alleviate the need for it?

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -1,8 +1,8 @@
-{-# LANGUAGE CPP                 #-}
```
Bubba: Big +1 for getting rid of this!

lorenzo:
@Bubba I fixed those 2 things

Bubba:
LGTM, tested it out locally and works great!

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -141,16 +161,15 @@ instance ModuleCache CachedCompletions where
         showModName :: ModuleName -> T.Text
         showModName = T.pack . moduleNameString
 
-#if __GLASGOW_HASKELL__ >= 802
         asNamespace :: ImportDecl name -> ModuleName
         asNamespace imp = fromMaybe (iDeclToModName imp) (fmap GHC.unLoc $ ideclAs imp)
```
alanz: Using `fmap` is a neat way of avoiding CPP. Must put it in my mental toolbox.

PR #792, opened 2018-08-27
Always defer typed holes

### LINKED ISSUE #790: Should we always defer typed holes?
https://downloads.haskell.org/~ghc/8.4.3/docs/html/users_guide/glasgow_exts.html#typed-holes

At the moment in hie, we rely on the user to decide whether or not to defer typed holes.

This means that in the normal case (i.e. no special action taken), the presence of a typed hole will be treated as an error by GHC and it will terminate compilation, and return an error message.  This means we do not get a `TypeCheckedModule`, and a large chunk of useful information is not generated.

I think we should always pass `-fdefer-typed-holes` as a GHC option to ghc-mod, unless explicitly told not to in a config setting.

There may also be other similar settings we could use, to enable maximum information availability.

e.g. `-fdefer-type-errors`

Bubba:
Always sets the `DynFlag` '-fdefer-typed-holes'
Closes #790

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
+++ b/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
@@ -52,6 +52,7 @@ import qualified Language.Haskell.LSP.Types        as LSP
 import           Language.Haskell.Refact.API       (hsNamessRdr)
 
 import           DynFlags
+import           EnumSet
 import           GHC
 import           IOEnv                             as G
 import           HscTypes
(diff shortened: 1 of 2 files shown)


alanz:
I wonder if we shouldn't go whole hog on this, and do `-fdefer-type-errors`

Behaviour from this PR is

> The result is that a hole or a variable will behave like undefined, but with the added benefits that it shows a warning at compile time, and will show the same message if it gets evaluated at runtime.

Instead we could do

> This behaviour follows that of the `-fdefer-type-errors` option, which implies `-fdefer-typed-holes` and `-fdefer-out-of-scope-variables`.

Bubba:
`-fdefer-type-errors` causes errors to become warnings, so we would need to transform the type-related warnings back into errors

alanz:
So I think there is value in bringing this in as is, then doing the other as a separate step.

PR #794, opened 2018-08-28
Remove IdeResponse, split out IdeM into IdeM and IdeDeferM

Bubba:
- `codeActionProviders` can no longer defer
- Add `ifCachedModule`/`ifCachedModuleAndData` counterparts to `withCachedModule[AndData]`
- Remove `getCachedModule`/`isCached` to simplify/normalise how cached modules are accessed
- Remove `isStale` in `CachedModule` since it wasn't used

### DIFF
--- a/app/MainHie.hs
+++ b/app/MainHie.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RankNTypes   #-}
 module Main where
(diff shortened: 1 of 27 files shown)


Bubba:
@alanz @Infinisil do you still see the hang with this branch? I suspect it could be caused by HaRe's code actions, especially around here: https://github.com/haskell/haskell-ide-engine/blob/c6d95a258d20d33da01cbc52871f62d646cf061d/hie-plugin-api/Haskell/Ide/Engine/ArtifactMap.hs#L112-L137
Maybe it's getting stuck in a recursive loop?

Bubba:
Should code actions also return instantly? In `LanguageClient-neovim` the request is triggered by the user rather than automatically when moving the cursor like in vscode, which would make sense for it to be a slow-running deferred request.

alanz:
>  Should code actions also return instantly?

In vscode, and emacs-lsp, it is sent as part of the cursor movement, much like a hover request.

So speedy calculation is important.

alanz:
This is behaving well for me so far, will likely merge tomorrow.

PR #801, opened 2018-09-03
Running tests for MacOS in travis

lorenzo:
Compiles and runs tests for the 4 versions of GHC we currently support.

Currently the `func-tests` suite is disabled for GHC < 8.4.3 as there seems to be a bug in the suite that needs to investigated further.


alanz:
LGTM. I have requested the committee to enable travis for this repo

PR #802, opened 2018-09-04
Liquid plugin

alanz:
Add a plugin for liquid haskell.

This is disabled by default, and has to be enabled via a config setting (similar to `hlintOn`)

The `diagnostic provider` interface has changed, to include the concept of an async provider, which can run an external process, and then send results when it finishes, without blocking the dispatcher.

### DIFF
--- a/src/Haskell/Ide/Engine/LSP/Config.hs
+++ b/src/Haskell/Ide/Engine/LSP/Config.hs
@@ -1,9 +1,13 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Haskell.Ide.Engine.LSP.Config where
 
-import Data.Aeson
+import           Data.Aeson
+import qualified Data.Map  as Map
 import qualified Data.Text as T
-import Language.Haskell.LSP.Types
+import           Haskell.Ide.Engine.PluginsIdeMonads
+import           Language.Haskell.LSP.Types
+
+-- ---------------------------------------------------------------------
 
 -- | Callback from haskell-lsp core to convert the generic message to the
 -- specific one for hie
@@ -13,10 +17,13 @@ getConfigFromNotification (NotificationMessage _ _ (DidChangeConfigurationParams
     Success c -> Right c
     Error err -> Left $ T.pack err
 
+-- ---------------------------------------------------------------------
+
 data Config =
   Config
     { hlintOn             :: Bool
     , maxNumberOfProblems :: Int
+    , liquidOn            :: Bool
     } deriving (Show)
 
 instance FromJSON Config where
@@ -25,6 +32,7 @@ instance FromJSON Config where
     flip (withObject "Config.settings") s $ \o -> Config
       <$> o .:? "hlintOn"             .!= True
       <*> o .:? "maxNumberOfProblems" .!= 100
+      <*> o .:? "liquidOn"            .!= False
 
 -- 2017-10-09 23:22:00.710515298 [ThreadId 11] - ---> {"jsonrpc":"2.0","method":"workspace/didChangeConfiguration","params":{"settings":{"languageServerHaskell":{"maxNumberOfProblems":100,"hlintOn":true}}}}
 -- 2017-10-09 23:22:00.710667381 [ThreadId 15] - reactor:got didChangeConfiguration notification:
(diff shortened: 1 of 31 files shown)


src/Haskell/Ide/Engine/LSP/Config.hs
```diff
@@ -34,3 +42,20 @@ instance FromJSON Config where
...
 --                 {_settings = Object (fromList [("languageServerHaskell",Object (fromList [("hlintOn",Bool True)
 --                                                                                          ,("maxNumberOfProblems",Number 100.0)]))])}}
+
+instance ToJSON Config where
+  toJSON (Config h m l) = object [ "languageServerHaskell" .= r ]
+    where
+      r = object [ "hlintOn"             .= h
+                 , "maxNumberOfProblems" .= m
+                 , "liquidOn"            .= l
+                 ]
```
lorenzo: Can we add a unit test showing that `FromJSON` and `ToJSON` are symmetrical? I often run into bugs when I only change one of the instances and not the other

src/Haskell/Ide/Engine/LSP/Config.hs
```diff
@@ -34,3 +42,20 @@ instance FromJSON Config where
...
+
+-- ---------------------------------------------------------------------
+
+-- | For the diagnostic providers in the config, return a map of
+-- current enabled state, indexed by the plugin id.
+getDiagnosticProvidersConfig :: Config -> Map.Map PluginId Bool
+getDiagnosticProvidersConfig c = Map.fromList [("applyrefact",hlintOn c)
+                                              ,("liquid",     liquidOn c)
```
lorenzo: Is this brittany's formatting? looks off

src/Haskell/Ide/Engine/Plugin/Liquid.hs
```diff
@@ -4,8 +4,10 @@
 {-# LANGUAGE NamedFieldPuns      #-}
 module Haskell.Ide.Engine.Plugin.Liquid where
 
+import           Control.Concurrent.Async
 import           Control.Monad
 import           Control.Monad.IO.Class
+import           Control.Exception (bracket)
 #if __GLASGOW_HASKELL__ < 804
```
lorenzo: This line can now be removed

src/Haskell/Ide/Engine/Plugin/Liquid.hs
```diff
@@ -32,11 +38,13 @@ liquidDescriptor plId = PluginDescriptor
...
   , pluginDesc = "Integration with Liquid Haskell"
   , pluginCommands =
-      [ PluginCommand "sayHello" "say hello" sayHelloCmd
-      , PluginCommand "sayHelloTo ""say hello to the passed in param" sayHelloToCmd
+      [ PluginCommand "sayHello"   "say hello"                        sayHelloCmd
+      , PluginCommand "sayHelloTo" "say hello to the passed in param" sayHelloToCmd
       ]
   , pluginCodeActionProvider = Nothing
-  , pluginDiagnosticProvider = Just (DiagnosticProvider (S.singleton DiagnosticOnSave) diagnosticProvider)
+  , pluginDiagnosticProvider= Just (DiagnosticProvider
```
lorenzo: Alignment looks off here as well

test/functional/LiquidSpec.hs
```diff
@@ -0,0 +1,112 @@
...
+-- import           Language.Haskell.LSP       as LSP
+import           Language.Haskell.LSP.Types as LSP hiding (contents, error )
+import           Haskell.Ide.Engine.LSP.Config
+import           Test.Hspec
+import           TestUtils
+import           Utils
+
+-- ---------------------------------------------------------------------
+
+spec :: Spec
```
lorenzo: Can this be guarded by LH being installed? That or we need to also install it in travis

src/Haskell/Ide/Engine/LSP/Config.hs
lorenzo: Nevermind, I read further and the test seems to be present already

src/Haskell/Ide/Engine/Plugin/Liquid.hs
alanz: ok

alanz: Except it fails -Wall on GHC 8.4.3 if the #if is removed

src/Haskell/Ide/Engine/LSP/Config.hs
Bubba: Why does this need a `ToJSON` instance? Not opposed to it just curious

alanz: For the quickcheck roundtrip tests

alanz:
@Bubba updating the default logging to be off is fine.

In practical terms, I will still have two configs, one for logging, one for not, so that I can turn it on while debugging an individual test.

PR #810, opened 2018-09-07
[WIP] Add snippet support

### LINKED ISSUE #398: Support snippets
https://github.com/Microsoft/language-server-protocol/blob/master/snippetSyntax.md

They are a part of LSP protocol and a requested feature of vscode-hie-server:

https://github.com/alanz/vscode-hie-server/issues/32

Bubba:
Closes #398 
This would add snippet for the arguments to a function on completion.
# Todo
- [ ] Get it working for unsaturated types

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -49,6 +49,7 @@ import qualified DynFlags                                     as GHC
 import           Packages
 import           SrcLoc
 import           TcEnv
+import           Type
 import           Var
 
 getDynFlags :: CachedModule -> DynFlags
@@ -73,9 +74,9 @@ instance ModuleCache NameMapData where
 data CompItem = CI
   { origName     :: Name
   , importedFrom :: T.Text
-  , thingType    :: Maybe T.Text
+  , thingType    :: Maybe Type
   , label        :: T.Text
-  } deriving (Show)
+  }
 
 instance Eq CompItem where
   (CI n1 _ _ _) == (CI n2 _ _ _) = n1 == n2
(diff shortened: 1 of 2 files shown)


alanz:
Interesting

Bubba:
Should there be a config option to disable this? The current behaviour is that when a function is autocompleted snippets for its arguments are inserted with labels for their types. But if you're currying/writing a lot of point-free code I could imagine this could get pretty annoying pretty quickly.
(It would be amazing if we could automatically deduce wether to show these snippets based of the type of the function that the user is currently implementing)

alanz:
I think anything that is likely to spam the user should have a disable option.

lorenzo:
it can also be annoying for neovim users, as the client does not properly implement snippets

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -261,8 +286,8 @@ instance ModuleCache CachedCompletions where
       , importableModules = moduleNames
       }
 
-getCompletions :: Uri -> PosPrefixInfo -> IdeDeferM (IdeResult [J.CompletionItem])
-getCompletions uri prefixInfo = pluginGetFile "getCompletions: " uri $ \file -> do
+getCompletions :: Uri -> PosPrefixInfo -> Bool -> IdeDeferM (IdeResult [J.CompletionItem])
```
lorenzo: Can we wrap that `Bool` in a newtype? Give that we don't have much docs it would be useful to read that the Bool is for

bubba: Would some haddock documentation suffice instead? The setting is read in as a bool from JSON so it would be nice to not have to convert it

alanz:
It does not look like there is a test that has snippets disabled?

PR #811, opened 2018-09-08
Hover response can now be null

### LINKED ISSUE #804: Returning malformed hover response when empty
At the moment we return a hover response of

```json
{"result":{"contents":[]},"jsonrpc":"2.0","id":2}
```
when we have no hover information.

This should be

```json
{"result":null,"jsonrpc":"2.0","id":2}
```
The empty contents is confusing the hover renderer in emacs-lsp.

See https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#hover-request-leftwards_arrow_with_hook

alanz:
Closes #804

### DIFF
--- a/src/Haskell/Ide/Engine/Transport/LspStdio.hs
+++ b/src/Haskell/Ide/Engine/Transport/LspStdio.hs
@@ -475,7 +475,9 @@ reactor inp = do
                 -- TODO: maybe only have provider give MarkedString and
                 -- work out range here?
                 let hs = concat hhs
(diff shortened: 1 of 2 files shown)


alanz:
The changes on haskell-lsp are not against master, as lsp-test still needs to be updated to use master.

lorenzo:
Is this a fix for #804?

alanz:
@lorenzo yes, thanks for the prompt

Bubba:
Should we create a PR to haskell-lsp for these changes?

PR #815, opened 2018-09-09
Cache parsed modules

bubba:
This PR reworks large parts of the caching API:

- `UriCache` has been split up into several parts:
  1. `UriCacheResult` is now used to make it more clear the distinction between `UriCache` and `UriCacheFailed` (previously one was a record and the later wasn't, which was a horrible idea)
  2. `CachedModule` is now replaced with `CachedInfo` which **doesn't** contain the actual GHC module - this is held separately inside `UriCache`
- `CacheableModule` is a new class which `TypecheckedModule` and `ParsedModule` are instances of. `UriCache` stores both and the class defines a function for extracting the module from it. I would like to bike shed on this name (and others), since the naming around `ModuleCache` and `GhcModuleCache` is getting confusing
- `ifCachedModule`/`withCachedModule` are now polymorphic in which type of module from GHC they should take.

`withCachedModuleAndData` still waits for typechecked modules. Would their be use cases of caching data generated from the parsed module?

The first significant benefit we get from this is now document symbols only rely on being able to parse the module, and hie can now respond to requests even when the module doesn't typecheck and there is no cache for it available. 

The next milestone would be to cache the renamed module separately from the typechecked module, but as @alanz has pointed out this may be difficult due to how GHC intertwines the two processes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,9 +22,9 @@ defaults: &defaults
 
     - restore_cache:
         keys:
-          - stack-cache-07-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "stack-build.txt" }}-{{ checksum "all-cabal.txt" }}
-          - stack-cache-07-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "stack-build.txt" }}
-          - stack-cache-07-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "resolver.txt" }}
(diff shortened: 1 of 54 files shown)


alanz:
The principle looks good.  What are your concerns about the cache naming?

bubba:
@alanz I think having a separate `ModuleCache` and `GhcModuleCache` modules is confusing, I would like to see them just merged into the one. I'd also like to see `GhcModuleCache` renamed to something like `AllCaches`/`GlobalCaches`, but I see there's a to-do to move that into ghc-mod-core. I'm not sure what parts would be able to go into ghc-mod though, since any `UriCache` stuff seems specific to hie.

Issue #680, opened 2018-07-12
HIE seems to be able to typechek only one module


maksbotan:
I use HIE with LanguageClient_neovim. Versions:

```
$ ~/.local/bin/hie --version
Version 0.2.0.0, Git revision e159e35a5471bcfa8576d663cb82180dc9fe53ab (dirty) (1555 commits) x86_64 ghc-8.2.2
$ stack --version
Version 1.7.1, Git revision 681c800873816c022739ca7ed14755e85a579565 (5807 commits) x86_64 hpack-0.28.2
```

I build HIE with `$ stack --resolver=lts-11.2 --stack-yaml=stack-8.2.2.yaml install`, since we have `LTS-11.2` in the project I develop.

The project consists of a library and an executable in separate Cabal files. The executable is build with `-dynamic`, so I also have to add `-fexternal-interpreter` to its Cabal file.

However, it seems that HIE can typecheck only the first file I open in nvim. All subsequent files result in 'didn't get typechecked module`. Here's a part of log with `--debug --vomit`:

```
2018-07-12 23:45:52.549029635 [ThreadId 4] - ---> {"jsonrpc":"2.0","method":"textDocument/didOpen","params":{"textDocument":{"languageId":"haskell","text":" ...
2018-07-12 23:45:52.550559802 [ThreadId 10] - ****** reactor: got message number:3
2018-07-12 23:45:52.55069708 [ThreadId 10] - ****** reactor: processing NotDidOpenTextDocument
2018-07-12 23:45:52.5508884 [ThreadId 9] - ghcDispatcher:got request 3 with id: Nothing
2018-07-12 23:45:52.749975698 [ThreadId 9] - file mapping state is: fromList [("/../MainWindow.hs",FileMapping {fmPath = "/tmp/ghc-mod24974/MainWindow24973-0.hs", fmTemp = True}),("/.../WalletTree.hs",FileMapping {fmPath = "/tmp/ghc-mod24975/WalletTree24973-1.hs", fmTemp = True})]
2018-07-12 23:45:52.750217547 [ThreadId 9] - ghcDispatcher: top of loop
2018-07-12 23:45:52.750346578 [ThreadId 9] - ghcDispatcher:got request 3 with id: Nothing
2018-07-12 23:45:52.750479733 [ThreadId 9] - ghcDispatcher:Processing request as version matches
2018-07-12 23:45:52.806597695 [ThreadId 9] - ghcDispatcher: top of loop
[...]

Issue #662, opened 2018-06-27
Message box saying "Got error while processing diagnostics: Failing due to -Werror." in Sublime


Pastafarianist:
I am using Sublime and I am working with a codebase I cannot edit freely for _reasons_. Anyway, while browsing code, often a message box pops up that says

```
Got error while processing diagnostics: 
Failing due to -Werror.
```

This is annoying and useless. It would be much more preferable to see a specific list of warnings in the box at the bottom of the window.

I'm not sure though whether the issue is in HIE or in Sublime LSP client, so if this is the wrong issue tracker to post this, I will move it to the LSP client tracker.

Pastafarianist:
Looks like a dupe of https://github.com/haskell/haskell-ide-engine/issues/391.

alanz:
This should be fixed via #817

PR #817, opened 2018-09-10
Ignoring the -Werror setting

### LINKED ISSUE #449: Manage -Werror
If hie has been built via stack install, and you open it in hie, then errors per file are not shown because pedantic mode  enables `-Werror`, and this complains about something in the cabal file.

It probably makes sense to actively turn off `-Werror` in all cases.

lorenzo:
When having -Werror enabled, the experience with HIE degrades
significantly as everything is treated as a type error. This makes sure
that warnings are always treated as warnings.

closes #449, #680 and #662

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@
 /test/testdata/redundantImportTest/stack.yaml
 /test/testdata/completion/stack.yaml
 /test/testdata/definition/stack.yaml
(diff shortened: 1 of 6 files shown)


alanz:
LGTM, I have been wanting this too for a while. Especially when working on hie.

alanz:
Merge when the tests pass.

PR #825, opened 2018-09-13
Update to haskell-lsp-0.8 and fix snippet not respecting client capabilities

### LINKED ISSUE #822: Snippet support dosen't respect client declining the feature.
When using HIE with neovim, which unfortunately docent support snippets in a good way, I tried deregistering:
The LSP spec has a ```snippetSupport``` option in ```TextDocumentClientCapabilities``` and it seems that isn't respected by HIE.

I saw that there is global config to turn it of but it seems more logical to do it on a client basis according to the spec

bubba:
Closes #822 
The haskell-lsp-0.8 update mainly involves importing/removing Lens imports.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,6 +21,3 @@
 [submodule "submodules/haskell-lsp"]
 	path = submodules/haskell-lsp
 	url = https://github.com/alanz/haskell-lsp.git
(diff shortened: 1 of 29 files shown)


alanz:
LGTM

PR #831, opened 2018-09-15
Debouncing the call to diagnostics

lorenzo:
This avoids constant recompilation while typing by only dispatching the
recompilation for thhe last event after a short timeout.

I'm personally very execited about this, since it solves one of the
biggest problems I had with hie: it kept the laptops fans spinning a
full speed.

I'm not yet sure all tests pass for this, I'd like first to open the
discussion on whether or not this is a reasonable approach.


src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -97,33 +105,50 @@ run
...
+            dispatchDiagnostics (Just (MostRecent req)) = performDiagnostics req
+
+        -- Debounces messages published to the diagnostics channel.
+        let diagnosticsQueue tr = forever $ do
+              inval <- liftIO $ atomically $ readTChan diagIn
+              Debounce.send tr (Just $ MostRecent inval)
 
+        tr <- Debounce.new
+          (Debounce.forMonoid $ react . dispatchDiagnostics)
+          (Debounce.def { Debounce.delay = 350000, Debounce.alwaysResetTimer = True })
```
alanz: This number means 0.35sec, if I understand it correctly.

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -730,8 +757,18 @@ reactor inp = do
 
 -- ---------------------------------------------------------------------
 
+-- | Queue a diagnostics request to be performed after a timeout. This prevents recompiling
+-- too often when there is a quick stream of changes.
 requestDiagnostics :: DiagnosticTrigger -> TrackingNumber -> J.Uri -> J.TextDocumentVersion -> R ()
-requestDiagnostics trigger tn file mVer = do
+requestDiagnostics dt tn uri mVer= do
```
alanz: Given we now have `performDiagnostics`, `requestDiagnosticsNormal` and `requestDiagnostics`, it is getting confusing.

Perhaps we should do something like
- `requestDiagnostics` becomes `queueDiagnosticsRequest`
- `performDiagnostics` goes back to being `requestDiagnostics`

alanz:
Looks like a neat implementation, TIL `Control.FoldDebounce`.

src/Haskell/Ide/Engine/Transport/LspStdio.hs
lorenzo: that's correct

src/Haskell/Ide/Engine/Transport/LspStdio.hs
lorenzo: Good idea

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -98,33 +107,49 @@ run
...
         versionTVar <- atomically $ newTVar Map.empty
+        diagIn      <- atomically newTChan
         let dEnv = DispatcherEnv
               { cancelReqsTVar = cancelTVar
               , wipReqsTVar    = wipTVar
               , docVersionTVar = versionTVar
               }
-        let reactorFunc = runReactor lf dEnv cin diagnosticProviders hps sps
-              $ reactor rin
+        let react = runReactor lf dEnv cin diagIn diagnosticProviders hps sps
```
bubba: Are there any caveats to look out for when running a second instance of R ?

alanz: Good point.  In principle it need not be in R, if we encapsulate things properly

lorenzo: runReactor is almost an alias for runReaderT, so only one reactor is actually running at a time.

alanz: And the TVar updates are protected by `atomically`, so I guess at first blush it should be ok.

The danger comes if we ever change the definition of `R`, and forget/do not realise that it is used twice.

alanz: @bubba what do you think?

bubba: Is it possible to just pass the diagnostics chan to `queueDiagnosticsRequest`/`Debounce.forMonoid` without the rest of the stuff in `R`?

lorenzo: I can try removing the new chan from `R` if you both agree it is a good idea.

alanz: I think not running two instances of a monad with a shared state is a good idea.

I *do* think that in this case it is safe, given that it is a `runReaderT`, and we are using `TVar`s which are accessed atomically.  But that is only the case now, things may change in future.

alanz: And my words are not all that clear. I am with @bubba on this one.

lorenzo: Cool, I'll change it then

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -98,33 +123,49 @@ run
...
+        let dispatchDiagnostics :: Option (MostRecent DiagnosticsRequest) -> R ()
+            dispatchDiagnostics req = option (pure ()) (requestDiagnostics . coerce) req
+
+        -- Debounces messages published to the diagnostics channel.
+        let diagnosticsQueue tr = forever $ do
+              inval <- liftIO $ atomically $ readTChan diagIn
+              Debounce.send tr (coerce . Just $ MostRecent inval)
 
+        tr <- Debounce.new -- Debounce for 350ms
+          (Debounce.forMonoid $ react . dispatchDiagnostics)
```
lorenzo: I can't seem to get rid of the need for running this in the `R` monad. The reason is that `requestDiagnostics` lives there, and I need to call that function from inside the Debounce callback.

There are some gymnastics I can make, like duplicating the code and creating a separate `REnv` that I can pass as a parameter to `requestDiagnostics`, but that seems a lot dirtier than doing this. I any suggestions @alanz @bubba ?

alanz: Theoretically a reader environment should be safe.  The complications happen when you have shared vars in it.  Which are things I have mixed feelings about at the best of times.

bubba: I'm not quite sure why `dispatcherEnv` is even inside `REnv`. I'm going to try pulling it out, see what happens

bubba: Looks like they're very tightly coupled. There's a lot of dispatch related stuff like cancelling requests etc. going on in LspStdio. I think we should maybe try moving that stuff into Dispatcher so we can hide the `TVars`.

lorenzo: Can we do that as a separate task and not mix it with this PR?

alanz: I am happy to merge it as is, provided we make an issue for the cleanup.

@bubba ?

PR #834, opened 2018-09-16
Fix completions not returning anything when there's no prefix

bubba:
Might want to turn on whitespace ignore in the diff


alanz:
LGTM, merge when CI is happy

PR #841, opened 2018-09-24
macOS deployment via TravisCI and Github Releases

bubba:
This adds testing and deployment via GitHub Releases for macOS builds. Releases will only be deployed whenever a tag is pushed. Hopefully in the future we will be able to tar the separate binaries together. 

Please squash this!


lorenzo:
I will refactor this to use the stages we had before, so that we can avoid the 50 minutes time limit

alanz:
@lorenzo over to you

PR #847, opened 2018-09-27
Completion contexts

### LINKED ISSUE #829: Autocompletion should avoid snippets if completing parts of a type signature
When completing a type signature and there are types in scope sharing their names with one of the constructors, you can easily get unwanted snippets in the autocompletion. For instance:

```haskell
get :: Uri
```
`Uri ` is both a type and a constructor, so I pick  the constructor without realising:

```haskell
get :: Uri Text
```

The `Text` placeholder is unwanted. There are 2 possibilities here:

1. Entirely disable snippets if we can detect we are completing a type signature
2. Avoid suggesting things other than types and classes if we can detect we are completing a signature. This may be unwanted for `-XDataKinds`

bubba:
Adds the idea of "contexts" when triggering code completion, by generating an interval map that stores whether or not the current declaration is a type or value declaration. This allows for code completions to be filtered by type or values, and can be further expanded on to include other types of declarations such as instances etc.
Fixes #829 

# Todo
- [x] Add tests


alanz:
Conceptually this is a good idea.

I am just wondering about generating a context map for the entire AST, every time it is parsed.

i.e. would it make sense to instead have a function that takes a location and the parsed source and returns the context?  So we only run it when needed.

lorenzo:
I think generating the map on demand sounds better, otherwise this looks great!

bubba:
Currently trying to fix an issue where extending a type signature (e.g. adding a `-> c` to `a -> b`) doesn't bring up the `TypeContext`, I think its related to how the old/new position mapping doesn't do anything for changes on the same line

test/functional/CompletionSpec.hs.orig
```diff
@@ -0,0 +1,191 @@
+{-# LANGUAGE OverloadedStrings #-}
```
lorenzo: This file show not be here :)

bubba: Ugh I hate merge commits 😩

test/functional/CompletionSpec.hs.orig
```diff
@@ -0,0 +1,191 @@
...
+    compls <- getCompletions doc (Position 0 24)
+    let item = head $ filter ((== "OverloadedStrings") . (^. label)) compls
+    liftIO $ do
+      item ^. label `shouldBe` "OverloadedStrings"
+      item ^. kind `shouldBe` Just CiKeyword   
+  
+  it "completes with no prefix" $ runSession hieCommand fullCaps "test/testdata/completion" $ do
+    doc <- openDoc "Completion.hs" "haskell"
+    _ <- skipManyTill loggingNotification (count 2 noDiagnostics)
+<<<<<<< HEAD
```
Anrock: Unresolved merge conflict.

bubba:
@alanz @Lorenzo @Anrock this is ready for re-review

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -294,14 +294,20 @@ updatePositionMap uri changes = pluginGetFile "updatePositionMap: " uri $ \file
     cacheInfoNoClear file info'
     return $ IdeResultOk ()
   where
-    f (+/-) (J.Range (Position sl _) (Position el _)) txt p@(Position l c)
+    f (+/-) (J.Range (Position sl sc) (Position el ec)) txt p@(Position l c)
       | l < sl = Just p
       | l > el = Just $ Position l' c
+      | l == sl && c <= sc = Just p
+      | l == sl && l == el && c <= nec && newL == 0 = Just $ Position l ec
+      | l == sl && l == el && c > nec && newL == 0 = Just $ Position l (c +/- (nec - sc))
```
lorenzo: Can you add a comment on what each case is covering?

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -34,10 +36,11 @@ import           DataCon
 import           Exception
 import           FastString
 import           Finder
-import           GHC
+import           GHC                                          hiding (getContext)
```
lorenzo: It would be nice if we could import this qualified instead

bubba:
I think this PR should probably be merged by now - This corner test case is causing some major headache and is going to require a revisit to how we parse and cache the results from GHC. Would it be a reasonable to comment out this test case for now and create a new issue for it?

src/Haskell/Ide/Engine/Transport/LspStdio.hs
bubba: Added some ASCII diagrams!

bubba:
To clarify it's the same issue as before: *sometimes* the position doesn't get mapped back correctly when extending a line, causing the wrong completion context to be applied, causing value code completions to show up. But this only seems to happen if it takes too long to typecheck the module. Adding a delay inside `setTypecheckedModule` causes the test to fail consistently. Otherwise it just acts flaky locally and fails on CircleCI

PR #864, opened 2018-10-09
Refactor DispatcherEnv our of the Reactor

### LINKED ISSUE #835: Separate dispatcher stuff away from LspStdio
Currently `R` inside LspStdio holds a lot of mutable vars that are used by the dispatcher, which has caused some pains in #831. This should be tidied up so that these vars are handled inside `Dispatcher.hs` exclusively, so that multiple instances of `R` can be safely run.

lorenzo:
This introduces a new `Scheduler` module that that takes much of the code previously in the Dispatcher and removes the channel interface between the reactor and the dispatcher. Instead, explicitly calls to the Scheduler are made to queue requests.

closes #835


src/Haskell/Ide/Engine/Scheduler.hs
```diff
@@ -0,0 +1,197 @@
...
+     -- ^ The scheduler to run.
+  -> ErrorHandler
+     -- ^ A handler for any errors that the dispatcher may encounter.
+  -> CallbackHandler m
+     -- ^ A handler to run the requests' callback in your monad of choosing.
+  -> C.ClientCapabilities
+     -- ^ List of features the IDE client supports or has enabled.
+  -> IO ()
+runScheduler Scheduler {..} errorHandler callbackHandler caps = do
+  let dEnv = DispatcherEnv
```
bubba: It's great that `cancelTVar`/`wipTVar` are no longer exposed! Could we go one step further and maybe merge Dispatcher/Scheduler so that `DispatcherEnv` is also just an internal structure? It would be really nice if it wasn't exported at all

bubba: Does any other module other than `Scheduler` import `Dispatcher`?

lorenzo: I don't think anything else is importing the dispatcher, I could move the full contents of it into the scheduler now. Do you think I should do that here?

bubba: That sounds good, we'll probably need to do a little bikeshedding about the naming at some point

lorenzo: @bubba I moved all the dispatcher code into the Scheduler, now the DispatcherEnv record is hidden

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -115,35 +113,26 @@ instance Semigroup (MostRecent a) where
...
   let dp lf = do
-        cancelTVar  <- atomically $ newTVar S.empty
-        wipTVar     <- atomically $ newTVar S.empty
-        versionTVar <- atomically $ newTVar Map.empty
         diagIn      <- atomically newTChan
-        let dEnv = DispatcherEnv
-              { cancelReqsTVar = cancelTVar
-              , wipReqsTVar    = wipTVar
-              , docVersionTVar = versionTVar
-              }
```
bubba: Getting rid of this is great

lorenzo:
@alanz are you ok with merging this?

alanz:
I would like to take a proper look, when I can concentrate.  It may be a few days.

alanz:
@lorenzo is this up to date? I want to start looking at it.

lorenzo:
Yes, it is up to date, a few commits behind, but nothing relevant in master that would affect this @alanz

app/MainHie.hs
```diff
@@ -146,8 +144,8 @@ run opts = do
...
   -- launch the dispatcher.
   if optJson opts then do
-    pin <- atomically newTChan
-    jsonStdioTransport (dispatcherP pin plugins' ghcModOptions) pin
+    scheduler <- newScheduler plugins' ghcModOptions
+    jsonStdioTransport scheduler
   else do
-    pin <- atomically newTChan
-    lspStdioTransport (dispatcherP pin plugins' ghcModOptions) pin origDir plugins' (optCaptureFile opts)
+    scheduler <- newScheduler plugins' ghcModOptions
```
alanz: This line could be moved out of the if statement, it is in both legs.

src/Haskell/Ide/Engine/LSP/Reactor.hs
```diff
@@ -81,31 +79,38 @@ reactorSend' f = do
...
+-- | Updates the version of a document and then sends the request to be processed
+-- asynchronously.
+updateDocumentRequest
+  :: (MonadIO m, MonadReader REnv m) => Uri -> Int -> PluginRequest R -> m ()
+updateDocumentRequest uri ver req = do
+  sc <- asks scheduler
+  liftIO $ Scheduler.sendRequest sc (Just (uri, ver)) req
+
+-- | Marks a s requests as cencelled by its LspId
+cancelRequest :: (MonadIO m, MonadReader REnv m) => J.LspId -> m ()
```
alanz: I would expect these functions (all the ones that do `asks scheduler` to move to the `Scheduler` module.
Would that make sense?

alanz: Yes, but I suspect the only constraint we really need is that the `asks` call succeeds.

To me it is about putting all the stuff related to scheduling in one place, so we can look at it holistically.

lorenzo: How would that look?

alanz: And looking more closely, that is exactly what you proposed. Sorry.

lorenzo: @alanz I moved the implementation of those functions to the `Scheduler` module. Decided to keep the same functions in the `Reactor` module in order to provide some more specific type signatures. Using the generic function from `Scheduler` in `LspStdio` required adding many annoying type hints. In my opinion this version looks cleaner.

src/Haskell/Ide/Engine/LSP/Reactor.hs
```diff
@@ -81,31 +83,34 @@ reactorSend' f = do
...
+-- | Updates the version of a document and then sends the request to be processed
+-- asynchronously.
+updateDocumentRequest
+  :: (MonadIO m, MonadReader REnv m) => Uri -> Int -> PluginRequest R -> m ()
+updateDocumentRequest = Scheduler.updateDocumentRequest
+
+-- | Marks a s requests as cencelled by its LspId
+cancelRequest :: (MonadIO m, MonadReader REnv m) => J.LspId -> m ()
+cancelRequest lid =
+  liftIO . flip Scheduler.cancelRequest lid =<< asks scheduler
```
lorenzo: This was the only function that I could not generalise, I kept getting type errors in the `Scheduler` module.

alanz:
LGTM, thanks. I reckon we can merge when the CI passes

lorenzo:
I'll merge now, tests passed for circle-ci

PR #869, opened 2018-10-11
Add example code for haskellforall.com blog post

alanz:
http://www.haskellforall.com/2018/10/detailed-walkthrough-for-beginner.html

As demonstrated at haskellX 2018.

A lot of this is plumbing through access to the VFS to be able to get the source text for a LSP `Range`, which is useful for all plugins anyway.

The actual work is in `HfaPlugin.hs`, where only the first 75 lines are for the plugin the rest is the code from @Gabriel439 's blog post.


alanz:
Does anyone know what is wrong with the appveyor build? It seems stack is no longer installed?

Issue #867, opened 2018-10-11
Autocompletion for language extensions


lorenzo:
It would be nice if hie could complete language extensions when it can detect the right completion context

cronokirby:
What would be the place to look for adding this?

lorenzo:
Look at how completions for import modules are done:

https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/Plugin/HieExtras.hs#L328-L335

The list of visible modules is taken from the ghc-mod api. You could take the list of all available Language extensions from the same api, here is how it was done for the modules list:

https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/Plugin/HieExtras.hs#L198

And this is how it could be done for the language extensions:

https://github.com/DanielG/ghc-mod/blob/master/GhcMod/Exe/Lang.hs#L10

@cronokirby let me know if you want to volunteer for this task and if you need further help.

PR #882, opened 2018-10-16
Add completion for Language Extensions

cronokirby:
First try at resolving #867.

This adds completion for language extensions.

This changes the behaviour of completion generation to only generate completions for language extensions on lines beginning with "{-# LANGUAGE", perhaps we should modify this so it works even if the line just contains "LANGUAGE": `{-#LANGUAGE` and `{-#   LANGUAGE` would then also work.

This adds a `cachedExtensions` field to the `CachedCompletions` record, which is filled via the Ghc api. Going through GhcMod would be much more complicated, since [in ghcmod](https://github.com/DanielG/ghc-mod/blob/master/GhcMod/Exe/Lang.hs#L10) a nice list of language extensions is squashed into a string, which we'd have to parse back. Instead I just get the list via [this](https://hackage.haskell.org/package/ghc-8.6.1/docs/DynFlags.html#v:supportedLanguagesAndExtensions)

I wasn't sure what kind of [completion](https://github.com/alanz/haskell-lsp/blob/2bdec6b34f7fa76b71c4d0931557f64ae9077c75/haskell-lsp-types/src/Language/Haskell/LSP/Types/Completion.hs#L22) to give Language Extensions, so I went with `CiModule` since it seemed the most appropriate

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -343,9 +343,11 @@ getCompletions uri prefixInfo (WithSnippets withSnippets) = pluginGetFile "getCo
       filtExtensionCompls = filtListWith mkExtCompl cachedExtensions
 
       result
-        | "import " `T.isPrefixOf` fullLine      = filtImportCompls
-        | "{-# LANGUAGE" `T.isPrefixOf` fullLine = filtExtensionCompls
-        | otherwise                              =
+        | "import " `T.isPrefixOf` fullLine =
+          filtImportCompls
+        | "{-# language" `T.isPrefixOf` T.toLower fullLine = 
+          filtExtensionCompls
+        | otherwise =
           filtModNameCompls ++ map (toggleSnippets . mkCompl) filtCompls
     in return $ IdeResultOk result
 -- ---------------------------------------------------------------------
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -141,7 +141,7 @@ mkModCompl label =
 
 mkExtCompl :: T.Text -> J.CompletionItem
 mkExtCompl label =
-  J.CompletionItem label (Just J.CiModule) Nothing
+  J.CompletionItem label (Just J.CiKeyword) Nothing
     Nothing Nothing Nothing Nothing Nothing Nothing Nothing
     Nothing Nothing Nothing Nothing Nothing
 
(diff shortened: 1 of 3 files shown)


lorenzo:
I think this is looking good. Could you also add a test? This is the test for the import module completions:

https://github.com/haskell/haskell-ide-engine/blob/master/test/functional/CompletionSpec.hs#L34-L47

It uses this file for comparisons:

https://github.com/haskell/haskell-ide-engine/blob/master/test/testdata/completion/Completion.hs


You can run the functional tests with:

```bash
stack test haskell-ide-engine:func-test
```

cronokirby:
I've added a test to the suite, I've also played around with it on some of my projects and it works quite well

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -139,6 +139,12 @@ mkModCompl label =
     Nothing Nothing Nothing Nothing hoogleQuery
   where hoogleQuery = Just $ toJSON $ "module:" <> label
 
+mkExtCompl :: T.Text -> J.CompletionItem
+mkExtCompl label =
+  J.CompletionItem label (Just J.CiModule) Nothing
```
lorenzo: I think changing it to `CiKeyword` makes more sense, but if you think this is just bike shedding, then feel free to disregard :)

```suggestion
  J.CompletionItem label (Just J. CiKeyword) Nothing
```

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -322,16 +333,21 @@ getCompletions uri prefixInfo (WithSnippets withSnippets) = pluginGetFile "getCo
...
+                             | label <- Fuzzy.simpleFilter fullPrefix list
+                             , enteredQual `T.isPrefixOf` label
+                             ]
+
+      filtImportCompls = filtListWith mkImportCompl importableModules
+      filtExtensionCompls = filtListWith mkExtCompl cachedExtensions
+
+      result
+        | "import " `T.isPrefixOf` fullLine      = filtImportCompls
+        | "{-# LANGUAGE" `T.isPrefixOf` fullLine = filtExtensionCompls
```
alanz: This is fine for most use cases, but technically the keyword `LANGUAGE` is parsed to be case insensitive.

src/Haskell/Ide/Engine/Plugin/HieExtras.hs
cronokirby: I agree, this does look better.

alanz:
I just tried this, and it works great.

A possible tweak would be to recognise `{-#` and offer to complete with `LANGUAGE` or `OPTIONS_GHC`. and whatever else is valid.

lorenzo:
I’ll create a ticket for that

PR #885, opened 2018-10-18
Use internal library for testing

bubba:
Now that stack 1.9.1 fixes bug https://github.com/commercialhaskell/stack/issues/4265


lorenzo:
I guess you will need to bust the cache for the CI builders

bubba:
I think we need to upgrade the stack version on the image too

alanz:
And maybe we should wait a week or two to let more people update to stack 1.9.1 first

bubba:
We’ll need to upgrade the circle CI docker image to use stack 1.9.1.
Should it be enough to just add a `RUN stack upgrade` to the Dockerfile?

iGOTCHA34:
I need to get some info about how this works and what is the BEST way to do IT

alanz:
LGTM, merge when the CI passes

PR #892, opened 2018-10-20
Strip forall when completing simple types

### LINKED ISSUE #367: Implicit forall
Now neovim displays `print` in completion dialog as

```
print     [Function] [LC] forall a. Show a => a -> IO ()^@Prelude
```
Why not cut the `forall` for rank-1 types? It takes space in the list of completions so the signature gets truncated:
```
writeIORef         [Function] [LC] forall a. IORef a -> a ->..()^@Data.IORef
```
Also I wonder if it's feasible to suppress module names as well if someone wants to see longer signatures at the cost of less info.

cronokirby:
resolves #367 

Modify completion so that the `forall a.` part of a simple
signature is stripped.

Higher ranked quantification, e.g. `(forall a. a -> a) -> Int`
isn't affected by this change.

Higher kinded types are also stripped e.g `forall (m :: * -> *).`
will be stripped.

One "edge case" of sorts are when types of multiple kinds are involved:
before, `return`, for example, was completed with:
```haskell
forall (m :: * -> *). Monad m => forall a. a -> m a
```
there's 2 quantifications there. With the somewhat naive way I've implemented forall stripping,
this becomes:
```haskell
Monad m => forall a. a -> m a
```

Should there have been 2 foralls in our completion in the first place?
Can we assume there will only ever be a situation with a single constraint arrow `=>` followed by a forall, so that we can simple strip once, move past the `=>` and the strip again? or will that not work?

Otherwise it works for the use cases described in the issue, but these edge cases might be worth looking into in this PR, or in another issue.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -119,10 +119,15 @@ mkCompl CI{origName,importedFrom,thingType,label} =
           | otherwise = label <> " " <> argText
         argText :: T.Text
         argText =  mconcat $ List.intersperse " " $ zipWith snippet [1..] argTypes
+        stripForall t 
+          | T.isPrefixOf "forall" t =
+            -- We drop 2 to remove the '.' and the space after it
+            T.drop 2 (T.dropWhile (/= '.') t)
+          | otherwise               = t
         snippet :: Int -> Type -> T.Text
         snippet i t = T.pack $ "${" <> show i <> ":" <> showGhc t <> "}"
         typeText
-          | Just t <- thingType = Just $ T.pack (showGhc t)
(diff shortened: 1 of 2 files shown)


alanz:
So we are wanting changes to this?

On the one hand it does something that will be mostly useful, on the other we can use the actual AST and pretty printer to get a more precise version. 

I am not sure the extra work is worth it right now, maybe make an additional ticket to improve it?

lorenzo:
I think this is good to merge and creating a ticket for future improvements makes sense

bubba:
I'm happy to merge this

lorenzo:
Seems like we all agree to merge this, so I'll go ahead and also create another ticket for the future improvements.

lorenzo:
thanks @cronokirby!

PR #895, opened 2018-10-22
Completion for pragmas

### LINKED ISSUE #884: Autocomplete for file pragmas
We should provide autocompletion after writing `{-# `. Some of the valid options

- `{-# LANGUAGE`
- `{-# OPTIONS_GHC `
- `{-# INLINE `
- `{-# NOINLINE `
- `{-# INLINABLE `
- `{-# WARNING `
- `{-# DEPRECATED `
- `{-# ANN `
- `{-# RULES `
- `{-# SPECIALIZE `
- `{-# SPECIALIZE INLINE`
- `{-# UNPACK `

Ideally, it should detect if the pragma is already closed in the same line `#-}`. If it is not already closed, it should add the completion and add the closing sequence automatically.

Some of the pragmas are context sensitive, for a first initial version of this feature maybe we don't need to detect the contexts perfectly, and just autocomplete blindly after detecting the opening chars.

lorenzo:
closes #884

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HieExtras.hs
@@ -151,6 +151,12 @@ mkExtCompl label =
     Nothing Nothing Nothing Nothing Nothing Nothing Nothing
     Nothing Nothing Nothing Nothing Nothing
 
+mkPragmaCompl :: T.Text -> T.Text -> J.CompletionItem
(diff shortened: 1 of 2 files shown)


lorenzo:
I suggest reading the diff with whitespace ignore, since I re-formatted a few things that were getting out of hand https://github.com/haskell/haskell-ide-engine/pull/895/files?utf8=✓&diff=unified&w=1

PR #900, opened 2018-10-23
Remove liquid Haskell result files

meck:
These files get rewritten when the tests are run and dirties the repo.


meck:
It seems some tests for liquid Haskell depended on these files by reordering the tests this ensures they are present. Also added a check for the executable

jhrcek:
I'm not sure if it's ok that these files were removed.
After I finally managed to install liquidhaskell v 0.8.2.4 I'm getting these failures with master:
```
      test/unit/LiquidSpec.hs:38: 
      1) Liquid, liquid haskell plugin, runs the liquid haskell exe
           predicate failed on: ""
    
      test/unit/LiquidSpec.hs:52: 
      2) Liquid, liquid haskell plugin, reads errors from json file
           uncaught exception: IOException of type NoSuchThing (/home/jhrcek/Devel/github.com/jhrcek/haskell-ide-engine/test/testdata/liquid/.liquid/Evens.hs.json: openBinaryFile: does not exist (No such file or directory))
    
      test/unit/LiquidSpec.hs:91: 
      3) Liquid, liquid haskell plugin, reads types from vim.annot file
           uncaught exception: IOException of type NoSuchThing (/home/jhrcek/Devel/github.com/jhrcek/haskell-ide-engine/test/testdata/liquid/.liquid/Evens.hs.vim.annot: openFile: does not exist (No such file or directory))
    
      test/unit/LiquidSpec.hs:105: 
      4) Liquid, liquid haskell plugin, reads types from vim.annot file 2
           uncaught exception: IOException of type UserError (user error (Pattern match failure in do expression at test/unit/LiquidSpec.hs:108:7-13))
```

I think this should be reverted (except maybe for the `findExecutable` test, which is useful)

meck:
Test 1 indicates ```msg``` [here](https://github.com/haskell/haskell-ide-engine/blob/6d12938142278143da8d630a8c355ca5a0b47f25/test/unit/LiquidSpec.hs#L38) is empty, that should be the output of the liquid command.

What is ```which liquid``` ?

if I run ```liquid test/testdata/liquid/Evens.hs --json``` from the project dir manually I get the expected output:
```
RESULT
[{"start":{"line":9,"column":1},"stop":{"line":9,"column":8},"message":"Error: Liquid Type Mismatch\n  Inferred type\n    VV : {v : Int | v == (7 : int)}\n \n  not a subtype of Required type\n    VV : {VV : Int | VV mod 2 == 0}\n \n  In Context"}]
```

Test 3-4 fails since they depend on the products produced by test 1

jhrcek:
@meck sorry for confusion. It turns out that even though I had liquid was installed and present on path, I was missing the z3 SMT solver package. After installing it, all the tests are passing! Thank you

Issue #907, opened 2018-10-26
Add ghc-mod case split as a refactor code action


bubba:
Currently VSCode is the only client that supports ghc-mod case splitting since it relies on the vscose plugin to expose the command. We should also add it to the refactoring code actions (alongside HaRe stuff) so that other clients can get access to it too.

Avi-D-coder:
I'll give this a shot.

PR #889, opened 2018-10-20
Quickfix action to prefix unused terms with '_'

meck:
This resolves #843

PR #917, opened 2018-10-30
Make casesplit a HaRe action

Avi-D-coder:
Move `splitCaseCmd` from the GhcMod plugin to the HaRe plugin in order to avoid a circular dependency. 
Move the `casesplit`  tests to HaRe tests file. 
Add a `casesplit` CodeAction test.
I have created an associated [pull request](https://github.com/alanz/vscode-hie-server/pull/121) to avoid breaking vscode-hie-server use of `ghcmod:casesplit`.

closes: https://github.com/haskell/haskell-ide-engine/issues/907

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HaRe.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HaRe.hs
@@ -7,6 +7,8 @@
 module Haskell.Ide.Engine.Plugin.HaRe where
 
 import           Control.Lens.Operators
+import           Control.Lens.Setter ((%~))
+import           Control.Lens.Traversal (traverseOf)
 import           Control.Monad.State
 import           Control.Monad.Trans.Control
 import           Data.Aeson
@@ -22,12 +24,14 @@ import qualified Data.Text.IO                                 as T
 import           Exception
 import           GHC.Generics                                 (Generic)
 import qualified GhcMod.Error                                 as GM
+import qualified GhcMod.Exe.CaseSplit                         as GM
 import qualified GhcMod.Monad                                 as GM
 import qualified GhcMod.Utils                                 as GM
 import           Haskell.Ide.Engine.ArtifactMap
 import           Haskell.Ide.Engine.MonadFunctions
 import           Haskell.Ide.Engine.MonadTypes
 import           Haskell.Ide.Engine.PluginUtils
+import           Haskell.Ide.Engine.Plugin.GhcMod             (runGhcModCommand)
 import qualified Haskell.Ide.Engine.Plugin.HieExtras          as Hie
 import           Language.Haskell.GHC.ExactPrint.Print
 import qualified Language.Haskell.LSP.Core                    as Core
@@ -64,6 +68,9 @@ hareDescriptor plId = PluginDescriptor
           deleteDefCmd
       , PluginCommand "genapplicative" "Generalise a monadic function to use applicative"
           genApplicativeCommand
+
+      , PluginCommand "casesplit" "Generate a pattern match for a binding under (LINE,COL)"
+          splitCaseCmd
       ]
   , pluginCodeActionProvider = Just codeActionProvider
   , pluginDiagnosticProvider = Nothing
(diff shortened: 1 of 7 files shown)


Avi-D-coder:
What do you think about refactoring all the mkCodeActions in HaRe.hs?
see: Avi-D-coder/haskell-ide-engine@bd01ab3021789847f5022937f017c2fb689aeb75

alanz:
I think the intention behind this ticket is to make the case split command generally available, by adding a case to the HaRe Plugin code action generator to offer a command to split the case, which activates this command.

So moving the mechanics of it to the HaRe Plugin is half the job, the other half is to modify [codeActionProvider](https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/Plugin/HaRe.hs#L294) to generate the command.

Doing that would mean it is available in all HIE clients, and can be *removed* from the vscode client, as a special case.

src/Haskell/Ide/Engine/Plugin/HaRe.hs
```diff
@@ -339,3 +410,9 @@ codeActionProvider pId docId _ _ (J.Range pos _) _ =
           title = "Duplicate definition of " <> name
       dupCmd <- mkLspCommand pId "dupdef" title (Just args)
       return $ J.CodeAction title (Just J.CodeActionRefactor) mempty Nothing (Just dupCmd)
+
+    mkCaseSplitAction name = do
+      let args = [J.toJSON $ HP (docId ^. J.uri) pos]
+          title = "Case split on " <> name
+      splCmd <- mkLspCommand pId "casesplit" title (Just args)
+      return $ J.CodeAction title (Just J.CodeActionRefactorRewrite) mempty Nothing (Just splCmd)
```
Avi-D-coder: @alanz This generates the command. It works on neovim.

src/Haskell/Ide/Engine/Plugin/HaRe.hs
```diff
@@ -339,3 +410,9 @@ codeActionProvider pId docId _ _ (J.Range pos _) _ =
           title = "Duplicate definition of " <> name
       dupCmd <- mkLspCommand pId "dupdef" title (Just args)
       return $ J.CodeAction title (Just J.CodeActionRefactor) mempty Nothing (Just dupCmd)
+
+    mkCaseSplitAction name = do
```
Avi-D-coder: @alanz This generates the command. It works on neovim.

src/Haskell/Ide/Engine/Plugin/HaRe.hs
```diff
@@ -339,3 +410,9 @@ codeActionProvider pId docId _ _ (J.Range pos _) _ =
...
+      let args = [J.toJSON $ HP (docId ^. J.uri) pos]
+          title = "Case split on " <> name
```
Avi-D-coder: @alanz This generates the command. It works on neovim.

alanz:
Oops, sorry. I saw the matching change on vscode, and assumed it was just a switch over.

I think to ease the transition we should keep it exposed in ghc-mod for the time being.  This would involve re-instating the command there, and getting it to call the function from HaRe. Which hopefully does not cause an import loop.

Avi-D-coder:
I'll give moving splitCaseCmd and HarePoint to a new module a try tomorrow. I think that will avoid circular dependencies. 

It's easy to miss in all the duplicate mkCodeAction definitions. That's why I'm advocating for the linked refactor.

alanz:
The problem with the linked refactor is that not everyone updates their hie installation all that often, and vscode pushed updates pretty much immediately.

So we need to support the old and the new for a while, else things will break.

alanz:
In fact, the simplest might be to leave the command in the ghc-mod plugin, but just add the code action generation to it.

Then we can make an issue to migrate it to the HaRe module at a future date, when we have removed the explicit code from the vscode plugin.

lorenzo:
Can we have the ghc-mod action trigger the now migrated HaRe action via the client? I think that would be the cleanest migration path

src/Haskell/Ide/Engine/Plugin/HaRe.hs
```diff
@@ -294,20 +359,26 @@ hoist f a =
...
             , mkDuplicateAction name
             ]
-        else return (IdeResultOk [])
+        _   -> case getArtifactsAtPos pos (locMap info) of
+               [h] -> do
+                let name = Hie.showName $ snd h
+                debugm $ show name
+                IdeResultOk <$> sequence [
+                  mkCaseSplitAction name
+                  ]
```
bubba: The command generation stuff LGTM

Avi-D-coder:
@lorenzo  I'm not clear how we could call the HaRa command with a GhcMod command. What I have done is move anything that would case circular dependence to HieExtras.

Avi-D-coder:
@alanz I rebased the refactor on the new changes that don't break compatibility. What are your thoughts on: Avi-D-coder/haskell-ide-engine@9f910841646a75f9ef8bf8cbf66b274be2159b20

alanz:
@Avi-D-coder Its looking good, and the refactoring of the HaRe command construction makes it a lot clearer, in my opinion.
I suggest you either add it to this PR or make a new PR with it.

lorenzo:
@Avi-D-coder I was meaning something like this https://github.com/haskell/haskell-ide-engine/pull/889/files#diff-df7cc8b4c1792382d5165b401f9ba6d0R536

alanz:
So, playing with it, I see we test that the case split command happens, but not that a code action is generated.

It should probably be tested together with the other code actions in https://github.com/haskell/haskell-ide-engine/blob/master/test/functional/FunctionalCodeActionsSpec.hs#L226

Issue #903, opened 2018-10-24
DuplicateRecordFields leads to incorrect completions 


Avi-D-coder:
This bug is known to occur with vscode and LanguageClient-neovim.

#### Steps to Reproduce
```hs
{-# LANGUAGE DuplicateRecordFields #-}

data Foo = FooBar
  { bar :: Int
  }
```
Start typing `bar`
completion items: 
` $sel:bar:FooBar Function [+] [LC] Foo -> Int^@Main` 
`$tc'FooBar      Function [+] [LC] TyCon^@Main`

lorenzo:
I’m not sure I understand why this issue is related to DuplicateRecordFields. Can you elaborate a bit more?

Avi-D-coder:
`$sel:bar:FooBar Function [+] [LC] Foo -> Int^@Main` is suggested instead of `bar     Function [+] [LC] Foo -> Int Main` only when `DuplicateRecordFields` is active.

` $tc'FooBar Function [+] [LC] TyCon^@Main` seems to be suggested regardless of whether `DuplicateRecordFields` is used.

alanz:
I think we need to filter the suggestions, we are definitely returning junk for these.

alanz:
Trivial getting-started task: change completions like `$sel:bar:FooBar` into `bar`.

jhrcek:
@alanz I'd like to tackle this. I'd like to clarify general expectation about the fix:
Should the fix strip the [$sel and constructor name](https://ghc.haskell.org/trac/ghc/wiki/Records/OverloadedRecordFields/DuplicateRecordFields#Implementation) name "blindly" from every completion, or should it check that DuplicateRecordFields extension is enabled (I'm not sure if that's even possible yet) before doing that?

alanz:
I would strip it blindly, it is never a valid completion.

On vscode, go to File, Preferences, Settings. Then under Extensions find Haskell Language Server, and set "Trace:Server" to "Messages".

jhrcek:
After spending some time trying to figure out the best place to do the stripping I came up with this.
https://github.com/jhrcek/haskell-ide-engine/commit/5cbfbdea9471492b264ceb70487ee54168ee0b40

alanz:
LGTM, turn it into a PR

Avi-D-coder:
Just ran into `$f` prefix for derived instances.

alanz:
[...]

PR #918, opened 2018-10-31
Strip compiler generated stuff from completions

jhrcek:
Fix for https://github.com/haskell/haskell-ide-engine/issues/903

While working on this I noticed that there's similar issue when you hover over record selectors with DuplicateRecordFields enabled. I'd treat that in separate PR

onhover


alanz:
LGTM, happy for the other to be a separate PR.  And we can perhaps make one for the `$f` prefix too, once we have an appropriate test case for it.

I wonder if we should just strip anything of the form `$[a-z]+:`.

alanz:
And it looks like it is going to need a fixup, based on the PR that just landed.

jhrcek:
Rebased onto master.

`$[a-z]+:.`  doesn't seem to cover the `$f` and `$fx` thing.
But I agree this should be generalized. I'll grep through the GHC codebase to see how to best generalize `stripAutoGenerated` so it can also be reused for on hover functionality.

PR #930, opened 2018-11-09
[WIP] Support cabal new-install hie

alanz:
This branch passes all the stack tests, and can be used without stack by doing

```bash
cabal new-install hie
```
It does not yet run the tests properly, as they rely on stack.

This opens the road to a GHC 8.6.x version, which currently compiles with cabal but not stack. In my ghc-8.6 branch.


alanz:
Superseded by #962

PR #931, opened 2018-11-09
Add 8.4.4 to the Windows-friendly make script

fsoikin:


lorenzo:
Thanks!

I'll merge this right away since it looks good to me and there is no need for waiting for the CI to pass on this change.

fsoikin:
Thanks for merging @lorenzo 

I also wanted to ask about this: I'm very interested in this whole thing working on Windows, and I'd be willing to put some hours towards it.

In particular, this `.bat` tingles my spidey sense, it could be written in PowerShell (which is the current Windows shell of choice) much more elegantly and DRY. Would you be interested in that?

In addition, the main readme offers a way to build on Windows by installing `make` and running `make build-all`, but that doesn't really work, since Windows executables have the `.exe` extension, but `Makefile` refers to them by the UNIX extension-less names. Is that something that was missed, or am I the one who is missing something?

lorenzo:
@fsoikin Definitely, that would be great to have!

> Is that something that was missed?

Most likely, since most of use are not currently using windows. It will be really helpful if you could help us fixing those oversights.

fsoikin:
Great, thank you @lorenzo !

PR #932, opened 2018-11-09
Replace bat script with PowerShell, update Windows instructions in README

fsoikin:


README.md
```diff
@@ -200,28 +200,19 @@ stack --stack-yaml=stack-8.0.2.yaml install
 
 ### Installation on Windows
 
-#### The `make` tool
+The `Makefile` doesn't work on Windows due to several UNIX-specific things, such
```
Anrock: Uhm, since when? I'm using makefile to build hie exlusively and everything works.

fsoikin: I don't see how this could work: https://github.com/haskell/haskell-ide-engine/blob/master/Makefile#L22

When I build, there is no file `$(STACKLOCALBINDIR)/hie`, instead there is a file named `$(STACKLOCALBINDIR)/hie.exe` - note the extension. 

Could it be that you're running everything under Cygwin or under WSL or in Docker?

alanz:
I am not sure what the next step should be around this, not being a part of the windows world.

lorenzo:
just merging it will be enough. This is already a nice improvement for windows users

alanz:
Ok, did not realise it was an add, rather than a change.

alanz:
And I notice it deleted the .bat file.  Perhaps we should restore it, as an alternative way of doing things?

README.md
Anrock: @fsoikin no, i'm usually running it under cmd.exe but have msys binaries in path. The only thing i had to do is install make with msys.

alanz:
ok, as I say, I am not familiar with the ecosystem.

README.md
fsoikin: @Anrock so then, how does `cp $(STACKLOCALBINDIR)/hie $(STACKLOCALBINDIR)/hie-8.2.2` work for you? I'm assuming `cp` must come from `msys` (though I'm not sure it does), but what about the non-existent file `$(STACKLOCALBINDIR)/hie`?

fsoikin:
@lorenzo @alanz it is possible that the PowerShell script won't work on a very old version of Windows (pre-2000 or so), but I'm pretty sure that `stack` won't work there either :-)

@alanz also, here's prior conversation: https://github.com/haskell/haskell-ide-engine/pull/931#issuecomment-437465568

PR #935, opened 2018-11-11
Apply hlint suggestions and minor format fixes 

Avi-D-coder:


src/Haskell/Ide/Engine/Plugin/HieExtras.hs
```diff
@@ -82,8 +82,8 @@ getDynFlags = ms_hspp_opts . pm_mod_summary . tm_parsed_module
 
 -- ---------------------------------------------------------------------
 
-data NameMapData = NMD
-  { inverseNameMap ::  !(Map.Map Name [SrcSpan])
+newtype NameMapData = NMD
+  { inverseNameMap :: Map.Map Name [SrcSpan]
```
alanz: You have removed a strictness annotation here. This is probably not intended.

alanz: See also https://github.com/ndmitchell/hlint/issues/548

Avi-D-coder: My bad. I changed it back.

Avi-D-coder: For the time being I added annotations to ignore this use of `data` and the compatibility lambdas.

PR #948, opened 2018-11-18
Adding hie-nix instructions to readme.

benkolera:
I made a spacemacs layer and nix expression for those interested in spacemacs+hie+nix. It is the simplest thing that works rather than layering it on the haskell layer while I don't fully grok all the layers and interactions yet. This links it to the hie readme at @alanz 's request on irc.

May morph into a layer that can handle nix and non-nix hie project, get layered on the haskell layer or get merged into the official layer one day. Don't know which option is best yet, but it is a start! :slightly_smiling_face:


alanz:
Thanks

PR #962, opened 2018-11-28
Support for GHC 8.6 and preliminary new-build

alanz:
- Supports GHC 8.6.1 and 8.6.2
- Preliminary support for cabal new-build
  - can be installed via `cabal new-install hie`
  - can be used against projects built with cabal new-*

The support for use against new projects is preliminary, and likely buggy. At the very least we know it does not work for projects having an exe only.

But there is a new approach coming for `cabal-helper` and `ghc-mod`, so it is pointless spending too much time patching up an interim solution.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
+++ b/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE OverloadedStrings   #-}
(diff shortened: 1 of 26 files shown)


alanz:
Looks like the appveyor test failed for a network outage reaching stackage. 

Can someone restart that build?

lorenzo:
@alanz Only owners of the appveyor account can do it. If you can get a hold of them, try asking to include you as an account admin as well.

alanz:
It might be quickest to just kick the ci with a commit

src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -607,38 +608,102 @@ symbolProvider uri = pluginGetFile "ghc-mod symbolProvider: " uri $
...
+        goValD (L _ (VarBind _ _ _ _))                 = error "goValD"
+        goValD (L _ (AbsBinds _ _ _ _ _ _ _))          = error "goValD"
+        goValD (L _ (PatSynBind _ _))                  = error "goValD"
+        goValD (L _ (XHsBindsLR _))                    = error "goValD"
+#elif __GLASGOW_HASKELL__ >= 804
+        goValD (L _ (VarBind _ _ _))        = error "goValD"
+        goValD (L _ (AbsBinds _ _ _ _ _ _)) = error "goValD"
+        goValD (L _ (PatSynBind _))         = error "goValD"
+#else
+        goValD (L _ (VarBind _ _ _))           = error "goValD"
```
bubba: Should these errors return `[]`?

alanz: These are things that should never occur in the AST for `GhcPs`.  So  erroring out lets us know something is wrong, and should show up at dev time.

If anyone feels strongly about it ,I can make it `[]`, but that tells the next person who works on the code that these might be legitimate values.

bubba: Ok that makes sense that they shouldn’t exist in the parse stage, I would rather have it crash in that case!

PR #981, opened 2018-12-06
Use travis to generate MacOS binary releases

lorenzo:
The use of travis stages is so that we can stay in the allowed time limits for travis. Removed the testing part out of travis, since we have plenty tests already, but it can be added back with no problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ addons:
     packages:
     - z3
     - stack
-    - cabal-install
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,37 +20,103 @@ cache:
...
+       env: GHC_VER="8.4.2"
+       script: *test
+
+     - stage: test
+       env: GHC_VER="8.2.2"
+       script: *test
+
+     - stage: test
+       env: GHC_VER="8.2.1"
+       script: *test
```
bubba: Should we also add 8.6.1/8.6.2

lorenzo: Yeah, but I was waiting for your recent PR to be merged first.

alanz:
When you guys are happy, merge.

I presume we can get linux binaries out of this as well?

alanz:
And in terms of S3 or other, if we ask the haskell committee they may be able to help

lorenzo:
I think this is good to merge already. Maybe circleci is good for Linux deployments?

lorenzo:
yeah, I’ll start experimenting with the windows builds next. Likely in my own fork to not disturb circleci and appveyor

lorenzo:
Ok, merging this. I started to work on windows builds

PR #991, opened 2018-12-11
Add Shakefile as possible build script

### LINKED ISSUE #650: Shake instead of Make?
I understand installation of the various versions of hie via a Makefile is seen as being linux-centric, and causes problems on e.g. windows.

Could we avert this by using something like Shake?

And if so, is someone prepared to do it?

fendor:
Closes #650 
I implemented the Shakefile to be run directly from `stack`.

```
# one-stop-shop build everything
stack Shakefile.hs build-all
```

Features:

* `stack` is the only build requirement
* `shake` is used instead of `make`
* should work on all platforms (not fully tested yet)
* `dist` target works on windows out of the box, making it easier to create releases for windows
* submodules are synced / updated at the start -> works after clean clone of project

Quirks:
* `shake.yaml` is necessary to start the `stack Shakefile.hs ...` command before submodules are loaded.
* On first execution, a compiler is downloaded. However, no compiler is downloaded more than once.

Tested on: 
* [X] Fedora
* [X] Arch Linux
* [x] Windows
* [ ] MacOS
* [X] NixOS

Working on NixOS with the following commands:
```bash
nix-shell -p "haskellPackages.ghcWithPackages (pkgs: with pkgs; [zlib tar shake])" zlib
export PATH=$NIX_USER_PROFILE_DIR/profile/bin:$PATH && ghc -O2 ./Shakefile.hs && ./Shakefile build
```
However, the ghc versions 8.2.1 and 8.4.2 are not easily accessible and must be removed from the list of hieVersions in `Shakefile.hs` in order to make it work.

### DIFF
--- a/Shakefile.hs
+++ b/Shakefile.hs
@@ -1,5 +1,5 @@
 #!/usr/bin/env stack
--- stack --resolver lts-12.21 --install-ghc runghc --package shake
+-- stack --stack-yaml=shake.yaml  --install-ghc runghc --package shake
 
 import           Development.Shake
 import           Development.Shake.Command
@@ -40,7 +40,8 @@ main = shakeArgs shakeOptions { shakeFiles = "_build" } $ do
         forM_
             hies
             (\hie -> phony hie $ do
-                need ["submodules", "cabal"]
+                need ["submodules"]
+                need ["cabal"]
                 stackLocalDir <- getLocalBin
                 let versionNumber = drop 4 hie
                 buildHie stackLocalDir versionNumber
--- a/Shakefile.hs
+++ b/Shakefile.hs
@@ -81,9 +81,9 @@ buildHie localBinDir versionNumber = do
     $ execStackWithYaml versionNumber ["install", "happy"]
   execStackWithYaml versionNumber ["build"]
   execStackWithYaml versionNumber ["install"]
-  cmd_ "cp"
-       [localBinDir </> "hie" <.> exe]
-       [localBinDir </> "hie-" ++ versionNumber <.> exe]
+  copyFile'
+       (localBinDir </> "hie" <.> exe)
+       (localBinDir </> "hie-" ++ versionNumber <.> exe)
 
 buildCopyCompilerTool :: VersionNumber -> Action ()
 buildCopyCompilerTool versionNumber =
--- a/Shakefile.hs
+++ b/Shakefile.hs
@@ -42,6 +42,7 @@ main = do
   -- unset GHC_PACKAGE_PATH for cabal
   unsetEnv "GHC_PACKAGE_PATH"
   shakeArgs shakeOptions { shakeFiles = "_build" } $ do
+    want ["help"]
     phony "ghc" $ do
       ghc <- getGhcPath
       command_ [] ghc ["--version"]
--- a/Shakefile.hs
+++ b/Shakefile.hs
@@ -54,9 +54,15 @@ main = do
     phony "build"      (need (reverse $ map ("hie-" ++) hieVersions))
     phony "build-all"  (need ["build"] >> need ["build-docs"])
     phony "dist"       buildDist
-    phony "build-docs" (forM_ hieVersions buildDoc)
+    phony "build-docs" (need (reverse $ map ("build-doc-hie-" ++) hieVersions))
     phony "test"       (forM_ hieVersions test)
     phony "build-copy-compiler-tool" $ forM_ hieVersions buildCopyCompilerTool
+
+    forM_
+      hieVersions
+      (\version -> phony ("build-doc-hie-" ++ version) $ do
+        buildDoc version
+      )
 
     forM_
       hieVersions
@@ -170,7 +176,7 @@ helpMessage = do
   out ""
   out "Targets:"
   out
-    "    build                Builds hie for all supported GHC versions (8.2.1, 8.2.2, 8.4.2, 8.4.3, 8.4.4, 8.6.1 and 8.6.2)"
+    "    build                Builds hie for all supported GHC versions (8.2.1, 8.2.2, 8.4.2, 8.4.3, 8.4.4, 8.6.1, 8.6.2 and 8.6.3)"
   out
     "    build-all            Builds hie and hoogle databases for all supported GHC versions"
   out "    hie-8.2.1            Builds hie for GHC version 8.2.1 only"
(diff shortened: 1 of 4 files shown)


Anrock:
I've noticed there is a downside in requirement to download 8.6.2 to run shake script despite me only wanting to check help and build 8.4.4. Not sure if it's fixable.

lorenzo:
Yeah, having to download the compiler and dependencies to actually install something else is annoying.

Anrock:
I wonder if it's possible to get "whatever ghc is already available" from stack...

lorenzo:
Or at least the ghc I'm intending to use anyway

power-fungus:
@lorenzo @Anrock It is intended that the same resolver as in the global `stack.yaml` is used. Therefore, nothing extra is downloaded.
The initial download of the resolver and the ghc is necessary to eliminate all build exterior dependencies beside `stack`, which is a major benefit on non-linux platforms.

Anrock:
@power-fungus Can you clarify? 

Suppose i have clean stack install without any ghc downloaded yet. 

If i run `stack Shakefile ...` then stack uses resolver from `shake.yaml` with ghc-8.6.2 to run `Shakefile.hs` itself. So ghc-8.6.2 is downloaded in order to run `Shakefile` in addition to whatever ghc i need for building hie. 

If i run `make hie-8.4.4` it executes `stack --stack-yaml stack-8.4.4.yaml ...`, so stack downloads ghc-8.4.4 and nothing else.

Where does `stack.yaml` come in in case with Makefile?

fendor:
@Anrock `make hie-8.4.4` still requires the `cabal` target, which uses the local `stack.yaml` that requires `ghc-8.6.2` as well.

Anrock:
@fendor i see now! I thought it was just overlooked.

fendor:
I implemented the `dist` target in a platform independent way, e.g. `tar` package and `zlib` package.
I assume that we can use `zlib`, since git also requires that library and should be present on the system, anyways. 
However, I have to test that first. 

Tested on:
* [ ] Arch Linux
* [ ] Fedora
* [ ] Windows 
* [ ] MacOs

fendor:
I added the `icu-macos-fix` target. 
It would be nice if someone can validate that the `Shakefile.hs` also works on MacOs.

fendor:
The pull request should be ready for reviews now.

alanz:
I invoked it as `./ShakeFile.hs`, and got

```bash
-*- mode: compilation; default-directory: "~/mysrc/github/alanz/haskell-ide-engine/" -*-
Compilation started at Fri Dec 21 09:51:28

./Shakefile.hs 
Warning: No want/action statements, nothing to do
Build completed in 0.00s
```

I think it should have a default target, that prints out the available ones, as happens if you just type `make`.

Invoking it as `./ShakeFile.hs build-all` does kick off some activity though.

fendor:
I agree it should have a default, the help message seems like a good idea.

I have to check that. In my opinion, the last ghc should be 8.6.2, but I might be mistaken.

alanz:
The point is that the hie-8.2.1, hie-8.4.3 etc are all copies of the same
file, the last hie exe compiled

fendor:
If I understand you correctly, this behaviour should be similar to the Makefile.
```make
cp '$(STACKLOCALBINDIR)/hie' '$(STACKLOCALBINDIR)/hie-$*'
```
We could copy it directly from the compile path, just like when executing the `dist` target.

fendor:
@alanz The issue about the hie-versions not properly being copied should be resolved now.

alanz:
One thing though, current master builds for 8.6.3 too.

fendor:
I will update the pull request to also include 8.6.3. 
I would also update the `Readme` to use the `Shakefile`.

shake.yaml
```diff
@@ -0,0 +1,7 @@
+# Used to provide a different environment for the shake build script
+resolver: nightly-2018-12-01 # GHC 8.6.2
```
alanz: Is there any way we can do something like "use the system resolver on this machine"?

Shakefile.hs
```diff
@@ -170,7 +176,7 @@ helpMessage = do
   out ""
   out "Targets:"
   out
-    "    build                Builds hie for all supported GHC versions (8.2.1, 8.2.2, 8.4.2, 8.4.3, 8.4.4, 8.6.1 and 8.6.2)"
+    "    build                Builds hie for all supported GHC versions (8.2.1, 8.2.2, 8.4.2, 8.4.3, 8.4.4, 8.6.1, 8.6.2 and 8.6.3)"
```
alanz: We should eventually refactor this (and the next set of targets) to be generated from the list of `hieVersions`, so we can set that in one place only.

fendor: in the next version of the shakefile i would like to refactor this to obtain the list of hie-versions from present stack.yaml files in the directory. But i felt like this should be in another pull request.

shake.yaml
power-fungus: An extra `stack.yaml` is required since the other `yaml`s contain information about extra dependencies in the `submodules` folder, which is not present on a non-recursive clone or if the submodules change. Since the build script should take care of updating the git submodules, they are not allowed to be referenced within the `stack-yaml` that is used by `Shakefile.hs`.

It is possible, however, to include `system-ghc: true` within the `shake.yaml` ([see option](https://docs.haskellstack.org/en/stable/yaml_configuration/#system-ghc)) However, this seems to be incompatible with Docker and nix.

alanz:
Is this ready for merge yet?

fendor:
Technically yes, but the `README.md` needs updates that one should use `./Shakefile.hs`. This should be done tomorrow.
Also, it is not clear to me if we should use  `system-ghc: true` in `shake.yaml` now. I would like to test it tomorrow.
Moreover, should the `Makefile` and `BuilAll.ps1`  be deleted in this pull request to avoid confusion?

alanz:
I would say we should add it, and keep the existing Makefile and .ps1 file.

And add a section saying we now have experimental Shake build support, and to please try it and report.

Then we can see what kind of feedback we get.

fendor:
Ok, then I will revert the changes to the `README.md` and add a section for the `Shakefile.hs`.

fendor:
README is updated, but I did not manage to use the option `system-ghc: true` correctly.

I think, it is ready to merge now.

alanz:
Given #1008, I wonder if we should not use an LTS that is not for GHC 8.6.3.

Otherwise it looks good.  I think the build is failing because hackage was being restarted.

fendor:
Makes sense, LTS-12.25 looks fine?

Issue #997, opened 2018-12-15
Installation confusion / struggles


ianchanning:
My apologies if this comes across as negative, but I've been really struggling to install this. I've spent about 5 hours of which 3 hours was trying to understand what needs to be installed and 2 hours installing which eventually failed because it exhausted my disk space.

I'm running:

* Fedora 29
* Stack v1.9.3
* GHC v8.4.4

## 0. GHC Version

As a Haskell newbie, getting the GHC version is non-obvious, especially given how dependent the installation is on the GHC version adding this to the readme might help: `stack ghc -- --version`

I think this is typically one of the first things that a new Haskell developer will want to do as they want to get their tooling setup, so making instructions extra newbie friendly seems reasonable.

## 1. Fedora 

For fedora the equivalent libraries are non-obvious:

```
$ sudo dnf install libicu-devel ncurses-devel
```

## 2. Differing GHC versions

I think it might be simpler if you have entirely separate sections for different GHC versions rather than mixing the versions. 

There's references to:

v8.0.2, v8.6.2, 'all GHC versions', GHC 8.2.1 to 8.6.2.

If v8.0.2 doesn't work on master it would be simpler if there were just a link to the relevant branch and no further instructions.

## 3. VS Code instructions mixed in with install instructions

The installation seems to be biased towards VS Code with instructions mixed in specifically for that in the installation section.

> To install all supported GHC versions, name them as expected by the VS Code plugin, and also build a local hoogle database, you need the make tool (on Windows, see the further advice below). Use the command:
>
> ```
> make build-all
> ```

If you don't have VS Code installed this sentence means nothing and so you're left confused.

It's not clear if you need to run `make build-all` and `stack install` or just one of them if you're not using VS Code.

## 4. Time to `make`

Running `make build-all` takes 50+ minutes (and still going) on a Core i5, 8GB RAM, SSD laptop + 40MB/s internet connection.

## 5. The first instruction is to run `make build-all` which should almost never be required

This instruction appears to run `stack --stack-yaml=stack-<VERSION>.yaml install` for every single version. Given that it takes 40 minutes per version this becomes a slightly insane process. I've currently been waiting 2hr 20 mins and I don't want to kill the process because I don't want to leave it in a broken state.

## 6. The GHC version does not match the latest stack GHC version

I have the latest stack version (v1.9.3) which comes with GHC v8.4.4. Given that the primary installation mechanism for this is via stack, I would have thought that the default version of GHC should also be v8.4.4.

[...]

PR #998, opened 2018-12-16
Updated README installation instructions

ianchanning:
This are some modifications based on the confusion/problems I hit with #997. 

* Fedora library installation instructions 
* Added a step to check GHC version before installing
* Handle each GHC version independently 
* Put the step to install all GHC versions at the end and with a warning for the installation size and time taken

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,14 +234,6 @@ stack install cabal-install
 cabal update
 ```
 
-`make` doesn't work on Windows due to several UNIX-specific things, such
-as the `cp` command or extensionless executable names. Instead, a PowerShell
-script is provided specifically for this purpose:
-
-```
-git clone https://github.com/haskell/haskell-ide-engine --recursive
-cd haskell-ide-engine
-```
 #### Install *all* available GHC versions
 
 *Warning*: Requires 20+ GB of space and potentially more than 2 hours to install, so please be patient!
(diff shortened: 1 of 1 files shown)


alanz:
The preferred installation mechanism is actually via the makefile, as it makes sure the repo is synced, installs the required cabal libraries if missing, and makes copies of the executables with suffixes to be able to tell them apart.

So stack install will install a single version of hie, and the hie-wrapper exe will then be useless.

So, the split out/reorganise makes sense, but the "active" command should be something like

```bash
make hie-8.6.2
```

Also, running just `make` prints a list of the available targets, including the compiler versions.

ianchanning:
Ah ok, sorry. I just blindly followed the commands in the README! 

I'll have another try and re-submit. 

I think the GHC version check step is still useful?

alanz:
I think clearly documenting how to install for just the versions of GHC you care about is a good thing. And that is basically what the version check does.

ianchanning:
I've replaced all `stack install` steps with `make` plus with the addition of the hoogle docs for an individual version. I followed these steps as well to install both the v8.6.2 and v8.4.4 versions and tested them with a yesod project using Vim and Neovim.

alanz:
looks good, thanks for all the effort in doing this with verification.

PR #1013, opened 2018-12-29
build/makefile improvements

2mol:
So this is a first baby step to make the build process do a little less unnecessary work. For now it only makes sure that for the single build process everything is compiled with the same resolver.

tl;dr: it makes it download one less stack resolver and build about a 30% faster.

This is not complete, and likely makes the building of all 8 versions slightly worse, since cabal might be builtg 8 times. (AFAIK otherwise `make` is smart enough to build cabal only once, even if we loop over `hie-%` 8 times.)

## Test

I ran the builds on a macbook air i5.

### Prep

Clean up everything:

`rm -rf .stack-work ~/.stack ~/.local/bin/hie* ~/.local/bin/happy ~/.local/bin/cabal`

### Single HIE build

Tested with 8.4.4, including docs.

`make hie-8.4.4 && make build-doc-8.4.4 && date`

| Makefile change            | compilation time | `~/.stack` size    | `~/.local/bin` size              |
| -------------------------- | ---------------- | ------------------ | -------------------------------- |
| unchanged                  | 1h               | 5.3G (2 resolvers) | 334M (3 x `hie` + cabal & happy) |
| only use one resolver      | 40min            | 3.5G (1 resolver)  | 334M                             |
| without building cabal     |                  |                    |                                  |
| linking instead of copying |                  |                    |                                  |

### Build & install all HIE (currently 8 versions)

TODO


2mol:
Tested on Linux, and there I still have the very weird problem that two resolvers are being downloaded.

I can't figure out right now where the command comes from that triggers the downloading of the nightly resolver.

```
...
Copied executables to /home/bla/.local/bin:
- hie
- hie-wrapper
Downloading nightly-2018-12-27 build plan ...
Downloaded nightly-2018-12-27 build plan.
Preparing to install GHC (tinfo6) to an isolated location.
This will not interfere with any system-level installation.
Preparing to download ghc-tinfo6-8.6.3 ...
ghc-tinfo6-8.6.3: download has begun
ghc-tinfo6-8.6.3:    7.49 MiB / 173.91 MiB (  4.30%) downloaded...
ghc-tinfo6-8.6.3:   35.69 MiB / 173.91 MiB ( 20.52%) downloaded...
ghc-tinfo6-8.6.3:   64.36 MiB / 173.91 MiB ( 37.01%) downloaded...
ghc-tinfo6-8.6.3:   92.71 MiB / 173.91 MiB ( 53.31%) downloaded...
ghc-tinfo6-8.6.3:  121.64 MiB / 173.91 MiB ( 69.95%) downloaded...
ghc-tinfo6-8.6.3:  150.83 MiB / 173.91 MiB ( 86.73%) downloaded...
ghc-tinfo6-8.6.3:  173.91 MiB / 173.91 MiB (100.00%) downloaded...
Downloaded ghc-tinfo6-8.6.3.
Unpacking GHC into /home/bla/.stack/programs/x86_64-linux/ghc-tinfo6-8.6.3.temp/ ...
Configuring GHC ...
Installing GHC ...
Installed GHC.
stack --stack-yaml=stack-8.4.4.yaml exec hoogle generate
Starting generate
Reading Cabal... 6.25s
...
```

alanz:
That looks like something is running against the default `stack.yaml`, which now uses a recent nightly.

So look for a stack command without a `--stack-yaml=` in it.

Or, rename `stack.yaml` and see what breaks. Except it may just use the default (global) project.

Makefile
```diff
@@ -25,9 +21,9 @@ ghc:
 # ------------------------------------------------------
 
 ## Builds hie for GHC version % only
-hie-%: submodules cabal
+hie-%: STACKLOCALBINDIR=$(shell stack --stack-yaml=stack-$*.yaml path --local-bin)
+hie-%: submodules cabal-%
 	stack --stack-yaml=stack-$*.yaml install happy
-	stack --stack-yaml=stack-$*.yaml build
```
alanz: We do the build before install because it should be a no-op, but all 'stack install' checks is that the artifact exists, so will not rebuild changes if this is an updated workspace.

alanz:
LGTM

To me it makes sense to merge this, since we are still deciding on using the `install.hs` script only.

2mol:
let me check first that it doesn't make the `build-all` case significantly worse :)

fendor:
I think, this PR can be closed since there is no makefile anymore.

PR #1021, opened 2019-01-01
Document reactorPidCache

bubba:


src/Haskell/Ide/Engine/LSP/Reactor.hs
```diff
@@ -31,6 +31,7 @@ import qualified Language.Haskell.LSP.Types    as J
 data REnv = REnv
   { scheduler         :: Scheduler.Scheduler R
   , lspFuncs          :: Core.LspFuncs Config
+  -- | The process ID of HIE. See 'HasPidCache'
   , reactorPidCache   :: Int -- TODO:AZ: do we need this? what is it for?
```
alanz: Should probably remove my TODO comment too :)

PR #1024, opened 2019-01-05
Use cabal to detect ghc version if cabal.project exists

### LINKED ISSUE #1023: Respect GHC version specified in cabal.project
In my project I have the following `cabal.project` file:

```
with-compiler: ghc-8.6.3
packages: .

```

However, as far as I can tell, HIE doesn't use GHC 8.6.3 and tries to use the "default" one (which is 8.4.4 in my case):

```
readCreateProcess: /Users/araga/src/github/hie/.stack-work/install/x86_64-osx/lts-12.26/8.4.4/bin/cabal-helper-wrapper "--with-ghc=ghc" "--with-ghc-pkg=ghc-pkg" "--with-cabal=cabal" "v1-style" "/Users/araga/src/haskell-works/hw-xml" "/Users/araga/src/haskell-works/hw-xml/dist-newstyle/build/x86_64-osx/ghc-8.4.4/hw-xml-0.1.0.4" "package-db-stack" "flags" "compiler-version" "ghc-merged-pkg-options" "config-flags" "non-default-config-flags" "ghc-src-options" "ghc-pkg-options" "ghc-lang-options" "ghc-options" "source-dirs" "entrypoints" "needs-build-output" (exit 1): failed
```

Executing this command manually I see that it tries to load a config file for GHC 8.4.4:

```
cabal-helper-wrapper: /Users/araga/src/haskell-works/hw-xml/dist-newstyle/build/x86_64-osx/ghc-8.4.4/hw-xml-0.1.0.4/setup-config: openFile: does not exist (No such file or directory)
```

Expected behaviour: use GHC that is specified in `cabal.project`.

lukel97:
Closes #1023

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/Base.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Base.hs
@@ -107,7 +107,7 @@ hieGhcDisplayVersion = compilerName ++ "-" ++ VERSION_ghc
 getProjectGhcVersion :: IO String
 getProjectGhcVersion =
   onException
-    (asum [tryStackProject, tryCabalProject, tryGhcVersion])
+    (asum [tryStackProject, tryCabalProject, trySystemGhc])
     (error "Couldn't find GHC version through stack, cabal, nor system GHC.")
 
   where
@@ -125,7 +125,7 @@ getProjectGhcVersion =
       L.infoM "hie" "Trying cabal GHC version"
       tryCommand "cabal v2-exec ghc -- --numeric-version"
 
-    tryGhcVersion = do
+    trySystemGhc = do
       isGhcInstalled   <- isJust <$> findExecutable "ghc"
       guard isGhcInstalled
       L.infoM "hie" "Trying system GHC version"


src/Haskell/Ide/Engine/Plugin/Base.hs
```diff
@@ -104,26 +105,31 @@ hieGhcDisplayVersion :: String
...
+      guard $ isStackProject && isStackInstalled
+      L.infoM "hie" "Trying stack GHC version"
+      tryCommand "stack ghc -- --numeric-version"
+
+    tryCabalProject = do
+      isCabalProject <- doesFileExist "cabal.project"
+      isCabalInstalled <- isJust <$> findExecutable "cabal"
+      guard $ isCabalProject && isCabalInstalled
+      L.infoM "hie" "Trying cabal GHC version"
+      tryCommand "cabal v2-exec ghc -- --numeric-version"
```
alanz: For older installed `cabal-install` versions `v2-exec` won't be understood.

I wonder if we should make a `runQuietCabal` which does `cabal --version`, and depending on the result uses the `vX-` prefixes or not.

alanz:
It looks good to me.

Heads up, I am considering delegating this functionality to `ghc-mod-core`, as it pretty much duplicates the `findCradle` functionality, and we may as well put it all in one place.

alanz:
E.g. See https://github.com/alanz/ghc-mod/blob/4ec1283866546a4c752bcbbf1e502d04f8817f3e/core/GhcMod/Cradle.hs#L51

```haskell
findCradle' :: (GmLog m, IOish m, GmOut m) => Programs -> FilePath -> m Cradle
findCradle' Programs { stackProgram, cabalProgram } dir = run $
    msum [ stackCradle stackProgram dir
         , cabalCradle cabalProgram dir
         , sandboxCradle dir
         , plainCradle dir
         ]
 where run a = fillTempDir =<< (fromJustNote "findCradle'" <$> runMaybeT a)
```

AlexeyRaga:
One problem to think about that at least I have:
```
$ cabal --version                                                                                                              ⤬
cabal-install version 2.4.1.0
compiled using version 2.4.1.0 of the Cabal library
```

But now:
```
$ cabal new-clean
$ cabal new-exec ghc -- --numeric-version
Resolving dependencies...
/Users/araga/src/github/avro/dist-newstyle/tmp/environment.-64999: createDirectory: does not exist (No such file or directory)
```

So until I do `new-build` it doesn't really work.

What I see is: if there is `cabal.project` then we can try parsing GHC version from `with-compiler: ghc-8.6.3` or fallback to `ghc --numeric-version` if the file doesn't exist or if it doesn't contain this line?

AlexeyRaga:
There is yet another problem with using `cabal new-exec ghc -- --numeric-version`:
```
$ cabal new-update
Downloading the latest package list from hackage.haskell.org
To revert to previous state run:
    cabal new-update 'hackage.haskell.org,2019-01-04T22:46:17Z'
$ XXX=$(cabal new-exec ghc -- --numeric-version)
$ echo $XXX
Resolving dependencies...
8.4.4
```

Note that after `cabal new-update` the result of this command contains `Resolving dependencies...` line. Which, I think, is wrong, but it is what it is and we can's use it directly...

src/Haskell/Ide/Engine/Plugin/Base.hs
AlexeyRaga: We can use `new-*` for now. Because applying `v2-` instead of `new-` should be done wider (wherever we do `build`, `configure`, etc.

PR #1031, opened 2019-01-07
CI improvements

bubba:
- Add a hspec formatter that produces an xml file that CircleCI can use to produce a [test summary](https://circleci.com/gh/bubba/haskell-ide-engine/2290) Gives a slightly nicer way to look at test results instead of reading the terminal output
- Fix cabal file change detection
  - The aggregate `all-cabal.txt` file changed every build so every job was redundantly saving the cache at the end
  - Changed it to ignore submodules/test .cabal files (I believe when it was first added there were no submodules in HIE so depending on submodule .cabal files was probably unintended) and sort them before `cat`ing so `all-cabal.txt` file is now deterministic
  - Shaves 6 minutes off of each job
- Cache Hoogle DB at `~/.hoogle`
  - Only generate it if `~/.hoogle` doesn't exist
  - Shaves 2 minutes off of each job
- Add documentation about CircleCI in docs folder

[A complete build on CircleCI now takes about ~20 minutes!](https://circleci.com/workflow-run/edf00754-c1dd-455d-9c82-e84828418209)

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -7,12 +7,3 @@ packages:
          ./submodules/cabal-helper/
          ./submodules/ghc-mod/
          ./submodules/ghc-mod/core/
-
-
--- allow-newer:ekg-core-0.1.1.4:base
--- allow-newer:ekg-json-0.1.0.6:base
-
-source-repository-package
-  type:     git
-  location: https://github.com/LeastAuthority/hspec-jenkins
-  tag:      2899766bbd36216b81b1719423a5c2443458ae06
--- a/cabal.project
+++ b/cabal.project
@@ -11,3 +11,8 @@ packages:
 
 -- allow-newer:ekg-core-0.1.1.4:base
 -- allow-newer:ekg-json-0.1.0.6:base
+
+source-repository-package
+  type:     git
+  location: https://github.com/LeastAuthority/hspec-jenkins
+  tag:      2899766bbd36216b81b1719423a5c2443458ae06
--- a/stack-8.2.1.yaml
+++ b/stack-8.2.1.yaml
@@ -27,8 +27,6 @@ extra-deps:
 - sorted-list-0.2.1.0
 - syz-0.2.0.0
 - yaml-0.8.32
-- github: LeastAuthority/hspec-jenkins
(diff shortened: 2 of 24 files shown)


stack-8.2.1.yaml
```diff
@@ -27,6 +27,8 @@ extra-deps:
 - sorted-list-0.2.1.0
 - syz-0.2.0.0
 - yaml-0.8.32
+- github: LeastAuthority/hspec-jenkins
+  commit: e5ad80f176779c99fc6a8c044afc2f89ceb43103
```
alanz: This should be a submodule, so we don't have to tweak every single config file for it.  That is why we do the others that way.

alanz:
This looks great, I have been wanting to have the junit output for a long time.

Sort out the submodule and it can merge.

bubba:
The tags differ for 8.2.x though since the master branch has some incompatible changes with whatever version hspec is pinned to. Is there a way to go about this with submodules too? 

Alternatively we could just get rid of the package and incorporate the formatter into the testutils module since it’s very small.

test/utils/TestUtils.hs
```diff
@@ -202,4 +201,88 @@ stackFileContents resolver = unlines
   , "flags: {}"
   , "extra-package-dbs: []"
   ]
+
```
alanz: If the block below originated elsewhere, provide details/credit.

cabal.project
```diff
@@ -11,3 +11,8 @@ packages:
 
 -- allow-newer:ekg-core-0.1.1.4:base
 -- allow-newer:ekg-json-0.1.0.6:base
+
+source-repository-package
+  type:     git
+  location: https://github.com/LeastAuthority/hspec-jenkins
+  tag:      2899766bbd36216b81b1719423a5c2443458ae06
```
alanz: I presume this is no longer needed?

test/utils/TestUtils.hs
bubba: I had added an attribution to the fork at `xmlFormatter`, is this enough?
https://github.com/haskell/haskell-ide-engine/blob/fb686fa6d5f288ede6235bc31090b81302484fd1/test/utils/TestUtils.hs#L223-L227

PR #1039, opened 2019-01-09
Remove EKG

bubba:
With increasing build times it would be wise to start trimming some dependencies.
Speak now or forever hold your peace!


Anrock:
Not sure why review from me was requested, as i'm not quite familiar with hie internals yet and never used EKG. But i'm always for removing things!

@bubba resolve conflicts, please.

Issue #1033, opened 2019-01-07
Remove Makefile & build-all.ps1 and leave only Shakefile


Anrock:
Hey, three build scripts are too much.

Right now I'm making a PR that removes both Makefile and build-all.ps1 and updates README about that (also some DRY cleanups).

Is there any objections or specific blockers to using Shake script as the only build script?

Todo:
- [x] Add 8.4.2 and 8.2.1 versions to `hieVersions` list in Shakefile #1034
- [x] Remove v1 prefix from cabal commands (#1043)
- [x] Rename Shakefile to install.hs and update README (#1044)
- [x] Remove mentions of Make/build-all.ps1 from README (~~#1120~~)
- [x] Actually remove Makefile and build-all.ps1 scripts. (#1140)
- [x] Make sure that Shakefile implements everything as Makefile does (~~#1032~~, ???) 
- [ ] Add some CI tests (separate PR by @bubba)

Anrock:
@alanz @bubba @fendor

fendor:
Following Blockers: Shakefile has the issue that not all supported ghc's are being installed. The list of ghc's is not complete.

Anrock:
@fendor it's seems 8.4.2 and 8.2.1 are the only ones that's missing. Would it be enough to just add them to `hieVersions` list?

fendor:
yes that is enough

bubba:
Can you check that #1032 is also covered in the Shakefile?

Anrock:
@fendor afaik it isn't handled right now, so it's a separate issue.

bubba:
@leifmetcalf being able to define it in terms of targets and build rules means it can work out the dependency graph for each binary and only build what’s needed. And it also provides the command line interface for free

leifmetcalf:
@bubba The current `install.hs` is just a bunch of `phony` rules, though.

bubba:
@leifmetcalf that is pretty confusing. I had to take a look at it again, but I think its because the path to the hie-x binaries are buried somewhere inside `.stack-work`, so the targets can't rely on Shake for detecting if its dirty or not since we don't know the path. But thankfully `stack` already takes care of not doing redundant rebuilding. The phony rules still provide dependencies though between the different levels of installation that can occur (e.g. build hie for one version or for all versions)

Anrock:
@bubba i think this is ready to land as soon as CI tests for install.hs are implemented.

bubba:
[...]

PR #1044, opened 2019-01-11
Rename Shakefile.hs to install.hs

Anrock:
Part of #1033

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -172,7 +172,7 @@ helpMessage = do
   let out = liftIO . putStrLn
   out ""
   out "Usage:"
-  out "    stack Shakefile.hs <target>"
+  out "    stack install.hs <target>"
   out ""
   out "Targets:"
   out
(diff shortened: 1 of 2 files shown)


install.hs
```diff
@@ -172,7 +172,7 @@ helpMessage = do
   let out = liftIO . putStrLn
   out ""
   out "Usage:"
-  out "    stack Shakefile.hs <target>"
+  out "    stack install.hs <target>"
```
Anrock: @fendor I wonder if hardcoded filename should be replaced with something like `argv[0]`. I doubt this file would be renamed again anytime soon, but one less place to check is a good thing.

fendor: I agree, that would be better.
In the next 2 weeks I'd like to add various improvements to it.

fendor:
The changes look good, the Readme has also been properly updated and it still works on my machine.

PR #1048, opened 2019-01-13
Change --recursive to --recurse-submodules

leifmetcalf:
The `--recursive` option is not listed in the [git-clone documentation](https://www.git-scm.com/docs/git-clone). It seems to have been renamed to `--recurse-submodules`. This commit changes the example `git clone` command in the readme to use the new option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ In order to avoid problems with long paths on Windows you can do the following:
 #### Download the source code
 
 ```bash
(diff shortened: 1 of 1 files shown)


alanz:
From what version of git did it change?

Because I would rather err on the side of supporting older git versions, than adopt the new parameter name, if it does not work for older versions.

alanz:
2010. Pretty sure it is safe :)

PR #1056, opened 2019-01-15
Make install.hs use tar executable

leifmetcalf:
This has some benefits:
- No `zlib` c library dependency
- No need for `shake.yaml`
- Stack `script` instead of `runghc`
- Everything contained in `install.hs`

`tar` works everywhere, so I see no reason to avoid it.

The behaviour of this code is slightly different. It creates a tar archive containing a folder named "." which in turn contains all the hie binary files. When the tar archive is extracted, it will extract all the hie binaries in the current folder as usual, so this shouldn't affect too much.

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -1,16 +1,9 @@
 #!/usr/bin/env stack
 {- stack
-  --stack-yaml shake.yaml
-  --install-ghc
-  runghc
+  script
+  --resolver lts-12.25
   --package shake
-  --package tar
-  --package zlib
 -}
-
-import qualified Data.ByteString.Lazy          as BS
-import qualified Codec.Archive.Tar             as Tar
-import qualified Codec.Compression.GZip        as GZip
 
 import           Development.Shake
 import           Development.Shake.Command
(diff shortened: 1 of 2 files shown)


Anrock:
@leifmetcalf you sure about tar being available on Windows out-of-the-box?

fendor:
You are kind of dropping the support for nix, which I would like to keep.
Also, the `tar` command does not exist on windows without further additional setup.
The advantage of `zlib` is that it is always installed when `git` is installed, thus, our only requirements are that `git` and `stack` are installed, no further build instructions required.

fendor:
@lorenzo could the windows binaries be distributed this way? E.g. Download the release and be done with it?

leifmetcalf:
At least on Ubuntu 18.10, `git` does not depend on the zlib development files, `zlib1g-dev`, only on `zlib1g`. I think the `zlib` haskell package depends on the zlib dev files, but I'm not sure. This could be a distro-specific thing.

On Windows 10, `tar` was only added at the end of 2017. I don't think it works at all (without powershell) on earlier Windows versions. It's probably too early for `tar` to be depended on.

Since the `dist` option won't be used by end-users, I don't think the `zlib` dependency matters.

fendor:
Then it might be sensible to actually drop the `dist` target if it has hardly any use?

bubba:
@fendor Yes I think it would make sense to move the `dist` target to a separate build script, to minimise dependencies.

fendor:
I see this following issues with this PR:

* Removal of `shake.yaml` removes the declaration of dependencies for nix.
* `tar` is not available on windows by default, neither in powershell nor cmd.
* Dependency of `zlib` is not removed but turned from a compile-time dependency into a run-time dependency.

I will open an issue to discuss the fate of the `dist` target.

PR #1079, opened 2019-02-05
Added support for user configuration 

colbyn:
This is something I added a little while ago and thought others may find this useful (includes a doc file `./docs/hie-yaml-file.md`).

The primary feature is to just simply disable interactive (as you type) analysis, as mentioned in the `./docs/hie-yaml-file.md` file. For a reference the RLS project has a similar option called `build_on_save` (` https://github.com/rust-lang/rls#configuration `).


colbyn:
Certainly.

I’ve just implemented the necessary changes for using the preexisting settings format, and updated the `README.md` file. I’m not sure how it’s supposed to work in general. Perhaps it’s editor specific, yet for the life of me I couldn’t figure out how the json value/file is exchanged (from looking at the `getConfigFromNotification` function)… 

So as a workaround if the value isn’t available it defaults to checking for a `settings.json` file that then defaults to the `Default` instance, i.e. `configVal (onSaveOnly configFile) onSaveOnly >>= \case`.

I suppose the name of this branch doesn’t make any sense now. I wasn’t sure what to rename it to so I left it as is.

alanz:
@colbyn 

The `settings.json` format is described so that LSP clients can know what options to provide when sending the configuration message through.  So strictly speaking managing settings is a client issue, not a server issue.

It is also not clear to my whether the `onSaveOnly` mechanism is the right one to use for this, the flag makes sense, but I am not sure whether it should rather be implemented via the `TextDocumentSyncKind` setting for `TextDocumentSyncOptions` from the server, to switch it from incremental to full.

bubba:
@colbyn The settings.json location depends on the client, for LanguageClient-neovim it should be in `.vim/settings.json`. For VS Code you need to expose it via the extension, which is done in javascript. [I recently created a PR to add a new option to it here](https://github.com/alanz/vscode-hie-server/pull/134)

Issue #1064, opened 2019-01-26
Way to Disable Brittany?


Weizilla:
Is there a way to disable brittany? I would like to use a different formatter but I'm guessing that would be too difficult so I'm ok with disabling brittany in hie and just having my editor (VS Code) call a different formatter on save.

bubba:
Does unselecting Format on Save in the VS Code settings disable the formatting for you? 
<img width="963" alt="image" src="https://user-images.githubusercontent.com/2488460/51780440-5a4c9700-2106-11e9-98dd-77d965e75475.png">

Weizilla:
Ideally, I would like to use a different formatter but still have it execute on save and when using the Format Code editor command. 

I thought maybe VS Code delegated everything to hie including the formatting so if I turned off brittany in hie, I could tell VS Code to use something else like hindent. Would that not work?

Weizilla:
There's an issue open with the VS Code HIE extension asking for the same thing: https://github.com/alanz/vscode-hie-server/issues/79

bubba:
I think we could use the [dynamic capability stuff](url
https://microsoft.github.io/language-server-protocol/specification#client_registerCapability) to disable formatting with HIE, not sure if the internals support it yet. 
We could also just add the other formatter into HIE ;)

Issue #626, opened 2018-06-10
upload to hackage


thkoch2001:
I'd like to eventually have haskell-ide-engine in Debian. Debian has nice tooling to package from hackage, so it would help to have hie present there.

I see that there is #328 (presence in Stackage) which seems related, but I understand that there is no dependency missing on hackage at this moment?

alanz:
Unfortunately there is a chain of dependencies missing, anchored on cabal-helper and ghc-mod

l29ah:
Bump? Both cabal-helper and ghc-mod are present on hackage.

l29ah:
Oh, okay, ghc-mod is not fresh enough.

alanz:
This is basically the same as #328

PR #1080, opened 2019-02-09
Add Floskell and formatterProvider

bubba:
This will expose an LSP config setting for changing the formatting provider, and so should also address #1064 

Also worth noting: This moves Brittany off of `IdeGhcM` so it doesn't get clogged in the request queue, so requests should be faster.


.gitmodules
```diff
@@ -32,3 +32,6 @@
 	url = https://github.com/alanz/ghc-mod.git
 
 
+[submodule "submodules/floskell"]
+	path = submodules/floskell
+	url = https://github.com/bubba/floskell
```
lorenzo: What did you have to change in your fork?

hie-plugin-api/Haskell/Ide/Engine/Config.hs
```diff
@@ -36,6 +37,7 @@ instance Default Config where
     , liquidOn                    = False
     , completionSnippetsOn        = True
     , formatOnImportOn            = True
+    , formattingProvider          = "brittany"
```
lorenzo: Could this be a datatype instead of some text?

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -795,6 +795,26 @@ reactor inp diagIn = do
 
 -- ---------------------------------------------------------------------
 
+getFormattingProvider :: R FormattingProvider
+getFormattingProvider = do
+  providers <- asks formattingProviders
+  lf <- asks lspFuncs
+  mc <- liftIO $ Core.config lf
+  -- LL: Is this overengineered? Do we need a pluginFormattingProvider
+  -- or should we just call plugins straight from here based on the providerType?
```
lorenzo: I would personally not have this complication and just call the plugin directly.

alanz: I think having an entry in the plugin is a good thing.

Eventually we can manage config based on the installed providers too, and it opens the way to dynamically (or statically but from config) loading plugins.  Which in my opinion has always been a goal.

alanz:
LGTM

.gitmodules
bubba: [Adjust version bounds and tweak it so it builds with earlier and later haskell-src-exts](https://github.com/ennocramer/floskell/pull/10). I've made a PR for the changes so that hopefully we can drop the reliance on the submodule, since it brings us one step further away from #626

hie-plugin-api/Haskell/Ide/Engine/Config.hs
bubba: I actually originally had this as a datatype! But when adding the `pluginFormattingProvider` field to the plugin descriptor it made more sense to revert it back to just the `PluginId`, that way if a new plugin is added the developer doesn't need to edit `Config.hs` as well. 
I do agree though, if there are any tricks to having the `PluginId`s exported in a more type safe manner then I would much prefer that

PR #1094, opened 2019-02-18
Show message for build-doc-% on make/shake help

Hogeyama:
Both `make help` and `./install.hs help` have not displayed a message for `build-doc-%`. 

I also noticed that the command name differs between make and shake (`build-doc-%` in make, `build-doc-hie-%` in shake). Is this intended?


fendor:
No this is actually not intended.
It would be nice to unify Makefile and Shake targets and adapt the README accordingly.

Hogeyama:
Thank you!

PR #1096, opened 2019-02-19
Load Floskell configuration

AlexeyRaga:
## Changes
- Respect `floskell` configuration file that can be found according to `floskell`'s rules.


lorenzo:
LGTM

PR #1103, opened 2019-02-24
WIP: Don't need to reversemap files

wz1000:
Goes along with the changes in https://github.com/alanz/ghc-mod/pull/17

This paves the way to removing mappedFiles altogether.

This also fixes a bug: With current master, open a file, and, without saving, write something like

```haskell
foo id = id
```

You will get a diagnostic for name shadowing, but the file mentioned in the diagnostic will be the temporary mapped file instead of the actual file. This behaviour will not happen after the changes in this PR are applied. 

This eliminates the need for reverseMap altogether, and probably fixes a bunch of other bugs in the process, involving returning the mapped file to vscode instead of the actual file.

Needs testing.  It also needs to tested against code with TH and CPP. 

To test, you will need to

```bash
$ cd submodules/ghc-mod
$ git remote add wz1000 https://github.com/wz1000/ghc-mod
$ git fetch wz1000
$ git checkout wz1000/proper-mapped-files
```
and then build normally.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
@@ -225,7 +225,11 @@ cacheModule uri modul = do
         let defInfo = CachedInfo mempty mempty mempty mempty rfm return return
         return $ case muc of
           Just (UriCacheSuccess uc) ->
-            let newCI = (cachedInfo uc) { revMap = rfm }
+            let newCI = oldCI { revMap = rfm . revMap oldCI }
+                    --                         ^^^^^^^^^^^^
+                    -- We have to retain the old mapping state, since the
+                    -- old TypecheckedModule still contains spans relative to that
+                oldCI = cachedInfo uc
               in uc { cachedPsMod = pm, cachedInfo = newCI }
           _ -> UriCache defInfo pm Nothing mempty
 


wz1000:
This does not work with CPP

alanz:
> This does not work with CPP

Is that an observed behaviour? Because as I understand it the line numbers should still line up. Or is it a case of not actually running the preprocessor?

wz1000:
@alanz this is the error that gets triggered: https://gitlab.haskell.org/ghc/ghc/blob/master/compiler/main/GhcMake.hs#L2454

bubba:
Quick sanity check on macOS is working. Looks like the code action functional tests are failing though

alanz:
@wz1000  yes, the missing preprocessing. CPP is such a PITA

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -213,7 +212,7 @@ lookupCachedData fp tm info dat = do
 cacheModule :: FilePath -> (Either GHC.ParsedModule GHC.TypecheckedModule) -> IdeGhcM ()
 cacheModule uri modul = do
   uri' <- liftIO $ canonicalizePath uri
-  rfm <- GM.mkRevRedirMapFunc
+  rfm <- return id -- GM.mkRevRedirMapFunc
```
bubba: Can we remove `revMap` from CachedInfo then?
https://github.com/haskell/haskell-ide-engine/blob/e1aeb180ab8065bd6a8413e050c6277191d0d33f/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs#L44-L52

Issue #1083, opened 2019-02-12
Expose build plugin


alanz:
The jsonstdio implementation has a build plugin, which is not currently exposed via LSP.

Should we expose it, and if so how?

Perhaps based on the Build System Protocol [BSP](https://github.com/scalacenter/bsp) as proposed by the Scala people. 

cc @arrowd

bubba:
The BSP looks promising: The specification seems to fit Cabal projects well, and the underlying JSON RPC protocol remains the same so we should be able to reuse some stuff. Although I think it would still probably be worthwhile to start a new package for it, such as `haskell-bsp` or the like.

And in the meantime, as long as we also expose LSP commands for build commands, we can integrate them into `vscode-hie-server` ourselves

PR #1105, opened 2019-02-25
Remove JSON transport

bubba:
This PR removes the JSON transport in an attempt to simplify the internal architecture.
I have removed a lot of stuff aggressively and I expect I've removed too much: Please let me know if I should put anything back in!

# Changes
- Plugins now define their `PluginId` themselves. Will help with plugins becoming standalone (and maybe eventually with dynamic loading)
- Plugin names and descriptions have been removed as they were only used by the JSON transport and were for displaying human readable strings when querying them
- Commands are now more tightly coupled with LSP commands, and so functionality in `GhcMod.hs` that was exposed as a command but not exposed to LSP no longer define commands. 
- Modules that weren't actually Plugins in `Haskell.Ide.Engine.Plugin` have been moved up a level
- `PluginId` and `CommandId` are `newtype`'d
- Case splitting has been moved to HaRe

# Todo
- [ ] Add back build plugin functionality and figure out how to expose it through LSP #1083


mpickering:
What's the motivation for this so I can understand the internal architecture better?

Anrock:
@mpickering there is an option for hie binary for producing non LSP-compliant JSON on stdio. AFAIK it was implemented long time ago when LSP wasn't a thing (at least in hie) and since that time nobody cared about it because LSP is the main protocol of hie for 99% of use cases. 
Basically this PR should throw away code for non-lsp json transport and readjust/simplify other code.

bubba:
Eventually, plugins may be loaded dynamically and would be specified by the user. If this PR removes plugin names then we need to look at namespacing and preventing conflicts when registering LSP command IDs.

mpickering:
Adding a name to a plugin is not that hard so you should probably just keep the field.

bubba:
@alanz @mpickering I just realised: The commands were already namespaced with `pluginId`, which was kept. `pluginName` was just a vanity name for display it looks like

app/MainHie.hs
```diff
@@ -42,29 +39,26 @@ import           Haskell.Ide.Engine.Plugin.Floskell
...
+      , brittanyDescriptor
+      , ghcmodDescriptor
+      , haddockDescriptor
+      , hareDescriptor
+      , hsimportDescriptor
+      , liquidDescriptor
+      , packageDescriptor
+      , pragmasDescriptor
+      , floskellDescriptor
       ]
```
alanz: The reason I put in the plugin id string when the plugin list is assembled is so that people can easily change the list of installed plugins, and only need to worry about possible name clashes in one place, namely here.  By delegating it to the individual plugins we lose this capability.

Unless there is some other mechanism to disambiguate them?

src/Haskell/Ide/Engine/Hoogle.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE CPP               #-}
 {-# LANGUAGE OverloadedStrings #-}
-module Haskell.Ide.Engine.Plugin.Hoogle where
+module Haskell.Ide.Engine.Hoogle where
```
alanz: What is going on here? Do we no longer have a `Plugin` namespace?

src/Haskell/Ide/Engine/Hoogle.hs
```diff
@@ -21,29 +21,6 @@ import           System.Environment
...
-      [ PluginCommand "info" "Look up the documentation for an identifier in the hoogle database" infoCmd
-      , PluginCommand "lookup" "Search the hoogle database with a string" lookupCmd
-      ]
-  , pluginCodeActionProvider = Nothing
-  , pluginDiagnosticProvider = Nothing
-  , pluginHoverProvider = Nothing
-  , pluginSymbolProvider = Nothing
-  , pluginFormattingProvider = Nothing
-  }
-
```
alanz: Where has this moved to? why?

bubba: The Hoogle plugin didn't have any providers, and the `lookup` and `info` commands were just used as a part of haddock when it provides documentation (which are now just called directly instead of via the reactor). If the plugin doesn't provide any providers, it becomes a no-op to HIE, so I thought it would make sense to remove it as a plugin and make it more of a utility module for Hoogle related functions. i.e. if the user disabled the Hoogle plugin, nothing would happen: haddock still calls the methods here regardless.

app/MainHie.hs
bubba: I think this might need a larger discussion about how loading plugins will eventually look like.
I'm also thinking it might make more sense to remove `pluginId` as a field from `PluginDescriptor`, and have HIE generate a unique id at launch instead?

bubba: Scrap that thought: It's used by liquid/ghc-mod and floskell/brittany to get specific plugins

alanz:
@bubba I think I broke this with the recent commits, it needs a refresh

alanz:
To avoid confusion, perhaps you should close the PR?

bubba:
Superseeded by #1489 and #1492

PR #1108, opened 2019-02-27
Remove some old cruft from README

DanielG:
See individual commits for justification for throwing these paragraphs away.

The basic idea of a README should be to welcome new users and point them in the right direction. Droning on about old projects isn't really helping that goal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,38 +584,17 @@ This project is not started from scratch:
  - Fork this repo and hack as much as you can.
  - Ask @alanz or @hvr to join the project.
 
-## Architecture
-
-Below you find a short overview of the main architectural layers of `haskell-ide-engine`.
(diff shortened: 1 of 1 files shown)


Anrock:
Personally i'm all in for README as welcoming and quick start document. Everything else should go to docs.

alanz:
LGTM.

A part of me wonders if/when we should move toward a readthedocs type approach to this.

bubba:
@alanz is it time to reconsider the github wiki?

alanz:
Well, people are more likely to propose changes to a doc via a PR, tend not to want to edit a Wiki, because there is no approval stage.

So docs in the repo as we have now is good, with a PR flow.

So I guess we just need to consider how we organise the docs we have, so that it is useful.

And as a later/different concern decide whether we need to separately publish the docs. It is probably too early for that.

PR #1109, opened 2019-02-27
Bump floskell

luigy:
point back to upstream version


bubba:
Just took a quick look: looks like this is the commit that added support for the older GHC versions:
https://github.com/ennocramer/floskell/commit/6ed5af936ead2675e6a720f4f0538dc3c8d4c06b

PR #1118, opened 2019-03-07
Add GHC 8.6.4 support

alanz:
Some tests fail, need to investigate


fendor:
Build script support for 8.6.4 is missing. 
It should suffice to add "8.6.4" to `hieVersions`

alanz:
Yes, I saw that. After making a duplicate.  I just want to confirm these tests pass first, as I figured it makes little difference in the scheme of things.  Ideally the new `monad-dijkstra` ends up on hackage soon, and we can toss the dependency completely.

PR #1120, opened 2019-03-08
Remove make section from readme

Anrock:
As a step to deprecation of Make\powershell script, i'm removing mentions about them from README, leaving only shake script as installation method. Makefile\powershell script itself are still there.

* Remove sections about makefile usage and build-all.ps1from README.
* Minor improvements / rewording, reshuffling sections
* I've deliberately removed sections about `build-all`-like targets. IMO there are no scenarios when end-users will actually need it, but there were a number of issues when users were confused with 20gb installations and long build times

@alanz @fendor @bubba please review changes, in case i've deleted something important.

Part of #1033


README.md
```diff
@@ -142,7 +142,13 @@ To install HIE, you need stack version >= 1.7.1.
 
 HIE builds from source code, so there's a couple of extra steps.
 
-#### Linux pre-requirements
+#### Common pre-requirements
+
+* `stack` must be in your PATH
```
fendor: Maybe clarify that the `stack` command must be fully functioning? 
E.g. on ubuntu, I installed stack and `stack setup` failed out of the box?
But maybe we can expect that everybody understands that in your path means, it must be working?

Anrock: Hm. IIRC, stack will complain if `stack setup` wasn't ran before and provide a suggestion to run `stack setup`.
I can also add a small clarification like "If you run `git --version` and see an output instead of 'command not found' message, that means that git is in your PATH".

Anrock: OTOH, all that time before wording was the same and there were no issues regarding git/stack/whatever not being in PATH or stack setup. But i guess i'll add clarification anyway.

fendor: You dont have to, you are probably right.

alanz:
LGTM

PR #1125, opened 2019-03-10
update the snippet for .spacemacs/user-config

chkl:
The function `lsp-haskell-enable` does not seem to exist anymore.

Per https://github.com/emacs-lsp/lsp-haskell the lsp-haskell mode should be enabled with `   (add-hook 'haskell-mode-hook #'lsp)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,8 +458,9 @@ and then activate [`lsp-haskell`](https://github.com/emacs-lsp/lsp-haskell) in y
 ```lisp
 (defun dotspacemacs/user-config ()
   "..."
+  (setq lsp-haskell-process-path-hie "hie-wrapper")
   (require 'lsp-haskell)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -458,8 +458,9 @@ and then activate [`lsp-haskell`](https://github.com/emacs-lsp/lsp-haskell) in y
 ```lisp
 (defun dotspacemacs/user-config ()
   "..."
+  (setq lsp-haskell-process-path-pie "hie-wrapper")
```
alanz: Is `lsp-haskell-process-path-pie` a typo? I would expect the last part to be `hie`, not `pie`

chkl: Oh. damn it. Just 2 lines and I still couldn't get it right on the first attempt. Fixed!

alanz: Some study showed one-liner bug fixes were the most likely to be wrong, and have to be re-done.

So you are on the curve :)

PR #1128, opened 2019-03-11
Fix table of contents (installation section) in README.md

antonlogvinenko:
Links in TOC section describing installation are broken. I fixed the links, but I had to rename inner "Installation" section to "Installation with Shake" since there already exists "Installation" section (parent) and #installation link would wrongly refer to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,16 +24,13 @@ we talk to clients.__
     - [Installation with Nix](#installation-with-nix)
     - [Installation on ArchLinux](#installation-on-archlinux)
     - [Installation from source](#installation-from-source)
-      - [Linux pre-requirements](#linux-pre-requirements)
-      - [Windows: long paths (optional)](#windows-long-paths-optional)
+      - [Common pre-requirements](#common-pre-requirements)
+      - [Linux-specific pre-requirements](#linux-specific-pre-requirements)
+      - [Windows-specific pre-requirements (optional)](#windows-specific-pre-requirements-optional)
       - [Download the source code](#download-the-source-code)
-      - [Choose your GHC version](#choose-your-ghc-version)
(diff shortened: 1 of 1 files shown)


Anrock:
Oops, my bad, totally forgot about TOC.

Wouldn't it be better to rename installation with Shake to `Building` or `Compiling` maybe? I specifically removed `with Shake` pieces everywhere as unneeded clarification after Makefile and powershell scripts mentions were removed.

Also, personally these installation sections inside installation section look confusing to me. Maybe they should be moved from under installation to top level? @alanz @fendor

antonlogvinenko:
> Wouldn't it be better to rename `installation with Shake` to `Building` or `Compiling` maybe? I specifically removed `with Shake` pieces everywhere as unneeded clarification after Makefile and powershell scripts mentions were removed.
Sure, I've updated PR and renamed to "Building".

> Also, personally these installation sections inside installation section look confusing to me. Maybe they should be moved from under installation to top level? @alanz @fendor
I thought about it, and noticed that the contents of "Building" section logically belong to "Installation from source" section. Maybe we should merge them? Then there will be just "Installation" section with three subsections "From Nix", "on ArchLinux", "From source" — these three then could be moved to top level.

Anrock:
@antonlogvinenko thanks for update, LGTM then.

Regarding sections: building section is definitely belongs to installation from source as a direct continuation after installing required libs and downloading source code.

antonlogvinenko:
> Regarding sections: building section is definitely belongs to installation from source as a direct continuation after installing required libs and downloading source code.
Ok, I moved "Building" section inside "Installation from source".

Anrock:
@antonlogvinenko thank you

PR #1136, opened 2019-03-16
Remove dist target

### LINKED ISSUE #1059: Purpose of target `dist` in install.hs
In the PR #1056 it is noted that target `dist` is usually not necessary for installation of HIE, thus, the dependency `zlib` could be removed from install.hs.

Things we could do with `dist`:
* Drop it without replacement
* Leave it unchanged in install.hs
* Extract a build script, potentially written in `shake`

fendor:
Closes #1059 

No one claimed that they are actively using it, thus, remove to reduce dependencies.

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -5,14 +5,7 @@
   runghc
   --package shake
   --package directory
-  --package tar
-  --package zlib
 -}
-
-import qualified Data.ByteString.Lazy          as BS
-import qualified Codec.Archive.Tar             as Tar
-import qualified Codec.Compression.GZip        as GZip
-
 import           Development.Shake
 import           Development.Shake.Command
 import           Development.Shake.FilePath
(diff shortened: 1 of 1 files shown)


power-fungus:
@leifmetcalf This is not true. If no stack yaml is not given within the Script, then it finds the default one. In there, the submodules are referenced. When these are not initialized, the script fails to start. This was the reason for a new stack yaml

leifmetcalf:
> @leifmetcalf This is not true. If no stack yaml is not given within the Script, then it finds the default one. In there, the submodules are referenced. When these are not initialized, the script fails to start. This was the reason for a new stack yaml

Right, which is why you would use [`stack script`](https://docs.haskellstack.org/en/stable/GUIDE/#stack-configuration-for-scripts)

PR #1140, opened 2019-03-18
Remove non shake scripts

Anrock:
Part of #1033


Anrock:
@alanz @bubba @fendor are we ready for that?

fendor:
In my opinion, we are.

alanz:
So long as the README no longer has any references to these I am happy to do this.

Issue #1116, opened 2019-03-03
./install.hs build is not idempotent


alanz:
For current master (tag 0.7.0.0), on Debian Testing, if I run `./install.hs build` repeatedly, it copies the pre-build versions of all except for GHC 8.6.2, which it rebuilds every time, saying that `temporary` is not installed.

See https://gist.github.com/alanz/ffe5c4c0fa1b3fe79a3e2183dd033082

alanz:
cc @fendor

fendor:
Is fixed by setting the resolver in `shake.yaml` to something different to the other `stack-*.yaml` files.
What would be a sensible resolver? I think we would only download a different snapshot, thus, no ghc is downloaded twice.

alanz:
What is the cause of it, if that is the fix? Seems odd to me.

fendor:
`temporary` is an extra dep in `stack-8.6.2.yaml` and not in `shake.yaml`. In the latter, a different version of temporary is needed, thus, the respective temporary package is unregistered.
`shake.yaml` is currently being used to install `cabal-install`, to avoid that a nightly resolver is being used.

alanz:
ok, so making sure that we use the same version of temporary for both should sort out the problem? Perhaps we should check if we already have one that does not override it, or set `shake.yaml` to have the override too.

fendor:
That would increase the compilation time of the shake file even more.
But yes, we could use the same extra deps everywhere

fendor:
I think, choosing a different resolver is faster and more efficient than trying to sync up the dependencies between any `stack-*.yaml` and `shake.yaml`.

fendor:
@alanz Can you check if this really fixes the issue?

fendor:
Fixed on windows and nixos, at least.

alanz:
@fendor I can confirm it works for me. Thanks.

PR #1141, opened 2019-03-18
Remove redundant shake.yaml

leifmetcalf:
As of #1136 we no longer need the nix config in `shake.yaml`.

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -1,8 +1,7 @@
 #!/usr/bin/env stack
 {- stack
-  --stack-yaml shake.yaml
-  --install-ghc
-  runghc
+  script
+  --resolver nightly-2018-12-17
   --package shake
   --package directory
 -}
--- a/install.hs
+++ b/install.hs
@@ -1,7 +1,7 @@
 #!/usr/bin/env stack
 {- stack
   script
-  --resolver nightly-2018-12-17
+  --resolver nightly-2018-12-15
   --package shake
   --package directory
 -}


power-fungus:
**This does not work** (as described in [my comment](https://github.com/haskell/haskell-ide-engine/pull/1136#issuecomment-473623859))

On a fresh clone, this fails with the following message (just tested)
```
Stack looks for packages in the directories configured in the 'packages' and 'extra-deps' fields defined in your stack.yaml
The current entry points to $PWD/submodules/HaRe/ but no .cabal or package.yaml file could be found there.
```
The reason is that the script finds the `stack.yaml` file and tries to execute this script in the environment specified there.

Beeing able to run the script even with uninitialized submodules is a very critical feature. Before, there were a lot of issues regarding build errors where the submodules where not initialized correctly

fendor:
I agree, I dont think we need the `shake.yaml` anymore.

install.hs
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env stack
 {- stack
-  --stack-yaml shake.yaml
-  --install-ghc
-  runghc
+  script
+  --resolver nightly-2018-12-17
```
fendor: resolver should not be equal to a resolver in one of the `stack-$VER.yaml`'s. See #1116 and the respective fix #1137.
tldr, I request changing the resolver to `nightly-2018-12-15`

leifmetcalf: Done!

PR #1145, opened 2019-03-25
Actually delete shake.yaml

leifmetcalf:
Somehow this didn't make it into the original PR, sorry!


Anrock:
@leifmetcalf please resolve conflicts

power-fungus:
While investigating the build system in more detail, I just discovered that the `shake.yaml` is still used in a few places inside the `install.hs` file. At @fendor's request, I will review the whole `install.hs` file and will delete the references `shake.yaml` in there, if possible.

I recommend closing this PR since it does not leave the build system in a consistent state.

PR #1150, opened 2019-03-26
Always overwrite hoogle

### LINKED ISSUE #1147: Hoogle already exists during shake installation
During installation with 
`stack ./install.hs cabal-build-all`, the following is raised:

```
Symlinking 'hoogle'
cabal: Path '/Users/timpierson/.cabal/bin/hoogle' already exists. Use
--overwrite-policy=always to overwrite.

  at src/Development/Shake/Internal/Args.hs:81:69-79:
* Depends on: cabal-build-all
  at /Users/timpierson/arity/haskell-ide-engine/install.hs:91:31-71:
* Depends on: cabal-build-docs
  at /Users/timpierson/arity/haskell-ide-engine/install.hs:92:31-77:
* Depends on: cabal-build-doc-8.6.2
* Raised the exception:
user error (Development.Shake.command_, system command failed
Command: cabal new-install hoogle
Exit code: 1
Stderr:
cabal: Path '/Users/timpierson/.cabal/bin/hoogle' already exists. Use
--overwrite-policy=always to overwrite.

)
```

This persists even after deleting `~/.cabal/bin/hoogle`

fendor:
closes #1147 

Always overwrite the executable, don't fail.

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -136,8 +136,8 @@ updateSubmodules = do
 
 validateCabalNewInstallIsSupported :: Action ()
 validateCabalNewInstallIsSupported = when (os `elem` ["mingw32", "win32"]) $ do
-  liftIO $ putStrLn $ embedInStars cabalInstallNotSuported
-  error cabalInstallNotSuported
+  liftIO $ putStrLn $ embedInStars cabalInstallNotSupported
+  error cabalInstallNotSupported
 
 configureCabal :: VersionNumber -> Action ()
 configureCabal versionNumber = do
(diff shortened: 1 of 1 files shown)


mpickering:
Is the intention here that if a user has already installed hoogle themselves then this installer will overwrite it!?

I don't think the installation script should overwrite anything a user has done at all. If HIE needs some specific executables then it should surely install them somewhere it manages itself?

fendor:
While I agree, we actually already do that when using `stack`.

fendor:
Close in favour of #1151

PR #1152, opened 2019-03-26
Support for Pattern Synonyms

anton-dessiatov:
Please don't merge as is, let's wait if alanz/haskell-lsp#143 gets merged.

The thing is that when used on modules that have pattern synonym declarations, hie kept crashing due to ```error``` call at line 693 of src/Haskell/Ide/Engine/Plugin/GhcMod.hs file. I have fixed this, but it required to introduce a new SymbolKind in ```haskell-lsp``` package.

This pull request reflects current state of things including referencing my forked version of ```haskell-lsp```. Once required changes to ```haskell-lsp``` get released, I'm going to update this pull request and remove my fork reference.

Meanwhile if there are problems with an approach I've taken, I have plenty of time to fix them.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
+++ b/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
@@ -682,20 +682,29 @@ symbolProvider uri = pluginGetFile "ghc-mod symbolProvider: " uri $
           map (\n -> Decl LSP.SkVariable n [] l) $ hsNamessRdr p
 
 #if __GLASGOW_HASKELL__ >= 806
+        goValD (L l (PatSynBind _ idR)) = case idR of
+          XPatSynBind _ -> error "xPatSynBind"
+          PSB { psb_id = ln } ->
+#else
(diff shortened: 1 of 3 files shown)


anton-dessiatov:
It turns out that one does not simply add a new Symbol Kind to Microsoft LSP. I have removed custom Symbol Kind and ended up using Function kind for pattern synonyms. Please merge this pull request if you consider it appropriate.

anton-dessiatov:
Whoops, it fails for GHC 8.6+. Will fix now.

anton-dessiatov:
Fixed, but leaved out an 'error' call in case for GHC 8.6+ pattern synonym binding evaluates to XPatSynBind. I'm no GHC expert and could use some hint on under which circumstances could it ever evaluate to XPatSynBInd. In my testing environment I never managed to witness that case (tested with GHC 8.6.4).

I would say that anyway it's better than before and could probably be merged.

anton-dessiatov:
Sure! Added tests.

bubba:
@alanz is this good to merge?

Issue #1146, opened 2019-03-25
Under no circumstances should HIE ever stop responding


KillyMXI:
The title is a quote from #461.

I was looking through opened issues and it really struck me how common this pattern:
On any unexpected input, HIE just shuts in itself.
And it is a worst way to handle the situation when we have a really unstable piece of software with a lot of things "unexpected".
As a user, I:
- can't continue my work at least with the parts that known to work;
- have to go through a lot of trouble to gather debug information.

The issues with the same pattern:
- #451 
- #461 
- #556 
- #983 
- https://github.com/alanz/vscode-hie-server/issues/124 (I came from there initially)
- maybe more

In no way we can work productively with HIE while you will unhurriedly play whack-a-mole with the issues.
I don't want to learn to avoid certain lines just because you can't handle them.
And restarting HIE to recover it's functionality takes too long.

Messaging itself MUST be more fault-tolerant.
- Failed to handle a request? - report back as useful failure message as you can in given situation and get ready to process following requests. This is not about debug logging - this is to explain to user why his request wasn't handled;
- IDE extensions should be able to tell user what went wrong and what can be done about it;
- User should be able to use other parts of HIE functionality while you're working on fixes for reported issues.

mpickering:
Seems like this should be one of the higher priority issues to fix.

Anrock:
Agreed. I've pinned the issue to help with visibility.

alanz:
I think the first step is to have some reproducable bug reports.

Because the underlying architecture is supposed to make it robust, in that the main loop keeps processing regardless, we have deferred execution to provide a result if the GHC invocation fails, and for things like Hover it should respond immediately based on their being a currently cached result.

mpickering:
I think actions like "hover" block because the implementation was inefficient. 

I also observed that reporting a large number of diagnostics (500) causes something to pause, not sure if that was a HIE or vscode issue.

alanz:
I am pretty sure that the default setting for number of issues is to limit it to 100. Which are sorted to be the first ones, so perhaps the sorting process is where the delay comes from.

alanz:
And I think the current delayer is requesting completions, if anyone wants to take a look. It should work like Hover, if there is no currently cached module it returns immediately with an empty result.

bscottm:
[...]

Issue #1143, opened 2019-03-21
Hlint data-files can not be easily distributed


fendor:
`hlint` demands that its data-file locations are known at compile time. If you remove these files, for example `hlint.yaml`, hlint will fail with a message and lets `hie` crash

```
hlint: user error (Failed to find requested hint files:
  ~/.cabal/store/ghc-8.4.4/hlint-2.1.15-8782d6ab3dc227b1c49b92fdeae8ecd837138c4d8f742c2f87f0c13927a32609/share/hlint.yaml
)
```
In this case, hlint has stored its data-files in `~/.cabal/store`.

Unfortunately, if you copy the executable `hie` instead of installing from source, `hie` itself chokes on that error, since the data-files are not copied, thus, not available. We should, at the very least, catch that error and avoid crash.

Following options to resolve this issue:
* Fix this upstream, hlint should just create this file in some location or use defaults.
* Provide all data-files in a custom datadir directory.
* Turn `hlint` into an official run-time dependency and demand a working installation.

This is a blocker for #1068

power-fungus:
Moreover: `hie` it crashes when `hlint` does, which is undesirable (in my opinion). Can this be fixed?

nponeccop:
Is this a duplicate of #400?

power-fungus:
@nponeccop It does tackle a similar issue. 

I am currently in the process of evaluating runtime-dependencies of `hie` and found out that the 2 issues described are still relevant.
However, the runtime-dependency for `hlint` should be tackled in a seperate issue to the one for `cabal-helper`.

fendor:
I propose the following, slightly weird, solution:
Use the package `file-embed` to embed the content of the data-files at compile time into `hie`.
On startup, check if these files exist at a pre determined location, e.g. `$XDG_DIRECTORY/haskell-ide-engine/hlint`. If yes, supply it to hlint as the `--data-dir`. Otherwise, write the content of the embedded files into that directory and continue.

Pro:
* Solves the run-time dependency of hlint on its data-files
Con:
* Might be error prone.
* Increases size of `hie` binary.

power-fungus:
I would like to tackle this at install-time. My plan:
* in `install.hs`: call `cabal get hlint`, then store the `hlint/data` into `$HIE_DIR/hlint-data`
* in `hie`: call `hlint` with `--datadir=$HIE_DIR/hlint-data`

[...]

PR #1153, opened 2019-03-27
Prevent hie crash if hlint crashes

fendor:
Related to #1143 

Also improves stability of hie, thus related to #1146 
If hlint chokes on anything, dont kill hie.
An error message might be desirable but I dont know how to do that.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
+++ b/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
@@ -5,7 +5,10 @@
 module Haskell.Ide.Engine.Plugin.ApplyRefact where
 
 import           Control.Arrow
-import           Control.Lens hiding (List)
+import           Control.Exception              ( IOException
+                                                , try
+                                                )
+import           Control.Lens            hiding ( List )
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Except
 import           Data.Aeson                        hiding (Error)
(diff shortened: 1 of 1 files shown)


mpickering:
There are already other functions which catch exceptions in HIE. Did you look at using one of those rather than rolling your own using `try`? In particular catching`SomeException` seems like quite a broad brush.

fendor:
Indeed, catching `SomeException` is terrible.

fendor:
Now we catch `IOExceptions`!

power-fungus:
@mpickering I am actually in favor of aggressive error-catching from `hlint`. I think the `hie` language server should not be brought down by a internal error in any plugin like `hlint`. 

It is the best thing for `hie` to do is to stop the plugin `hlint` and possibly restart it since `hie` is still quite useful if one plugin is removed.

mpickering:
@power-fungus The problem is catching async exceptions such as interrupts.

power-fungus:
Indeed, async exceptions should not be caught

bubba:
I would argue that language servers should be a bit overdramatic and crash frequently: VS Code automatically restarts the server if it dies, and from what I recall with the ghc-mod plugin, the design is to propagate as many fatal errors as possible

fendor:
Hlint errors dont have to be fatal ones though. In that case, hlint would just not work, while other functionality is still working.

Anrock:
I'm against being overdramatic, too. While VSCode may handle this well, LanguageClient-neovim stumbles on server crashes and i suspect other client may do this too. 

As a side note: LSP should be general overall, so coding against behaviour of specific client isn't right thing to do - remember embrace&extend&extuinguish. There are plenty of VSCode-specific quirks in other clients/servers already.

Avi-D-coder:
The problem with crashing frequently is many language clients don't restart servers. If the server can recover it should try to. If possible plugins should be restarted. 
A message on stderr might also be a good idea.   Does anyone know how many clients handle stderr?

alanz:
LGTM, now it is catching IOException.

PR #1164, opened 2019-04-06
Only run diagnostics onSave

lorenzo:
This introduces a behaviour change in that hie will only run diagnostics on a file after it has been open and saved, unless the user hasn't manually configured the client to run diagnostics on chages.

I have found that running diagnostics as I type creates too much noise in my editor as more often than not I'm in a broken state. Additionally, this it makes HIE appear that it runs faster.

* Added the `diagnosticsOnChange` client config value
* Refactored the reactor a bit to not have hardcoded defaults smattered everywhere, which in some cases were inconsistent.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Config.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Config.hs
@@ -33,7 +33,7 @@ data Config =
 instance Default Config where
   def = Config
     { hlintOn                     = True
-    , diagnosticsOnChange         = False
+    , diagnosticsOnChange         = True
     , maxNumberOfProblems         = 100
     , diagnosticsDebounceDuration = 350000
     , liquidOn                    = False
--- a/hie-plugin-api/Haskell/Ide/Engine/Config.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Config.hs
@@ -21,6 +21,7 @@ getConfigFromNotification (NotificationMessage _ _ (DidChangeConfigurationParams
 data Config =
   Config
     { hlintOn                     :: Bool
+    , diagnosticsOnChange         :: Bool
     , maxNumberOfProblems         :: Int
     , diagnosticsDebounceDuration :: Int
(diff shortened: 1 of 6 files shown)


bubba:
I’m confused as to how the tests are passing, since they aren’t sending any didSave notifications.

lorenzo:
@bubba only 2 tests where calling `noDiagnostics` after executing a text change. The other tests were only asserting that diagnostics were returned after a file was opened.

bubba:
@lorenzo some of the tests do send applyedit requests though:
https://github.com/haskell/haskell-ide-engine/blob/4c64789597cec9e73c9aeb901d9f6d0bb58251d9/test/functional/CompletionSpec.hs#L16-L30

lorenzo:
> Are there any other language servers that only return on save by default

All of the language servers I used for work do it on save: Golang, Python, Elm, PHP... The one I struggle with the most is the Haskell one because of all the constant noise I get while I'm in an obviously broken state as I move stuff around. I'm confident most people must be having the same type of problems as I do.

The problem is not the latency, which is also annoying, but the constant showing of errors as I type.

lorenzo:
> some of the tests do send applyedit requests though:

Right, but the test is not doing any assertions on the diagnostics after changing the test. The only thing that I changed here is running diagnostics, the rest is kept unchanged.

alanz:
I agree that the default should be to preserve the current behaviour, rather than only generating diagnostics on save.

bubba:
> The problem is not the latency, which is also annoying, but the constant showing of errors as I type.

I do agree though that we should definitely support this workflow. I can also imagine a wide variety of scenarios such as having really large modules which take a long time to typecheck where on save behaviour is probably preferred.

lorenzo:
Ok, seems there does not seem to be lots of support for making this the default, I'll make it a configuration option only and revert to the previous default.

lorenzo:
@bubba @alanz I set the default to `True` to preserve the old behaviour. This can be merged whenever you like

PR #1167, opened 2019-04-08
Change Hsimport to use configured Formatter on import

### LINKED ISSUE #1115: importModule always formats with brittany
We now have formatProviders, but the `importModule` function always uses brittany if `formatOnImport` is enabled.

Rather use the configured formatter.

fendor:
Closes #1115 

Adds a function to get the configured formatter plugin.
Configures HsImport to use the configured formatter.

Add PluginCommands to format lines of text for Birttany and Floskell.
Add FormatTextCmdParams for the Cmd.

TODO:
* ~[ ] Formatting Options should be honored.~ Will not be adressed in this PR.


src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -79,25 +75,51 @@ importModule uri modName =
...
+                    )
+            else 
               return $ IdeResultOk (J.WorkspaceEdit mChanges mDocChanges)
 
-  where formatTextEdit confFile (J.TextEdit r t) = do
-          -- TODO: This tab size of 2 spaces should probably be taken from a config
-          ft <- fromRight t <$> liftIO (Brittany.runBrittany 2 confFile t)
-          return (J.TextEdit r ft)
+ where
+flatten :: List [a] -> List a
```
bubba: Are these functions supposed to be inside the where? The indentation might be putting them at the top level

src/Haskell/Ide/Engine/Support/HieExtras.hs
```diff
@@ -799,3 +801,12 @@ prefixes =
   , "$c"
   , "$m"
   ]
+
+-- ---------------------------------------------------------------------
+
+getFormattingPlugin :: Config -> IdePlugins -> Maybe (PluginDescriptor, FormattingProvider)
+getFormattingPlugin config plugins = do
```
bubba: Can this happen inside `R` so `Config` and `IdePlugins` don’t need to be passed?

fendor: Hardly, since IdeGhcM does not implement R, or can execute R, so other plugins would have no access to it.

src/Haskell/Ide/Engine/Plugin/HsImport.hs
fendor: Yes they are, maybe one of the formatters did something weird.

src/Haskell/Ide/Engine/Support/HieExtras.hs
```diff
@@ -799,3 +801,12 @@ prefixes =
   , "$c"
   , "$m"
   ]
+
+-- ---------------------------------------------------------------------
+
+getFormattingPlugin :: Config -> IdePlugins -> Maybe (PluginDescriptor, FormattingProvider)
+getFormattingPlugin config plugins = do 
+  let providerName = formattingProvider config
+  fmtPlugin <- Map.lookup providerName (ipMap plugins)
```
fendor: This is actually wrong, since we look up the name, but the Plugins Map is of Type `Map PluginId PluginDescriptor`

fendor: But it works, since due to some lucky coincidence `pluginId == pluginName`.

fendor: I think it would be better if PluginId would be a `newtype`.

src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -79,25 +75,51 @@ importModule uri modName =
...
+
+  formatTextEdit :: PluginDescriptor -> J.TextEdit -> IdeGhcM [J.TextEdit]
+  formatTextEdit plugin edit@(J.TextEdit r t) = do
+    result <- runPluginCommand
+      (pluginId plugin)
+      "formatText"
+      -- TODO: should this be in the configs?
+      (dynToJSON $ toDynJSON $ FormatTextCmdParams t
+                                                   r
+                                                   (FormattingOptions 2 True)
```
fendor: How could we encode this in a way that it is available to a IdeGhcM?

src/Haskell/Ide/Engine/Plugin/Brittany.hs
```diff
@@ -25,47 +25,98 @@ data FormatParams = FormatParams Int Uri (Maybe Range)
...
 brittanyDescriptor :: PluginId -> PluginDescriptor
 brittanyDescriptor plId = PluginDescriptor
-  { pluginId       = plId
-  , pluginName     = "Brittany"
-  , pluginDesc     = "Brittany is a tool to format source code."
-  , pluginCommands = []
+  { pluginId                 = plId
+  , pluginName               = "Brittany"
+  , pluginDesc               = "Brittany is a tool to format source code."
+  , pluginCommands           = [ PluginCommand "formatText"
```
bubba: I would prefer if we can use the `pluginFormattingProvider`s directly rather than calling a `PluginCommand`, since `PluginCommands` get exposed to LSP

src/Haskell/Ide/Engine/Plugin/HsImport.hs
bubba: I think we can just use the `formatting provider` directly here no? Since Hie.getFormattingPlugin should return both the plugin and the `FormattingProvider`, which is just a function.

src/Haskell/Ide/Engine/Support/HieExtras.hs
bubba: I see, I didn't notice it was being called elsewhere outside of LspStdio.hs. I feel like this should be moved into PluginIdeMonads.hs though, and made polymorphic over `MonadIde`, which should have access to both. We'll need to figure out how to give `R` access to it though

src/Haskell/Ide/Engine/Plugin/HsImport.hs
fendor: FomattingProvider works in IdeDeferM. Can this be integrated into IdeGhcM? I could not find a way.

fendor: Moreover, R doesnt know the `FormattingOptions` either.

src/Haskell/Ide/Engine/Support/HieExtras.hs
fendor: I think what we really need is some kind of way to let plugins call each other. A bunch of plugins need to format after they do something.
Maybe a separate issue for discussion would be nice?

src/Haskell/Ide/Engine/Plugin/Brittany.hs
fendor: I think what we really need is some kind of way to let plugins call each other. A bunch of plugins need to format after they do something.
Maybe a separate issue for discussion would be nice?

bubba:
@fendor I've created a PR for your PR on your fork, it changes HsImport to use the formatting providers directly, by applying `TextEdits` to `TextEdits`

fendor:
Imo, ready to merge when CI has finished.

fendor:
Ok, that was awkard. Although these are not clean commits, I think everything should be as expected now...

src/Haskell/Ide/Engine/Plugin/Floskell.hs
```diff
@@ -31,20 +31,16 @@ floskellDescriptor plId = PluginDescriptor
...
-              FormatDocument -> (fullRange contents, contents)
-              FormatRange r  -> (r, extractRange r contents)
-            result = reformat config (uriToFilePath uri) (BS.fromStrict (T.encodeUtf8 selectedContents))
-        in  case result of
-              Left  err -> return $ IdeResultFail (IdeError PluginError (T.pack err) Null)
-              Right new -> return $ IdeResultOk [TextEdit range (T.decodeUtf8 (BS.toStrict new))]
+    let (range, selectedContents) = case typ of
+          FormatDocument -> (fullRange contents, contents)
+          FormatRange r  -> (r, extractRange r contents)
+        result = reformat config (uriToFilePath uri) (BS.fromStrict (T.encodeUtf8 selectedContents))
```
mpickering: Why use `uriToFilePath` and not the `file` variable?

fendor: No particular reason, it was like that when I found it.

src/Haskell/Ide/Engine/Plugin/Brittany.hs
```diff
@@ -20,52 +17,62 @@ import qualified Language.Haskell.LSP.Types.Lens       as J
...
-                      (T.pack $ "brittanyCmd: " ++ unlines (map showErr err)) Null)
-          Right newText ->
-            return $ IdeResultOk [J.TextEdit (fullRange text) newText]
+provider
+  :: MonadIO m
+  => Text
+  -> Uri
+  -> FormattingType
+  -> FormattingOptions
+  -> m (IdeResult [TextEdit])
```
mpickering: Do you use this at another type of than `IdeM`?

fendor: No. It is an abstraction, so it might be used in another type, but not yet. Afaik, by using another Monad, it might be possible to either wait for a response or to get it immediately. 
If this is bad style, we can drop it, though.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -212,16 +211,24 @@ type SymbolProvider = Uri -> IdeDeferM (IdeResult [DocumentSymbol])
...
+-- Failing means here that a IdeResultFail is returned.
 -- This can be used to display errors to the user, unless the error is an Internal one.
 -- The record 'IdeError' and 'IdeErrorCode' can be used to determine the type of error.
-type FormattingProvider = Uri -- ^ Uri to the file to format. Can be mapped to a file with `pluginGetFile`
+type FormattingProvider = T.Text -- ^ Text to format
+        -> Uri -- ^ Uri of the file being formatted
         -> FormattingType  -- ^ How much to format
         -> FormattingOptions -- ^ Options for the formatter
-        -> IdeDeferM (IdeResult [TextEdit]) -- ^ Result of the formatting or the unchanged text.
+        -> IdeM (IdeResult [TextEdit]) -- ^ Result of the formatting or the unchanged text.
```
alanz: This used to be `IdeDeferM` because brittany needs the GHC parsed source to work. Now that it is `IdeM`, will be potentially lose formatting under some circumstances?

bubba: Brittany's `FormattingProvider` doesn't use `ifCachedModule/withCachedModule` on master at the moment, so it looks like it wouldn't have been waiting for the parsed source
https://github.com/haskell/haskell-ide-engine/blob/a48a02cbd7efdf5ddc923ab7458921398a8c29cd/src/Haskell/Ide/Engine/Plugin/Brittany.hs#L42-L64

It looks like I moved it from `IdeGhcM` to `IdeM` in this [commit](https://github.com/haskell/haskell-ide-engine/commit/d39b4eec4fd0c8e4fcb406b425f98f4a3dda0e3a), and in retrospect I should have given a second thought as to why it was In `IdeGhcM` in the first place. 
`runBrittany` just runs in `IO`: does it access the same ghc-mod instance as HIE?

alanz: > runBrittany just runs in IO: does it access the same ghc-mod instance as HIE?

Good question. I think once my HaRe update to use the new hie-bios is done we will have a clearer view of the API, and can ask brittany to expose something that makes use of the module caching

If it does not, then IdeDeferM makes no difference.

fendor:
In my opinion, this is ready to merge.

Issue #1008, opened 2018-12-25
Note: GHC 8.6.3 broken on windows for TH


alanz:
It segfaults.

See https://ghc.haskell.org/trac/ghc/ticket/16057

Anrock:
As a consequnce - hie-8.6.3 can't be build on Windows due to fclabels not being buildable: https://github.com/sebastiaanvisser/fclabels/issues/37

alanz:
Maybe we should adjust the windows build script to skip it by default?

Anrock:
@alanz you mean replace building with some warning in hie-8.6.3 target? Sounds like a good idea.

djleonskennedy:
Hello @alanz will it work on downgraded GHC ?

alanz:
It should work on 8.6.2

djleonskennedy:
@alanz thank you

andersyr:
So that others can avoid the same mistake I made, the [ReadMe](https://github.com/haskell/haskell-ide-engine/blob/master/README.md#install-specific-ghc-version) should use hie-8.6.2 as the **Nightly** installation example and point out that 8.6.3 is currently broken. I.e.,

Install **Nightly** (and hoogle docs):
```
stack ./install.hs hie-8.6.2
stack ./install.hs build-doc-8.6.2
```
(Note, nightly GHC-8.6.3 is currently broken. Please use an alternate version until it's patched.)

alanz:
The nightly is now 8.6.4, and will be used in master when https://github.com/haskell/haskell-ide-engine/pull/1129 lands.

jneira:
Hi, i am getting to install master using stack in windows 10 (with some caveats about cabal-helper-wrapper exe location) but in windows 7 (with stack.yaml and stack-8.6.4.yaml) it fails due to https://github.com/kazu-yamamoto/unix-time/issues/51

Anrock:
`install.hs` no longer tries to build 8.6.3 on Windows, closed by #1168

Issue #1155, opened 2019-03-31
New install.sh builds hie-8.6.3 on Windows... and fails


bscottm:
Might want to either comment it out or make it a platform-dependent addition to the list of versions compiled.

BTW: Would also be handy to have a script that can execute the build, at least on Windows.

fendor:
GHC 8.6.3 does not work on windows, see #1008 

What do you mean with script? It is possible to execute the install.hs with `stack install.hs hie-8.6.4`, for example.

fendor:
You mean, since GHC 8.6.3 does not work on windows, there should be a check/warning for it?

What kind of script? You can execute `install.hs` with `stack install.hs`.

jneira:
Maybe the script (`install.cmd`) could be simply:
```
stack install --stack-yaml=stack-8.6.2.yaml
```
... cause afaik it is the last stack config that works in windows.

In general, it could be a place to put the hacks to install it in windows with the least possible friction for users.

Anrock:
@jneira Pls no additional platform-dependant scripts. We just recently merged a PR that deleted 3 of them, because synchronizing them is a pain and most of active devs don't have windows to test.

Put everything build-related into `install.hs`. Proper fix for this would be an exception for Windows for 8.6.3 target that just bails early with error message, just nobody implemented it.

> ... cause afaik it is the last stack config that works in windows.
Nope, latest is 8.6.4.

jneira:
> @jneira Pls no additional platform-dependant scripts. We just recently merged a PR that deleted 3 of them, because synchronizing them is a pain and most of active devs don't have windows to test.

sorry for being late to the party but the project is being built with appveyor in windows, right? the script could be tested with it automatically, including a smoke test executing hie-wrapper with no error output.

sorry if it is obvious but what were the caveats to sync it?

it would be great to get a balance between user and dev experience 

> Nope, latest is 8.6.4.

In windows 10 it works but it doesnt for me in windows 7 ( see https://github.com/haskell/haskell-ide-engine/issues/1008#issuecomment-479774564)

jneira:
@Anrock never mind, `install.hs` could be the right place to do the windows handling as you can get with it almost the same behaviour of a cmd script

power-fungus:
I am currently in the process of reviewing the install-script. What feature exactly is requested here?

alanz:
[...]

PR #1168, opened 2019-04-09
Build simplifications

power-fungus:
This pull-request is the result of my review of the `install.hs` file.

Following improvements were made:
* hoogle-database is only generated once. The same file was generated every time.
* `cabal-install` is not installed via `stack` if the executable `cabal` can already be found.

A review from @fendor has been conducted and the suggestions have been included.

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -1,7 +1,7 @@
 #!/usr/bin/env stack
 {- stack
-  script
-  --resolver nightly-2018-12-15
+  runghc
+  --stack-yaml=shake.yaml
   --package shake
   --package directory
   --package extra
@@ -147,7 +147,7 @@ main = do
 
 
 buildIcuMacosFix :: VersionNumber -> Action ()
-buildIcuMacosFix version = execStackWithYaml_
+buildIcuMacosFix version = execStackWithGhc_
   version
   [ "build"
   , "text-icu"
@@ -163,15 +163,15 @@ updateSubmodules = do
 
 validateCabalNewInstallIsSupported :: Action ()
 validateCabalNewInstallIsSupported = when isWindowsSystem $ do
-  liftIO $ putStrLn $ embedInStars cabalInstallNotSuported
-  error cabalInstallNotSuported
+  liftIO $ putStrLn $ embedInStars cabalInstallNotSuportedFailMsg
+  error cabalInstallNotSuportedFailMsg
 
 configureCabal :: VersionNumber -> Action ()
 configureCabal versionNumber = do
   ghcPath <- getGhcPath versionNumber >>= \case
     Nothing -> do
-      liftIO $ putStrLn $ embedInStars (ghcVersionNotFound versionNumber)
-      error (ghcVersionNotFound versionNumber)
+      liftIO $ putStrLn $ embedInStars (ghcVersionNotFoundFailMsg versionNumber)
+      error (ghcVersionNotFoundFailMsg versionNumber)
     Just p -> return p
   execCabal_
     ["new-configure", "-w", ghcPath, "--write-ghc-environment-files=never"]
@@ -230,7 +230,7 @@ installCabal = do
         else Nothing
   -- install `cabal-install` if not already installed
   when (isNothing cabalExe) $
-    execStack_ ["install", "--stack-yaml=shake.yaml", "cabal-install"]
+    execStackShake_ ["install", "cabal-install"]
   execCabal_ ["update"]
   ghc <- getStackGhcPathShake
   execCabal_ ["install", "Cabal-2.4.1.0", "--with-compiler=" ++ ghc]
@@ -238,24 +238,24 @@ installCabal = do
 
 checkStack :: Action ()
 checkStack = do
-  Stdout stackVersion <- execStack ["--numeric-version"]
+  Stdout stackVersion <- execStackShake ["--numeric-version"]
   let (parsedVersion, "") : _ =
         stackVersion & trim & readP_to_S parseVersion & filter
           (("" ==) . snd)
   unless (parsedVersion >= makeVersion [1, 9, 3]) $ do
-    liftIO $ putStrLn $ embedInStars stackExeIsOld
-    error stackExeIsOld
+    liftIO $ putStrLn $ embedInStars stackExeIsOldFailMsg
+    error stackExeIsOldFailMsg
 
 
 stackBuildHie :: VersionNumber -> Action ()
 stackBuildHie versionNumber =
-  execStackWithYaml_ versionNumber ["build"]
+  execStackWithGhc_ versionNumber ["build"]
     `actionOnException` liftIO (putStrLn stackBuildFailMsg)
 
 -- | copy the built binaries into the localBinDir
 stackInstallHie :: VersionNumber -> Action ()
 stackInstallHie versionNumber = do
-  execStackWithYaml_ versionNumber ["install"]
+  execStackWithGhc_ versionNumber ["install"]
   localBinDir      <- getLocalBin
   localInstallRoot <- getLocalInstallRoot versionNumber
   let hie = "hie" <.> exe
@@ -266,15 +266,15 @@ stackInstallHie versionNumber = do
 
 buildCopyCompilerTool :: VersionNumber -> Action ()
 buildCopyCompilerTool versionNumber =
-  execStackWithYaml_ versionNumber ["build", "--copy-compiler-tool"]
+  execStackWithGhc_ versionNumber ["build", "--copy-compiler-tool"]
 
 stackTest :: VersionNumber -> Action ()
-stackTest versionNumber = execStackWithYaml_ versionNumber ["test"]
+stackTest versionNumber = execStackWithGhc_ versionNumber ["test"]
 
 stackBuildDoc :: Action ()
 stackBuildDoc = do
-  execStack_ ["--stack-yaml=shake.yaml", "build", "hoogle"]
-  execStack_ ["--stack-yaml=shake.yaml", "exec", "hoogle", "generate"]
+  execStackShake_ ["build", "hoogle"]
+  execStackShake_ ["exec", "hoogle", "generate"]
 
 -- | short help message is printed by default
 shortHelpMessage :: Action ()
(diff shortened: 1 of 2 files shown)


install.hs
```diff
@@ -198,36 +193,30 @@ cabalInstallHie versionNumber = do
...
   configureCabal versionNumber
   execCabal_ ["new-test"]
 
-installCabal :: GhcPath -> Action ()
-installCabal ghc = do
-  execStack_ ["install", "--stack-yaml=shake.yaml", "cabal-install"]
+installCabal :: Action ()
+installCabal = do
+  -- install `cabal-install` if not already installed
+  unlessM (existsExecutable "cabal") $ do
```
fendor: Requires a version check, what if the found cabal version is something like `1.2.0.0`? Do we know whether hie could handle that.?

install.hs
```diff
@@ -198,36 +193,30 @@ cabalInstallHie versionNumber = do
...
+    execStack_ ["install", "--stack-yaml=shake.yaml", "cabal-install"]
```
fendor: Maybe already check that a fitting stack version is installed? We had a bunch of issues that were fixed by upgrading stack.

install.hs
```diff
@@ -198,36 +193,30 @@ cabalInstallHie versionNumber = do
...
   execCabal_ ["update"]
+  ghc <- getStackGhcPath mostRecentHieVersion
```
fendor: This would require a third Stack snapshot, maybe prefer `shake.yaml`  or use the same resolver as the script?

power-fungus: This is indeed an issue. Maybe `Cabal` needs to be installed using all `ghc`-versions? I suppose that it does not really have any benefit to install `Cabal` for a different `ghc` than is actually being used by `hie` (if only one `hie` version is required). 

@alanz do you know if my hypothesis about installing `Cabal` a different `ghc` is true?

install.hs
power-fungus: in there, only the cabal-version can be checked. This has been implemented in a further commit. 

Where should I check the `stack` version?

install.hs
```diff
@@ -167,15 +164,13 @@ configureCabal versionNumber = do
...
-      Just p  -> return $ (version, p) : found
+findInstalledGhcs = mapMaybeM
+  (\version -> do
+    getGhcPath version >>= \case
+      Nothing -> return Nothing
+      Just p  -> return $ Just (version, p)
   )
-  []
-  hieVersions
+  (reverse hieVersions)
```
nponeccop: What about something along the lines of this untested code:
```hs
findInstalledGhcs = reverse <$> mapMaybeM getBothPathAndVersion hieVersions where
    getBothPathAndVersion version = addVersion version <$> getGhcPath version
    addVersion version maybePath = (version,) <$> maybePath
```

It's somewhat convoluted that the two `<$>` are for different functors, but I think "maybe" in `maybePath` clarifies that.

You  can also use `catMaybes` from Data.Maybe from `base` to avoid the dependency

fendor: > What about something along the lines of this untested code:
> 
> ```haskell
> findInstalledGhcs = reverse <$> mapMaybeM getBothPathAndVersion hieVersions where
>     getBothPathAndVersion version = addVersion version <$> getGhcPath version
>     addVersion version maybePath = (version,) <$> maybePath
> ```
> 
> It's somewhat convoluted that the two `<$>` are for different functors, but I think "maybe" in `maybePath` clarifies that.
> 
> You can also use `catMaybes` from `Data.Maybe` from `base` to avoid the dependency

Imo this is kind of hard to read. I would prefer readability over conciseness.

install.hs
power-fungus: Checked in own target

install.hs
power-fungus: now uses the same ghc as the build process. If this actually helps with anything is another question...

install.hs
```diff
@@ -520,3 +548,9 @@ cabalInstallNotSuported =
     ++ "Unfortunately, `cabal new-install` is currently not supported on windows.\n"
     ++ "Please use one of the stack-based targets.\n\n"
     ++ "If this system has been falsely identified, please open an issue at:\n\thttps://github.com/haskell/haskell-ide-engine\n"
+
+-- | Error message when a windows system tries to install HIE via `cabal new-install`
+stackExeIsOld :: String
+stackExeIsOld =
+  "You The `stack` executable is outdated.\n"
```
fendor: Superfluous `You`

install.hs
```diff
@@ -520,3 +548,9 @@ cabalInstallNotSuported =
...
+    ++ "Please run `stack upgrade` to upgrade you stack installation"
```
fendor: `your`

install.hs
fendor: Maybe also add the version found and the version detected.

install.hs
```diff
@@ -4,23 +4,39 @@
   --resolver nightly-2018-12-15
```
fendor: Any decision on whether to use `shake.yaml` instead of `stack script`?

install.hs
```diff
@@ -198,36 +205,54 @@ cabalInstallHie versionNumber = do
...
+  when (isNothing cabalExe) $
+    execStack_ ["install", "--stack-yaml=shake.yaml", "cabal-install"]
   execCabal_ ["update"]
+  ghc <- getStackGhcPathShake
   execCabal_ ["install", "Cabal-2.4.1.0", "--with-compiler=" ++ ghc]
 
+
+checkStack :: Action ()
+checkStack = do
+  Stdout stackVersion <- execStack ["--numeric-version"]
```
fendor: Will this use the nightly compiler?

install.hs
```diff
@@ -449,18 +474,21 @@ getStackGhcPath ghcVersion = do
   Stdout ghc <- execStackWithYaml ghcVersion ["path", "--compiler-exe"]
   return $ trim ghc
 
+getStackGhcPathShake :: Action GhcPath
```
fendor: why the `Shake` suffix? 
Just as a question.

power-fungus: Uses the `shake.yaml` instead of `stack-*.yaml`

install.hs
power-fungus: :man_facepalming: indeed

install.hs
fendor: Right, that's reasonable. Maybe more documentation for new functions while you are at it? :)

install.hs
power-fungus: Will change back

install.hs
```diff
@@ -163,15 +163,15 @@ updateSubmodules = do
 
 validateCabalNewInstallIsSupported :: Action ()
 validateCabalNewInstallIsSupported = when isWindowsSystem $ do
-  liftIO $ putStrLn $ embedInStars cabalInstallNotSuported
-  error cabalInstallNotSuported
+  liftIO $ putStrLn $ embedInStars cabalInstallNotSuportedFailMsg
```
fendor: We have to actually add a TODO here, since in upstream cabal, the PR has been merged to support `new-install` on windows, see [#5837](https://github.com/haskell/cabal/issues/5837)

power-fungus: todo added

install.hs
```diff
@@ -435,33 +435,43 @@ allVersionMessage wordList = case wordList of
...
+  command [] "cabal"
 
 execCabal_ :: [String] -> Action ()
 execCabal_ = command_ [] "cabal"
 
 existsExecutable :: MonadIO m => String -> m Bool
 existsExecutable executable = liftIO $ isJust <$> findExecutable executable
 
+
+-- QUERY ENVIRONMENT
```
fendor: This could be more prominent, otherwise it can easily be overlooked.

install.hs
```diff
@@ -534,23 +544,23 @@ stackBuildFailMsg =
...
-cabalInstallNotSuported :: String
-cabalInstallNotSuported =
+cabalInstallNotSuportedFailMsg :: String
+cabalInstallNotSuportedFailMsg =
   "This system has been identified as a windows system.\n"
     ++ "Unfortunately, `cabal new-install` is currently not supported on windows.\n"
     ++ "Please use one of the stack-based targets.\n\n"
     ++ "If this system has been falsely identified, please open an issue at:\n\thttps://github.com/haskell/haskell-ide-engine\n"
 
 -- | Error message when a windows system tries to install HIE via `cabal new-install`
```
fendor: Wrong documentation

install.hs
power-fungus: Suggestions?

install.hs
power-fungus: What is the problem?

install.hs
fendor:
```haskell
-- ------------------------------------------------------------
-- * query environment
-- ------------------------------------------------------------
```
Have seen this in the cabal code base.
However, this is highly subjective suggestion, no need to do it.

install.hs
fendor: Documentation is duplicated for the function `stackExeIsOldFailMsg`.

install.hs
```diff
@@ -561,7 +561,8 @@ cabalInstallNotSuportedFailMsg =
     ++ "If this system has been falsely identified, please open an issue at:\n\thttps://github.com/haskell/haskell-ide-engine\n"
 
 -- | Error message when the `stack` binary is an older version
-stackExeIsOldFailMsg :: String
-stackExeIsOldFailMsg =
+stackExeIsOldFailMsg :: String -> String
+stackExeIsOldFailMsg stackVersion =
   "The `stack` executable is outdated.\n"
+    ++ "found version is `" ++ stackVersion ++ "`.\n"
```
fendor: How about the version that is required?

alanz:
I just checked out this branch, deleted my `.stack-work` and gave it `./install.hs build-all`.

It first built and ran hoogle, and then started building `Cabal-2.4.1.0`, which was not built by the prior version, and is already on my machine, as far as I know. 

It then started building `cabal-install-2.0.0.0`, which as far as I am concerned should *not* be installed, as it is an old version. This is for the `hie-8.2.1` target, I think.

Stopping the build and restarting it does not re-install Cabal or cabal-install, so I guess they are used for the specific hoogle in use. Which should probably be build against the `hie-8.6.4` baseline, rather than the `hie-8.2.1` one.

power-fungus:
@alanz Thank you for testing my PR.

`cabal-install` and `hoogle` are runtime-dependencies that need to be built if not already present on the system. To minimize the number of downloaded `ghc`s, they are built using the same ghc as the `install.hs` file is beeing executed. It should not build `cabal-install` or `hoogle` if the are already present in the path.

I am unable to reproduce the behavior that `cabal-install-2.0.0.0` is installed. For me, it is the version as provided in the resolver from the `shake.yaml`, which is `cabal-install-2.4.1.0`.

power-fungus:
Following additional features are implemented: 
* building with `8.6.3` is no option windows systems.
* check if installed `stack` version is too old
* install `cabal-install` via `stack` if the found `cabal` executable is too old

fendor:
LGTM, ready to merge in my opinion, I tested it on windows.

alanz:
If you guys are happy with this, then merge it.

Anrock:
I see it also closes #1008 and #1155

fendor:
In the short help message, the nightly ghc version seems to be excluded. If this is fine, then we can merge, otherwise this needs to be added to the short-help message.

install.hs
```diff
@@ -241,87 +280,84 @@ stackInstallHie versionNumber = do
...
+  hieVersions <- getHieVersions
   let out = liftIO . putStrLn
   scriptName <- liftIO getProgName
   out ""
   out "Usage:"
   out' ("stack " <> scriptName <> " <target>")
   out ""
   out "Targets:"
-  mapM_ (out' . showTarget spaces) targets
+  mapM_ (out' . showTarget (spaces hieVersions)) (targets hieVersions)
```
fendor: This should show the nightly ghc version as well. That's my last nitpick, afterwards we can merge.

PR #1174, opened 2019-04-15
Avoid crash in case of nonsensical hoogle db

fendor:
Since the hoogle db location can be defined by the user, it may be possible that the database file location is wrong. 
If the file does not exist, an error is reported to the user, but if the file exists but it is just not a hoogle db, hoogle plugin crashes, taking hie with it.

This catches the error and wraps into a `HoggleError` that can be shown to the user, e.g. to tell him that the db is garbage.

Issue related to: #1146.

* ~[ ] Actually show an error to the user.~ Not adressed in this issue.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/Hoogle.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Hoogle.hs
@@ -3,6 +3,8 @@
 module Haskell.Ide.Engine.Plugin.Hoogle where
 
 import           Control.Monad.IO.Class
+import           Control.Monad (join)
+import           Control.Exception
 import           Data.Aeson
 import           Data.Bifunctor
 import           Data.Maybe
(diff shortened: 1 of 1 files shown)


fendor:
It is actually unclear when to show a message to the user.

The main usage of Hoogle is to lookup a specifc function name, or used for HsImport. All of these use cases are Commands, executed by CodeActions. It is impossible by design to send messages from code actions. However, we could newtype results from Hoogle and unwrap them in LspStdio to show messages to the user.

How should a message be propagated to the user? It is definitely something we want to show to the user, since it is a configuration error.

lorenzo:
Could we do a initial health check in HIE startup and send messages to the users on the problems we found? 

If any fatal problem is found we could then disable the plugin or just let the stuff silently fail.

This would require some sort of callback function in plugins to perform this check

fendor:
A health check sounds great to me, it would further help to identify problems with plugins on start up. Moreover, to signal the health of hie in general. 
It could be a required field in plugin description

power-fungus:
I am thinking about this for a long time now and I am in the process of gathering things that could go wrong. However, a lot of components are currently changing dramatically so this has been postponed (hie-bios, cabal-helper-wrapper, hlint-stuff, ...).

In the next week, I will start implementing the first parts of the checking system.

fendor:
Then the issue of showing user this error is deferred and this is ready to merge?

PR #1184, opened 2019-04-20
remove installation of Cabal by cabal

power-fungus:
The `install.hs` calls `cabal install Cabal-X`. This command was copied from the `Makefile` several months ago.

I have done exhaustive tests on different machies (linux only) and have not seen any effect of this line and `hie` works even without this command. This PR removes this command from the installation process.

Comments on this change and tests on MacOS and Windows are welcome!

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -248,8 +248,6 @@ installCabal = do
   when (isNothing cabalExe) $
     execStackShake_ ["install", "cabal-install"]
   execCabal_ ["update"]
-  ghc <- getStackGhcPathShake
-  execCabal_ ["install", "Cabal-2.4.1.0", "--with-compiler=" ++ ghc]
 
 
 checkStack :: Action ()


Anrock:
I'll give it a shot on Windows soon. What should i remove to be sure it's working properly?

fendor:
Did you measure the installation time? I think, one point was to accelerate the installation.

power-fungus:
@Anrock on Linux, `hie` clearing means roughly:
* remove installed binaries from `$HOME/.local/bin`: `hie`, `hie-*`, `cabal`, `hoogle`
* remove all the caches: `$HOME/.stack`, `$HOME/.cabal`, `$HOME/.hoogle`

Installing:
* freshly clone the repository
* `stack install.hs build-all`
* let run overnight

power-fungus:
@fendor It was used to reduce the time during the first startup of `hie`. However I have noticed no speedup and the first startup was quite quick.

fendor:
Alright. Then LGTM.

alanz:
I'm happy to merge this, if anyone complains we will know to add it back, or put in instructions to do it.

Anrock:
Late to the party, but still - most of hies build just fine with build-all.

alanz:
So, I found my cabal-helper was not able to work with hie on hie itself anymore, but was reporting that it was building against `Cabal-2.4.0.1`.

After reverting this commit and re-installing hie, it works again.

So, the reason we did this originally, was to allow cabal-helper to actually work.  We need to revert this.

power-fungus:
I can do that. However, I have been unable to reproduce that behavior. A clean installation of `hie` does not make `cabal-helper` require `Cabal` on my test-vms, which provide the maximum repeatability and isolation I can provide. 

@alanz Can you give me instructions on how to reproduce this error? I will go back to my test-environment and figure out what is going on.

alanz:
It comes about trying to load hie itself, and shows cabal-helper complaining to stderr that the exe has two mains in it, one of which is `Paths_haskell_ide_engine`.

So our current tests do not show this as a problem, and perhaps we should add one for this specific case.

alanz:
And now I am no longer sure about this. I installed hie using this approach, and it now works where it did not before.  But the cabal-helper cache has it's exe called `cabal-helper0.9.0.0-Cabal2.4.0.1`.

alanz:
So perhaps we should wait for any issues to be raised, and consider this as an option if so.

power-fungus:
@alanz, I am also very unsure about this, In my tests in an non-clean environment, I got different results every time and were unable to reproduce any issues reliably.

Could you clarify in what state would you prefer to wait. Should we include `cabal v1-install Cabal-X` in the meantime?

I have a repository that describes a vagrant-vm I use to test the `hie` binaries and the required environment. I can work this a bit further and publish it to the collaborators here.

alanz:
I think we should keep our fix in reserve, and see if anyone raises an issue against `0.9.0.0` in the meantime.

PR #1185, opened 2019-04-21
Change 'build-docs' to 'build-doc' in README

ajeetdsouza:
In https://github.com/haskell/haskell-ide-engine/commit/8fd5cba63213736e8720799a57afab32c8ef544b, the `build-docs` flag for install.hs was changed to `build-doc`, but was not updated in the README. I have added this change.


fendor:
Thanks for catching that!

PR #1126, opened 2019-03-10
Implement the HIE Bios

mpickering:
This branch implements the HIE bios which is a simplified way of setting up a GHC session.

The library, `hie-bios`, is responsible for one thing and only one thing. That is to set up the correct GHC session. 

The guiding principle of `hie-bios` is:

> It is the responsibility of the build tool to describe the environment which a package should be built in.

This means that it is easy to implement support for any build tool.  I already implemented support for stack/cabal/hadrian/rules_haskell/obelisk. These are the main build tools I think. More information can be read in the [README](https://github.com/mpickering/haskell-ide-engine/tree/hie-bios/hie-bios#hie-bios)

I need help in finishing this PR which is why I am putting it up here.

* [x] HaRe support is broken at the moment as it relies on `ghc-mod`. In theory it could still use `ghc-mod` I think but I didn't give it much thought yet.
* [x] There are still many references to functions from `ghc-mod` which are not `ghc-mod` specific. They need to be moved into `haskell-ide-engine` or upstreamed.
* [x] Multi-project support is not implemented. The whole premise of how this works seemed quite fragile to me. Is it really enough to set a different set of `DynFlags` when changing between sessions? This is not something the GHC API is really designed for, I would think you would have to make a new GHC session for each component or spawn another process. If the approach of replacing DynFlags really does work then we just need to create a map from arguments to `DynFlags`. This also deals with creating a new session if the project configuration changes quite naturally as the cache will miss and a new session created. 
* [x] It all needs testing by someone who knows how HIE works now as I don't use it (yet). The whole reason for this PR was to get HIE working on GHC's codebase.

PR #1186, opened 2019-04-21
Typemap reimplementation

### LINKED ISSUE #1156: Hover causes far too much memory
The way that the types are collected from the AST by using syb uses a lot of memory. This was already identified when haddock started using HIE files for its hyperlinker. The solution is to 
make the traversal smarter to only recover types at the leaves rather than for all subexpressions.

This issues makes HIE unusable with GHC (or any big project) as the process of traversing a single module is too expensive. 

cc @wz1000

fendor:
Under the orchestration of @mpickering, reimplements the `genTypeMap` function in a new module.
A Typechecked module is only traveresed once.
This reimplementation should have a lower memory footprint but is likely to increase the first startup time of modules.
Some parts may no longer have a type on hover.

Implementation has already been merged into #1126. There it is said to be working!

* [x] More Documentation
* [x] Way more tests

Closes #1156

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Compat.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Compat.hs
@@ -7,6 +7,7 @@ import qualified GHC
 import qualified Type
 import qualified TcHsSyn
 import qualified TysWiredIn
+import qualified Var
 
 #if MIN_VERSION_filepath(1,4,2)
 #else
@@ -122,6 +123,17 @@ pattern HsMultiIfType t <-
 #else
     GHC.HsMultiIf t _
 #endif
+
+pattern FunBindType :: Type.Type -> GHC.HsBindLR GhcTc GhcTc
+pattern FunBindType t <-
+#if MIN_VERSION_ghc(8, 6, 0)
+    GHC.FunBind _ (GHC.L _ (Var.varType -> t)) _ _ _
+#elif MIN_VERSION_ghc(8, 4, 0)
+    GHC.FunBind (GHC.L _ (Var.varType -> t)) _ _ _ _
+#else
+    GHC.FunBind (GHC.L _ (Var.varType -> t)) _ _ _ _
+#endif
+
 
 #if MIN_VERSION_ghc(8, 6, 0)
 matchGroupType :: GHC.MatchGroupTc -> GHC.Type
--- a/hie-plugin-api/Haskell/Ide/Engine/Compat.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Compat.hs
@@ -3,7 +3,6 @@
 {-# LANGUAGE ViewPatterns #-}
 module Haskell.Ide.Engine.Compat where
 
-
 import qualified GHC
 import qualified Type
 import qualified TcHsSyn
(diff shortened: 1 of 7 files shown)


fendor:
@mpickering Anything missing for this PR? Or do you think we should rather wait for your PR to be merged?

hie-plugin-api/Haskell/Ide/Engine/TypeMap.hs
```diff
@@ -0,0 +1,165 @@
...
+      (\e acc -> do
+        e' <- e
+        a  <- acc
+        return (e' `IM.union` a)
+      )
+      v
+      (everythingButM f)
+      x
+
+-- | This instance tries to construct 'HieAST' nodes which include the type of
```
mpickering: Need to update this comment.

hie-plugin-api/Haskell/Ide/Engine/TypeMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- the node. Our next recourse would be to desugar it into a 'CoreExpr' then
+-- query the type of that. Yet both the desugaring call and the type query both
+-- involve recursive calls to the function and argument! This is particularly
+-- problematic when you realize that the HIE traversal will eventually visit
+-- those nodes too and ask for their types again.
+--
+-- Since the above is quite costly, we just skip cases where computing the
+-- expression's type is going to be expensive.
+--
+-- See #16233
```
mpickering: Add a reference here that it's GHC #16233 as that isn't obvious.

hie-plugin-api/Haskell/Ide/Engine/TypeMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- Intervals may overlap and return more specific results.
+genTypeMap :: GHC.GhcMonad m => TypecheckedModule -> m TypeMap
+genTypeMap tm = do
+  let typecheckedSource = GHC.tm_typechecked_source tm
+  hs_env <- GHC.getSession
+  liftIO $ types hs_env typecheckedSource
+
+collectAllSpansTypes'
+  :: GHC.GhcMonad m => Bool -> TypecheckedModule -> m [(GHC.SrcSpan, GHC.Type)]
+collectAllSpansTypes' = collectAllSpansTypes
```
mpickering: What is the point of this function?

fendor: This is a leftover, thanks for catching it, but it is probably in your codebase now as well.

fendor:
> 
> 
> You should quickly check the `ghc-mod` source to see if there are any other hover points it supports so this doesn't regress behaviour?
Will do, thanks!
> 
>
> Also, are there any tests for hover? It would be good to add some if there aren't but it shouldn't block merging this MR.
I think there are some tests, but I will investigate

hie-plugin-api/Haskell/Ide/Engine/TypeMap.hs
```diff
@@ -0,0 +1,145 @@
...
+      HsLitType t -> Just t
+      HsLamType t -> Just t
+      HsLamCaseType t -> Just t
+      HsCaseType t -> Just t
+      ExplicitListType t -> Just t
+      ExplicitSumType t -> Just t
+      HsMultiIfType t -> Just t
+
+      _                        -> Nothing
+  in  case tyOpt of
```
mpickering: Interesting this isn't pattern matched on.. perhaps this is why you were so confused.

fendor: It was certainly an additional barrier for me! 
However, since we established that this is a bug, it is actually more clear for me!

hie-plugin-api/Haskell/Ide/Engine/Compat.hs
```diff
@@ -27,3 +34,97 @@ isExtensionOf :: String -> FilePath -> Bool
...
+type GhcTc = GHC.GhcTc
+#else
+type GhcTc = GHC.Id
+#endif
+
+pattern HsOverLitType :: Type.Type -> GHC.HsExpr GhcTc
+pattern HsOverLitType t <-
+#if MIN_VERSION_ghc(8, 6, 0)
+    GHC.HsOverLit _ (GHC.overLitType -> t)
+#elif MIN_VERSION_ghc(8, 4, 0)
```
mpickering: Can just have the `else` clause here I think.

fendor: Yeah, I wanted to ask that anyway, do we prefer easy to read, e.g. always three clauses although they often completely overlap, or conciseness when it comes to CPP?

mpickering: I prefer not having the duplication and just having two clauses when possible.

fendor:
When the tests run through, this should be ready to merge, imo.

alanz:
Given what a big change this is, I suggest we merge it after the April release, which is due in a few days.

mpickering:
CI looks to be failing still on 8.2.2 and 8.2.1. 

@alanz Are you sure you don't want to merge this? It improves one of the big complaints about HIE which is high memory usage.

fendor:
@mpickering  Tests fail because the API of Show and Eq seem to have changed over different ghc versions?
However, this has no effect on the feature. 
Should the test case be solved via CPP?

fendor:
CI fails are imo unrelated to this PR.

mpickering:
This can be merged now I think?

alanz:
I think so, I presume the CI is transient.

alanz:
In other words, I guess I should actually take a look at it now.

PR #1189, opened 2019-04-22
Make sure the end of formatted text is properly indicated

### LINKED ISSUE #1188: New documentation fails for `bimap`
We end up with

Screenshot from 2019-04-22 10-32-16

Which is because of

    ```haskell
    `bimap` f g ≡ `first` f `.` `second` g```#### **Examples**

i.e. the closing three backticks should be on a newline, as should the start of the `Examples` header.

alanz:
Closes #1188


Hogeyama:
Hi! The changes on [this line](https://github.com/haskell/haskell-ide-engine/pull/1189/files#diff-642a9a925b719fa6c17c13b53efec7a3R196) and [this line](https://github.com/haskell/haskell-ide-engine/pull/1189/files#diff-642a9a925b719fa6c17c13b53efec7a3R191) are unnecessary because newlines are surely inserted by `T.unlines`, aren't they? It now outputs a redundant empty line like this:

````markdown
```haskell
> const 42 "hello"
42

```
````

alanz:
@Hogeyama you are correct, it is only actually `markupCodeBlock` that was faulty. I will revert that part.

PR #1190, opened 2019-04-22
Add window/progress reporting for typechecking

bubba:
# Todo

- [X] Add test
- [X] haskell-lsp[-types] upload
- [ ] Update vscode-hie-server

Depends on a couple of tweaks with the VS code extension

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
@@ -50,6 +50,7 @@ module Haskell.Ide.Engine.PluginsIdeMonads
   , runIdeM
   , IdeDeferM
   , MonadIde(..)
+  , Core.Progress(..)
   , iterT
   , LiftsToGhc(..)
   -- * IdeResult
@@ -350,6 +351,12 @@ class MonadIO m => MonadIde m where
   getConfig :: m Config
   getClientCapabilities :: m ClientCapabilities
   getPlugins :: m IdePlugins
+  -- 'withProgress' @title f@ wraps a progress reporting session for long running tasks.
+  -- f is passed a reporting function that can be used to give updates on the progress
+  -- of the task.
+  withProgress :: T.Text -> ((Core.Progress -> m ()) -> m a) -> m a
+  -- 'withIndefiniteProgress' @title f@ is the same as the 'withProgress' but for tasks
+  -- which do not continuously report their progress. 
   withIndefiniteProgress :: T.Text -> m a -> m a
 
 instance MonadIO m => MonadIde (ReaderT IdeEnv m) where
@@ -379,6 +386,10 @@ instance MonadIO m => MonadIde (ReaderT IdeEnv m) where
 
   getPlugins = asks idePlugins
 
+  withProgress t f = do
+    lf <- asks ideEnvLspFuncs
+    withProgress' lf t f
+
   withIndefiniteProgress t f = do
     lf <- asks ideEnvLspFuncs
     withIndefiniteProgress' lf t f
@@ -389,6 +400,9 @@ instance MonadIde IdeGhcM where
   getConfig = lift $ lift getConfig
   getClientCapabilities = lift $ lift getClientCapabilities
   getPlugins = lift $ lift getPlugins
+  withProgress t f = do
+    lf <- lift $ lift $ asks ideEnvLspFuncs
+    withProgress' lf t f
   withIndefiniteProgress t f = do
     lf <- lift $ lift $ asks ideEnvLspFuncs
     withIndefiniteProgress' lf t f
@@ -399,10 +413,20 @@ instance MonadIde IdeDeferM where
   getConfig = lift getConfig
   getClientCapabilities = lift getClientCapabilities
   getPlugins = lift getPlugins
+  withProgress t f = do
+    lf <- lift $ asks ideEnvLspFuncs
+    withProgress' lf t f
   withIndefiniteProgress t f = do
     lf <- lift $ asks ideEnvLspFuncs
     withIndefiniteProgress' lf t f
 
+withProgress' :: MonadIO m => Maybe (Core.LspFuncs Config) -> T.Text -> ((Core.Progress -> m ()) -> m a) -> m a
+withProgress' lspFuncs t f =
+  let mWp = Core.withProgress <$> lspFuncs
+    in case mWp of
+        Nothing -> f (const $ return ())
+        Just wp -> wp t f
+    
 withIndefiniteProgress' :: MonadIO m => Maybe (Core.LspFuncs Config) -> T.Text -> m a -> m a
 withIndefiniteProgress' lspFuncs t f =
   let mWp = Core.withIndefiniteProgress <$> lspFuncs
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
@@ -49,8 +49,17 @@ module Haskell.Ide.Engine.PluginsIdeMonads
   , IdeM
   , runIdeM
   , IdeDeferM
-  , MonadIde(..)
+  -- ** MonadIde and functions
+  , MonadIde
+  , getRootPath
+  , getVirtualFile
+  , getConfig
+  , getClientCapabilities
+  , getPlugins
+  , withProgress
+  , withIndefiniteProgress
   , Core.Progress(..)
+  -- ** Lifting
   , iterT
   , LiftsToGhc(..)
   -- * IdeResult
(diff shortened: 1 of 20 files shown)


hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -378,19 +379,36 @@ instance MonadIde IdeM where
...
   getRootPath = lift getRootPath
   getVirtualFile = lift . getVirtualFile
   getConfig = lift getConfig
   getClientCapabilities = lift getClientCapabilities
   getPlugins = lift getPlugins
+  withIndefiniteProgress t f = do
+    lf <- lift $ asks ideEnvLspFuncs
+    withIndefiniteProgress' lf t f
+
+withIndefiniteProgress' :: MonadIO m => Maybe (Core.LspFuncs Config) -> T.Text -> m a -> m a
```
bubba: I couldn't get this to type check with `withIndefiniteProgress = lift . withIndefiniteProgress`, this is a workaround. Open to suggestions to make this nicer

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -344,14 +344,15 @@ data IdeEnv = IdeEnv
   }
 
 -- | The class of monads that support common IDE functions, namely IdeM/IdeGhcM/IdeDeferM
-class Monad m => MonadIde m where
+class MonadIO m => MonadIde m where
```
alanz: I guess IO is needed to actually send the message back?

alanz:
What changes are needed for the vscode extension? How portable is this to other clients?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
bubba: Yes

bubba:
> What changes are needed for the vscode extension? How portable is this to other clients?

Mainly updating the client library and enabling the proposed features, https://github.com/alanz/vscode-hie-server/pull/150
Since there is a `ClientCapabilities` for the `window/progress` notifications, haskell-lsp won't send them unless the client declares it so it should be backwards compatible with other clients

bubba:
> Besides tests, I think it would be nice to have examples on how to integrate them into a plugin. E.g. some plugins might want to be able to report progress as well. Will that be possible?
> Otherwise, great job, the diff is clean and easy to follow!

Yup! Since `withIndefiniteProgress` is inside `IdeM`, any plugin can just wrap their computations inside it to start a progress session. Are there any long running plugins at the moment we can add this to?

alanz:
> Are there any long running plugins at the moment we can add this to?

Possibly the liquid haskell one

bubba:
@alanz @fendor an example:
https://github.com/haskell/haskell-ide-engine/blob/0bdb86a00146deb140402e2bc3bea41c5e835dd2/src/Haskell/Ide/Engine/Plugin/Liquid.hs#L123-L125

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -378,19 +386,53 @@ instance MonadIde IdeM where
...
+  withProgress t f = do
+    lf <- lift $ asks ideEnvLspFuncs
+    withProgress' lf t f
+  withIndefiniteProgress t f = do
+    lf <- lift $ asks ideEnvLspFuncs
+    withIndefiniteProgress' lf t f
+
+withProgress' :: MonadIO m => Maybe (Core.LspFuncs Config) -> T.Text -> ((Core.Progress -> m ()) -> m a) -> m a
+withProgress' lspFuncs t f =
+  let mWp = Core.withProgress <$> lspFuncs
```
lorenzo: wouldn't it be better to pass the actual `Core.withProgress` instead of getting it here?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -344,14 +345,21 @@ data IdeEnv = IdeEnv
...
   getVirtualFile :: Uri -> m (Maybe VirtualFile)
   getConfig :: m Config
   getClientCapabilities :: m ClientCapabilities
   getPlugins :: m IdePlugins
-
-instance MonadIde IdeM where
+  -- 'withProgress' @title f@ wraps a progress reporting session for long running tasks.
+  -- f is passed a reporting function that can be used to give updates on the progress
+  -- of the task.
+  withProgress :: T.Text -> ((Core.Progress -> m ()) -> m a) -> m a
```
lorenzo: It would be ore clear to name this `withProgressReport`

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
alanz: And the passed in function could be of the form `a -> IO ()`, so obviating the need for `MonadIO` too.

bubba: @Lorenzo currently there is the possibility that lspFuncs returns null, I think this is due to the existence of the JSON transport. I think this also asks the question, should we just give plugins full access to lsp funcs?

alanz: Either give full access, or make the progress reporting function available to them by default, as in my earlier comment. I think we already do something similar with accessibility to the VFS stuff. I would have to take a look though.

I am very much in favour of constraining what plugins/monads can do, so we do not get weird stuff happening

bubba: I've reworked the class in https://github.com/haskell/haskell-ide-engine/pull/1190/commits/8d1925dcfc8534d560fac5752ff96ac85dfce17a, so only the functions that require `MonadIO` have the constraint. @Lorenzo is it ok if once JSONStdio is gone and there is the guarantee of `LspFuncs` (#1105) I can revisit this to directly pass it instead?

mpickering:
One thing we noticed on IRC is that the type of `withProgress` needs to be modified so the continuation is of type `Progress -> IO ()`. Otherwise you can't use these functions to pass the callback into GHC for example.

bubba:
@mpickering see https://github.com/alanz/haskell-lsp/pull/154

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
lorenzo: Sure, that makes sense

src/Haskell/Ide/Engine/Plugin/Liquid.hs
```diff
@@ -120,7 +120,9 @@ diagnosticProvider DiagnosticOnSave uri cb = pluginGetFile "Liquid.diagnosticPro
     LiquidData mtid <- get
     mapM_ (liftIO . cancel) mtid
 
-    tid <- liftIO $ async $ generateDiagnosics cb uri file
+    let progTitle = "Running Liquid Haskell on " <> T.pack (takeFileName file)
+    tid <- withIndefiniteProgress progTitle $
+      liftIO $ async $ generateDiagnosics cb uri file
```
mpickering: This isn't right. The progress report needs to go inside the call to `async` and if you do that, you'll discover that the types for `withIndefiniteProgress` are not quite right still..

mpickering:
I don't think this is quite right yet. It's weird that the `haskell-lsp` functions are all parameterised over `MonadIO` but the other ones in `haskell-lsp` aren't. Once you fix those types to IO then you can provide a polymorphic interface in `hie-plugin-api` by using `UnliftIO`.

mpickering:
See https://github.com/mpickering/haskell-ide-engine/commit/3b41a98adc03fb852dd00e2df9388616e8959aea#diff-5c94e6cdb53262cd6976feddb9d4cedfR126

and https://github.com/mpickering/haskell-ide-engine/commit/ed9fefa1304ed28272422d9d0684b1487b09e597

For how I think this should be fixed.

src/Haskell/Ide/Engine/Plugin/Liquid.hs
bubba: You’re right, this completely slipped by me! I was testing it by adding a `threadDelay 10000000`…

bubba:
> I don't think this is quite right yet. It's weird that the `haskell-lsp` functions are all parameterised over `MonadIO` but the other ones in `haskell-lsp` aren't. Once you fix those types to IO then you can provide a polymorphic interface in `hie-plugin-api` by using `UnliftIO`.

You're right: I got ahead of myself and was thinking of `withProgress` being used in the context of `IdeM` etc., I agree that `haskell-lsp` functions should remain in `IO`.
I'll update this PR to cherry-pick in your changes, and I imagine the `MonadUnliftIO` instance can be used in lots of other places in HIE

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -337,6 +342,12 @@ runIdeGhcM ghcModOptions plugins mlf stateVar f = do
 data Defer a = Defer FilePath (UriCacheResult -> a) deriving Functor
 type IdeDeferM = FreeT Defer IdeM
 
+{-
+data IdeDeferM a = Defer FilePath (UriCacheResult -> IdeDeferM a)
+                 | IdeLeaf (IdeM a)
+                 deriving Functor
+                 -}
+
```
mpickering: Can delete this as well.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -400,23 +411,27 @@ getPlugins = idePlugins <$> getIdeEnv
 -- | 'withProgress' @title f@ wraps a progress reporting session for long running tasks.
 -- f is passed a reporting function that can be used to give updates on the progress
 -- of the task.
-withProgress :: (MonadIde m, MonadIO m) => T.Text -> ((Core.Progress -> m ()) -> m a) -> m a
-withProgress t f = do
+withProgress :: (MonadIde m , MonadIO m, MonadBaseControl IO m)
+             => T.Text -> Core.ProgressCancellable
+             -> ((Core.Progress -> m ()) -> m a) -> m a
```
mpickering: I still think you want the argument to the continuation to be in IO. The other arguments should be generalised though.

src/Haskell/Ide/Engine/Plugin/Liquid.hs
```diff
@@ -120,9 +120,10 @@ diagnosticProvider DiagnosticOnSave uri cb = pluginGetFile "Liquid.diagnosticPro
     LiquidData mtid <- get
     mapM_ (liftIO . cancel) mtid
 
-    let progTitle = "Running Liquid Haskell on " <> T.pack (takeFileName file)
-    tid <- withIndefiniteProgress progTitle $
-      liftIO $ async $ generateDiagnosics cb uri file
+    -- let progTitle = "Running Liquid Haskell on " <> T.pack (takeFileName file)
+  --  tid <- async $ withIndefiniteProgress progTitle $ (liftIO $ generateDiagnosics cb uri file)
```
mpickering: What's going on here?

bubba: Should be resolved now, needed to make `withIndefiniteProgressIO`

mpickering:
You should use the `lifted-async` package to fix the liquid haskell progress report rather than define the `*IO` variant.

lukelaumovi:
@mpickering perfect, that package is just what I'm looking for. Will update

bubba:
@mpickering `lifted-async` worked a charm!

alanz:
@bubba there are now conflicts here

bubba:
Will this be merged in time for #1227 ?

alanz:
Probably not. I need to land the release  tonight, as I am traveling the
next two days,  so can't predict time to pay attention.

alanz:
@bubba @mpickering Can you guys refresh this so we can merge?

PR #1194, opened 2019-04-23
Fix some of the tests with cabal new-build

michaelpj:
This doesn't fix them all, but it gets us some way.

I'll make issues for the other things I've found.


bubba:
The cabal test is failing because of #1195, we can ignore it

PR #1199, opened 2019-04-23
Undo unnecessary changes from #1189

alanz:
As per @Hogeyama  comment at
https://github.com/haskell/haskell-ide-engine/pull/1189#issuecomment-485690789


alanz:
And I just rebuilt with current master, and the problem is back.  I guess we need a decent test for it, but I am loathe to tie it to what a particular haddock library provided at a particular time.

alanz:
But reverting this diff makes no difference. I guess I did not fix it properly the first time.

Issue #1179, opened 2019-04-18
Explain how to install with cabal in README


ivanperez-keera:
The README favours installation methods other than cabal. For reasons that at too long to discuss here, some of us do not like using stack.

While the readme states that cabal can be used to install haskell-ide-engine, this is only briefly mentioned, and no explicit instructions are given.

As it turns out, this package compiles just fine with the good-old cabal sandboxes, which are a perfectly good way of keeping things isolated.

Either with v2 or with v1, it would be great if the README contained specific instructions of how to install this with cabal.

mpickering:
@ivanperez-keera I think the main reason for the complexity is that the installation script builds 3 different versions of hie, one for each of the last 3 GHC releases. 

There is also the problem of the large number of submodules which are because certain dependencies have
not made it onto hackage in months/years. We are in the process of trying to decouple from these dependencies so hopefully some progress in this regard by the end of the summer. 

But you are right that `cabal new-build` works fine to build one of the executables and this is how I do local development.

Anrock:
@mpickering could you make a PR and document howto use cabal new-build in README?

power-fungus:
I can write some documentation about this. However, this is not a good idea. Before we had install.hs we had new issues daily about some error that Was due to an incorrect installation. Officially supporting a different installation method will bring these back.

fendor:
The install.hs almost supports build with cabal
especially with the incoming build simplifications.
To make it possible, you need to compile `install.hs` somehow. Ghc needs to know about the packages `shake`, `directory`, and soon `extra`.
The location to install hie to is defined by `stack --local-bin`.
you may replace, for yourself, `stackLocalBin` with the location that you want hie to be in.
then it should be possible to use `./install cabal-build` to install only with cabal.

mpickering:
I'm just talking about the fact you can run `cabal new-build` to build the project if you just want one executable. I'm not say the installation script should be used to use cabal.

power-fungus:
as far as I know it the steps to build `hie` (directly after a fresh clone) are the following:

1. `git submodule sync && git submodule update --init`
2. somehow get `hoogle` (`>=5.0.17` to be safe) into the path and call `hoogle generate`
3. build `hie` with the build-system of your choice
4. install `hie` and the `hie-wrapper` into the path
5. repeat steps 3 and 4 of all required ghc-versions

Historically, a few more steps are included, (of which I am unsure if really needed)
1. call `cabal install Cabal-X`
[...]

PR #1200, opened 2019-04-24
expand documentation on the build system

power-fungus:
This PR updates the documentation of the build system based on the discussion in issue #1179. It documents the features of the `install.hs` script and the tradeoffs that have been made. Also, the changes of #1168 and #1184 have been documented.

Proof-reading and suggestions are welcome.

### DIFF
--- a/docs/Build.md
+++ b/docs/Build.md
@@ -10,41 +10,67 @@ The design of the build system has the following main goals:
 
 * works identically on every platform
 * has minimal run-time dependencies:
-    - a working installation of `stack`
+    - `stack`
     - `git`
-* is completely functional right after simple `git clone`
-* one-stop-shop for building all executables required for using `hie` in IDEs.
+* is completely functional right after a simple `git clone` and after every `git pull`
+* one-stop-shop for building and naming all executables required for using `hie` in IDEs.
+* fails with meaningful error messages in known circumstances
+* is able to modify the environment such that `hie` can be run
+    - setup `hoogle` database
+    - setup `hlint` data-files
 
 See the project's `README` for detailed information about installing `hie`.
 
+### Targets
+
+The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
+
+* `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
+* `build`: builds and installs `hie` binaries for all supported `ghc` versions.
+* `build-doc`: builds the hoogle-db required by `hie`
+* `cabal-*`: execute the same task as the original target, but with `cabal` instead of `stack`
+
+Each `stack-*.yaml` contains references to packages in the submodules. Calling `stack` with one of those causes the build to fail if the submodules have not been initialized already. The file `shake.yaml` solves this issue. Moreover, it specifies the correct version of `shake` and is used for installing all run-time dependencies such as `cabal` and `hoogle` if necessary.
+
+### Run-time dependencies
+
+`hie` depends on a correct environment in order to function properly:
+
+* `cabal-install`: If no `cabal` executable can be found or has an outdated version, `cabal-install` is installed via `stack`.
+* The `hoogle` database: `hoogle generate` needs to be called with the most-recent `hoogle` version.
+
+### Steps to build `hie`
+
+Installing `hie` is a multi-step process:
+
+1. `git submodule sync && git submodule update --init`
+2. `hoogle generate` (`hoogle>=5.0.17` to be safe)
+3. ensure that `cabal-install` is installed in the correct version
+4. `stack --stack-yaml=stack-<X>.yaml install` or `cabal new-install -w ghc-<X>`
+5. rename `hie` binary to `hie-<X>` in `$HOME/.local/bin`, where `<X>` is the GHC version used
+6. repeat step 4 and 5 for all desired GHC versions
+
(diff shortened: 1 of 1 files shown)


docs/Build.md
```diff
@@ -10,41 +10,67 @@ The design of the build system has the following main goals:
...
+6. repeat step 4 and 5 for all desired GHC versions
+
+This ensures that a complete install is always possible after each `git pull` or a `git clone`.
+
+### Safety checks
+
+The `install.hs` script performs some checks to ensure that a correct installation is possible and provide meaningful error messages for known issues.
+
+* `stack` needs to be up-to-date. Version `1.9.3` is required
+* `ghc-8.6.3` does not work in windows. Trying to install `hie-8.6.3` on windows is not possible
```
Anrock: I think adding some context with ref to #1008 would be nice.

docs/Build.md
```diff
@@ -10,41 +10,67 @@ The design of the build system has the following main goals:
...
-### Tradeoffs
+### Targets
 
-#### `shake.yaml`
+The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
 
-A `shake.yaml` is required for executing the `install.hs` file.
+* `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
+* `build`: builds and installs `hie` binaries for all supported `ghc` versions.
+* `build-doc`: builds the hoogle-db required by `hie`
```
nponeccop: Isn't hoogle-db optional? What is it used for? Documentation in tooltips?

docs/Build.md
```diff
@@ -10,41 +10,67 @@ The design of the build system has the following main goals:
...
+* `cabal-*`: execute the same task as the original target, but with `cabal` instead of `stack`
```
nponeccop: Is it cabal new-build? Building anything with _plain_ cabal is abomination since the days of cabal-sandbox :)

docs/Build.md
```diff
@@ -10,41 +10,67 @@ The design of the build system has the following main goals:
...
 
-* It contains the required version of `shake`.
-* In contrast to the other `*.yaml` it does not contain the submodules, which is necessary for `stack` to work even before the submodules have been initialized.
+Each `stack-*.yaml` contains references to packages in the submodules. Calling `stack` with one of those causes the build to fail if the submodules have not been initialized already. The file `shake.yaml` solves this issue. Moreover, it specifies the correct version of `shake` and is used for installing all run-time dependencies such as `cabal` and `hoogle` if necessary.
```
nponeccop: > The file `shake.yaml` solves this issue.

I think it's important to state that it solves the issue by running git binary itself (so no magic or Haskell-based libgit if such things exist).

docs/Build.md
```diff
@@ -10,41 +10,67 @@ The design of the build system has the following main goals:
...
 * works identically on every platform
 * has minimal run-time dependencies:
-    - a working installation of `stack`
+    - `stack`
     - `git`
-* is completely functional right after simple `git clone`
-* one-stop-shop for building all executables required for using `hie` in IDEs.
+* is completely functional right after a simple `git clone` and after every `git pull`
+* one-stop-shop for building and naming all executables required for using `hie` in IDEs.
+* fails with meaningful error messages in known circumstances
```
nponeccop: * prevents certain build failures by either identifying a failed precondition (such as wrong `stack` version) or by performing the necessary steps so users can't forget them (such as invoking `git` to update submodules)

docs/Build.md
power-fungus: Good idea!

docs/Build.md
power-fungus: yes, it uses `cabal new-*` in the background. What wording would you prefer?

docs/Build.md
power-fungus: I reworded the first sentence to "`ghc-8.6.3` is broken on windows.", which is a bit clearer What additional context would you like?

docs/Build.md
power-fungus: `hie` will display a warning on startup if it is not there or corrupt. It is therefor needed for a *complete* install of `hie`. @fendor do you know what it is used for?

The `install.hs` does not force you to execute the target: `build` installs all `hie-*` binaries, `build-doc` calls `hoogle generate` in some form.

docs/Build.md
power-fungus: I will include this wording

docs/Build.md
fendor: Hoogle is used for a bunch of stuff, like import of Modules, adding packages, tooltips and soon import-list imports as well.

docs/Build.md
fendor: I would not change that, since in the next major release, the nix-style builds are the default, making it legacy knowledge to know the difference between `v1-*` and `v2-*` commands.

docs/Build.md
nponeccop: Does it make sense to always do `build-doc` then? Why `build-doc` is a separate step now? If too many features rely on build-doc should we say somewhere that we build not merely to have doc but to support many features of HIE? I was under impression that build-doc was optional because my editor (nvim) doesn't have a capability to display docs.

docs/Build.md
nponeccop: ok then

PR #1220, opened 2019-04-30
Prevent hie crash if apply-refact crashes

### LINKED ISSUE #451: HIE stops working in VS Code when apply-refact fails
I've noticed the following issue with `apply-refact`: mpickering/apply-refact#33

I'm not sure whether this is intended behavior on the side of `apply-refact` or not, but...
As soon as apply-refact throws a parse error, HIE stops working completely in VS Code: no more hints or compiler errors are shown, hovering over terms stops working, etc.

It would be very nice if we would not have to reload the VS Code session when an external tool fails in some way.

Hogeyama:
Fixes #451

Note: We need to catch `SomeException` because `applyRefactorings` throw an error by `error`.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
+++ b/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
@@ -6,7 +6,9 @@ module Haskell.Ide.Engine.Plugin.ApplyRefact where
 
 import           Control.Arrow
 import           Control.Exception              ( IOException
-                                                , SomeException
+                                                , ErrorCall
+                                                , Handler(..)
+                                                , catches
                                                 , try
                                                 )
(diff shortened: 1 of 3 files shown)


fendor:
We never catch `SomeException`, because of async exceptions.
If the error has been issued by `error` you can catch ErrorCall

Hogeyama:
Oh, I missed the edit. I'll rewrite it by using `tryJust`

fendor:
If it works, I wouldn't change it. I thought, it may not do what we expected.

However, do you have some kind of tests to validate that this really fixes the problem? Would be great to have a test or two for regression testing!

fendor:
Looks great now!

Hogeyama:
Thank you!

Issue #1218, opened 2019-04-29
Should we update submodules with --recursive?


alanz:
As far as I know none of our dependencies *need* a recursive clone, and some of them bring in copies of submodules we already provide ourselves, which are used when building the package in standalone fashion.

So things will go faster and probably still work if we do not do the recursive update.

power-fungus:
Will test!

nponeccop:
See #1256

power-fungus:
closed by #1221

PR #1221, opened 2019-04-30
add possibility to run `install.hs` from cabal

### LINKED ISSUE #1179: Explain how to install with cabal in README

power-fungus:
closes #1179, for good! 

I spoke with @hvr, who hinted at the possibility that `install.hs` could be run via `cabal new-run install.hs` and I got this working.

(planned) Features:
* `install.hs` can be executed via `cabal`: `cabal new-run install.hs --project-file shake.project <target>`
* the targets `build` and `hie-*` will use the same build-system 

Tradeoffs and limitations:
* `shake.project` is needed for similar reasons `shake.yaml` is required.
* a own cabal-package in the folder `install/` is required
* `./install.hs` as an executable is only possible for either `stack` or `cabal`
* The argument `--project-file shake.project` is required and cannot be defined inside the script-comments. @hvr said that this could maybe be changed in the future.

Plan:
- [x] make `install.hs` executable via `cabal new-run` via a comment at the beginning
- [x] refactor `install.hs` logic into own cabal-package
- [x] mechanism for detecting if run from stack or cabal
- [x] make default targets (i.e. `hie-*`, `build`) use the same build-system as the `install.hs` has been called with
- [x] document this feature in readme


install/src/Cabal.hs
```diff
@@ -0,0 +1,109 @@
...
+  localBin <- getLocalBin
+  execCabal_
+    [ "new-install"
+    , "--write-ghc-environment-files=never"
+    , "--symlink-bindir=" ++ localBin
+    , "exe:hie"
+    , "--overwrite-policy=always"
+    ]
+  liftIO $ do
+    copyFile (localBin </> "hie" <.> exe)
```
fendor: Did you try, instead of resorting to `copyFile` using `copyFile' ... >> need ["filename"]` ?

install/src/Help.hs
```diff
@@ -0,0 +1,149 @@
+-- |Module for Help messages and traget descriptions
+module Help where
+
+import           Development.Shake
+import           Data.List                                ( intersperse
+                                                          , intercalate
+                                                          )
+
+import Env
+import Print
```
fendor: Maybe format the imports in the same style?

install/src/Install.hs
```diff
@@ -0,0 +1,144 @@
...
+  , "text-icu"
+  , "--extra-lib-dirs=/usr/local/opt/icu4c/lib"
+  , "--extra-include-dirs=/usr/local/opt/icu4c/include"
+  ]
+
+-- | update the submodules that the project is in the state as required by the `stack.yaml` files
+updateSubmodules :: Action ()
+updateSubmodules = do
+  command_ [] "git" ["submodule", "sync"]
+  command_ [] "git" ["submodule", "update", "--init"]
```
fendor: Is that a fix for #1218? Is that tested? Is that intentional? Should mention it in the PR.

install/src/Print.hs
```diff
@@ -0,0 +1,45 @@
...
+import           Data.List                                ( dropWhileEnd
+                                                          , dropWhile
+                                                          )
+import           Data.Char                                ( isSpace )
+
+out :: MonadIO m => String -> m ()
+out  = liftIO . putStrLn
+
+out' :: MonadIO m => String -> m ()
+out' = out . ("    " ++)
```
fendor: better naming since it has been promoted to a real util function? And/or documentation.

install/src/Stack.hs
```diff
@@ -0,0 +1,98 @@
...
+  `actionOnException` liftIO (putStrLn stackBuildFailMsg)
+
+-- | copy the built binaries into the localBinDir
+stackInstallHie :: VersionNumber -> Action ()
+stackInstallHie versionNumber = do
+  execStackWithGhc_ versionNumber ["install"]
+  localBinDir <- getLocalBin
+  let hie = "hie" <.> exe
+  liftIO $ do
+    copyFile (localBinDir </> hie)
```
fendor: Same as for `cabalInstallHie`

install/src/Stack.hs
```diff
@@ -0,0 +1,98 @@
...
+    error $ stackExeIsOldFailMsg stackVersion
+
+-- | Get the local binary path of stack.
+-- Equal to the command `stack path --local-bin`
+getLocalBin :: Action FilePath
+getLocalBin = do
+  Stdout stackLocalDir' <- execStackShake ["path", "--local-bin"]
+  return $ trim stackLocalDir'
+
+stackBuildData :: Action ()
```
fendor: -> `stackBuildHoogleDatabase` or something like that?
data is way more general than it actually is.

install/src/Version.hs
```diff
@@ -0,0 +1,24 @@
...
+versionToString :: RequiredVersion -> String
+versionToString = showVersion . makeVersion
+
+-- | Parse a version-string into a version. Fails if the version-string is not valid
+parseVersionEx :: String -> Version
+parseVersionEx = fst . head . filter (("" ==) . snd) . readP_to_S parseVersion
+
+
+checkVersion :: RequiredVersion -> String -> Bool
+checkVersion required given = parseVersionEx given >= makeVersion required
```
fendor: Document what `checkVersion` actually does.

shake.project
```diff
@@ -0,0 +1,2 @@
+packages:
```
fendor: add an additional cabal.project in the `install/` such that we can compile the executable from hand if we want to

install/src/Cabal.hs
power-fungus: I did not. This seemed to complex and I do not want `shake` to get in the way. A simple `copyFile` is indeed all that is needed here.

install/src/Help.hs
power-fungus: Done

shake.project
power-fungus: done

install/src/Install.hs
power-fungus: This is intentional. Tests have been successful.

install/src/Stack.hs
power-fungus: When other data (e.g. `hlint`'s data-files) is required, this function will do that as well. I would keep the name identical to the target and I dont want to change the name of the target

install/src/Install.hs
fendor: Then also mention it in the pr!

power-fungus:
also fixes #1218. Tests have been successful

power-fungus:
Currently, `stack` is still needed to get the `.local/bin` dir (`stack path --local-bin`). I'm unable to find a similar command using `cabal` only.

power-fungus:
Currently, `stack` is still needed to get the `.local/bin` dir (`stack path --local-bin`). I'm unable to find a similar command using `cabal` only.

power-fungus:
This PR is now ready. I have done a lot of tests on archlinux and fedora.

README.md
```diff
@@ -201,6 +201,18 @@ stack ./install.hs help
...
+
+The install-script can be invoked via `cabal` instead of `stack` with the command
+
+```bash
+cabal v2-run ./install.hs --project-file install/shake.project <target>
+```
+
+Unfortunalely, it is still required to have `stack` installed so that the install-script can locate the `local-bin` directory (on Linux `~/.local/bin`) and copy the the `hie` binaries to `hie-x.y.z`, which is required for the `hie-wrapper` to function as expected.
+
+For briefty, only the `stack`-based commands are presented in the following sections.
```
fendor: s/briefty/brevity/

power-fungus: done

install/src/Cabal.hs
```diff
@@ -0,0 +1,109 @@
...
+             (localBin </> "hie-" ++ versionNumber <.> exe)
+    copyFile (localBin </> "hie" <.> exe)
+             (localBin </> "hie-" ++ dropExtension versionNumber <.> exe)
+
+-- TODO: review
```
fendor: This TODO is resolved, afaik?

power-fungus: Yes

fendor:
There seem to be a lot of problems on windows. The script is not executable on windows.

jneira:
It seems the cabal script feature doesn't work in windows as-is.
I've been able to run hie-install:
* adding an executable stanza to `install/hie-install.cabal`:
```
executable install
  main-is: ../install.hs
  build-depends:      base >= 4.9 && < 5
                    , hie-install
  default-language:    Haskell2010
```
* Adding `./install`as package in `project.cabal`
* Running `cabal v2-run install -- help`
* Or running `cabal v2-run install --project-file install/shake.project help`without adding it to `cabal.project`


But it is not a "script" anymore, right? Dont know if we lose anything tough (it is even shorter if you add it to `cabal.project`!)

The error i've got with `cabal new-run install.hs` (removing the `{ -stack .. .}` part just in case) is
```console
# cabal v2-run install.hs --project-file install/shake.project help
Resolving dependencies...
Build profile: -w ghc-8.6.5 -O1
In order, the following will be built (use -v for more details):
 - fake-package-0 (exe:script) (configuration changed)
Configuring executable 'script' for fake-package-0..
Preprocessing executable 'script' for fake-package-0..
Building executable 'script' for fake-package-0..
[1 of 1] Compiling Main             ( Main.hs, path\to\haskell-ide-engine\dist-newstyle\build\x86_64-windows\ghc-8.6.5\fake-package-0\x\script\build\script\script-tmp\Main.o )

Main.hs:1:1: error:
    Could not load module `Prelude'
    It is a member of the hidden package `base-4.12.0.0'.
    Perhaps you need to add `base' to the build-depends in your .cabal file.
    it is a hidden module in the package `hackage-security-0.5.3.0'
    Use -v to see a list of the files searched for.
  |
1 | {- cabal:
  | ^

Main.hs:12:1: error:
    Could not load module `HieInstall'
    It is a member of the hidden package `hie-install-0.8.0.0'.
    Perhaps you need to add `hie-install' to the build-depends in your .cabal file.
    Use -v to see a list of the files searched for.
   |
12 | import HieInstall (defaultMain)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

jneira:
Although we would add `hie-install` to the set of project local packages, i would vote to not use the cabal script feature and go for `cabal new-run install -- <command>`, for all systems.
Or even better, to avoid confusions, `cabal new-run hie-install -- <command>`, moving `install` dir to `hie-install`.

jneira:
I've uploaded the branch with my tests: https://github.com/power-fungus/haskell-ide-engine/compare/install-via-cabal...jneira:install-via-cabal

jneira:
Another option is only add the executable stanza and inform windows users that their command would be `cabal new-run install --project-file=install\shake.project -- <command>`: https://github.com/power-fungus/haskell-ide-engine/compare/install-via-cabal...jneira:install-via-cabal-win?expand=1

However, i think it would be desirable that the command would be the same for all systems.

jneira:
Opened a ticket upstream: https://github.com/haskell/cabal/issues/6122

fendor:
Adding hie-install to `cabal.project` defeats the purpose of the script. It would require the user to manually initialise the submodules, of which previously the install script took care of. I would prefer it if this script stays independent of `hie`'s dependencies. 

Since cabal-new-install does not work on windows with version `2.4.1.0`, I think we can safely assume, that nobody on windows can use the cabal script. So, windows only has the possibility to run the stack script, while linux users can do both, I suggest we we wait until the issue is resolved upstream and we update this install script to work with `cabal 3.0.0.0`.

jneira:
@fendor ok, we can wait to full cabal support having an alternative for windows

power-fungus:
Thank you @jneira for your in-depth review. According to tests of @fendor one of the issues is that the root-module of the lib `hie-install` was called `Install` and the script-entry `install-hs`. On windows, paths are not case sensitive, so when looking for the module `Install.hs`, `ghc` finds the `install.hs`, which is not a valid lib-module. 

I renamed the module `Install` from the package `hie-install` to `HieInstall` and now I am able to start it in windows. See: https://dev.azure.com/haskell-ide-engine/haskell-ide-engine-power-fungus/_build/results?buildId=92

power-fungus:
@jneira I appreciate your review. 

I like the idea of the the commit you proposed. You added an executable to `hie-install` with `main-is: ../install.hs`. Does this work even if `install.hs` does not have a module-header and starts with `#!...`? How is this supposed to be called? `cabal v2-build --project-file install/shake.project && cabal v2-run exe:hie-install --project-file install/shake.project`?
Could you add an additional line in the `install.hs` to document how to call this and I am happy to accept an PR for this cange.

The [CI build](https://dev.azure.com/haskell-ide-engine/haskell-ide-engine-power-fungus/_build/results?buildId=92) shows that the script works fine on windows when run with `stack` :tada: For me, this is a good first step. 


I will do a bit more experiments in the next few days and integrate the `install.hs` process into the azure-ci pipeline.

jneira:
@power-fungus Yeah it would work even without any header cause it only  has to be a simple haskell module with `main :: IO()`. However t is unusual to point in a executable stanza to a module in the parent folder of the `.cabal` file (so arguably it is kinda a hack)
Otoh it seems that cabal support [scripting for windows in the dev 2.4 branch](https://github.com/haskell/cabal/issues/6122#issuecomment-507338322) so i think we can wait to the next cabal release that hopefully will have support for scripting and v2-installing for windows, as suggested by @fendor.
I am just building cabal from master to test the branch.

jneira:
With cabal built from 2.4 branch the script works

fendor:
So, stack works for windows right? 
Maybe for now, we could document in the readme that on windows, only execution with stack is supported. Otherwise, I am in favour of merging this!

power-fungus:
https://dev.azure.com/haskell-ide-engine/haskell-ide-engine-power-fungus/_build/results?buildId=103 shows the error on cabal on windows

fendor:
This PR is looking great, I think. We should wait for the next release, and merge it after it, so we can do some dog fooding.

power-fungus:
Let me rebase this and add ci integration and this is ready to merge

README.md
```diff
@@ -201,6 +201,18 @@ stack ./install.hs help
...
 
+##### Install via cabal
+
+The install-script can be invoked via `cabal` instead of `stack` with the command
+
+```bash
+cabal v2-run ./install.hs --project-file install/shake.project <target>
+```
+
+Unfortunately, it is still required to have `stack` installed so that the install-script can locate the `local-bin` directory (on Linux `~/.local/bin`) and copy the the `hie` binaries to `hie-x.y.z`, which is required for the `hie-wrapper` to function as expected.
```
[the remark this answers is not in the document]
power-fungus: Consistency with `stack` is not really required. Consistency with `hie-wrapper` is desired

azure-pipelines.yml
```diff
@@ -2,3 +2,5 @@ jobs:
 - template: ./.azure/linux-stack.yml
 - template: ./.azure/windows-stack.yml
 - template: ./.azure/macos-stack.yml
+- template: ./.azure/windows-installhs-stack.yml
```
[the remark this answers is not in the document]
power-fungus: done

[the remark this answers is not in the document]
power-fungus: see: https://dev.azure.com/haskell-ide-engine/haskell-ide-engine-power-fungus/_build/results?buildId=108

PR #1222, opened 2019-04-30
Add Documentation for Pragmas Plugin

fendor:


src/Haskell/Ide/Engine/Plugin/Pragmas.hs
```diff
@@ -53,15 +58,22 @@ addPragmaCmd = CmdSync $ \(AddPragmaParams uri pragmaName) -> do
 
 -- ---------------------------------------------------------------------
 
+-- | Offer to add a missing Language Pragma to the top of a file.
+-- Pragmas are defined by a curated list of known pragmas, see 'possiblePragmas'.
+-- May also offer to add unknown pragmas, if the pragma is a subexpression of
+-- an existing plugin.
```
alanz: I do not understand the last two lines, can you try use a different wording.

src/Haskell/Ide/Engine/Plugin/Pragmas.hs
```diff
@@ -71,13 +83,16 @@ codeActionProvider plId docId _ (J.CodeActionContext (J.List diags) _monly) = do
...
 
+-- | Find all Pragmas that have the search term as infix.
 findPragma :: T.Text -> [T.Text]
 findPragma str = concatMap check possiblePragmas
   where
     check p = [p | T.isInfixOf p str]
 
 -- ---------------------------------------------------------------------
 
+-- | Possible Pragama names.
```
alanz: Typo, `pragama`

src/Haskell/Ide/Engine/Plugin/Pragmas.hs
fendor: It was wrong anyways, thanks for making me look at it again.

src/Haskell/Ide/Engine/Plugin/Pragmas.hs
```diff
@@ -71,13 +81,16 @@ codeActionProvider plId docId _ (J.CodeActionContext (J.List diags) _monly) = do
 
 -- ---------------------------------------------------------------------
 
+-- | Find all Pragmas are an infix of the search term.
 findPragma :: T.Text -> [T.Text]
 findPragma str = concatMap check possiblePragmas
```
fendor: this could be just
`findPragma str = filter (\p -> T.isInfixOf p str) possiblePragmas` right?

alanz: I think so.  But that change should be in a separate PR, this one is documentation only

PR #1240, opened 2019-05-04
Troubleshooting for emacs

Infinisil:
- Fix for Emacs document sync problems. I'm not sure what the actual cause is, but this is a workaround.
- Fix for emacs-direnv loading, see https://github.com/wbolster/emacs-direnv/issues/17 as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ we talk to clients.__
   - [Documentation](#documentation)
     - [Architecture](#architecture)
   - [Troubleshooting](#troubleshooting)
+    - [Emacs](#emacs)
+      - [Parse errors, file state going out of sync](#parse-errors-file-state-going-out-of-sync)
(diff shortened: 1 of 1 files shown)


alanz:
@Infinisil if the file state gets out of sync for emacs on incremental changes, this may be a bug in `emacs-lsp`.  But I use it all the time and do not observe this, would it be possible to come up with some way of reproducing it reasonably reliably?

Infinisil:
Hmm I can't seem to reproduce it at will right now. I'll turn off that setting for now and see if the problem starts occurring again.

Infinisil:
Yeah after using it a while without this workaround, it still happens occasionally. It seems to happen more often at the `import` section, or with larger edits, but nothing reproducible as far as I can see. I'll try to capture the logs of it next time. Until this gets fixed however, these workarounds would be nice to have in the readme, and I know of at least another person that has complained about the out-of-sync problem.

PR #1241, opened 2019-05-04
Add an explicit stack file for GHC 8.6.5

alanz:
And schedule circle CI for it, even though it is a repeat of the
nightly one.

Very pleasantly surprised to see that `install.hs` needs no change to
pick this up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,6 +146,11 @@ jobs:
       - STACK_FILE: "stack-8.6.4.yaml"
     <<: *defaults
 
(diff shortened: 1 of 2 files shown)


alanz:
Note, it disables the circle CI build for GHC 8.6.3, as that version is basically deprecated anyway.

If anyone feels strongly we can re-enable it

alanz:
Ans I just realised that waiting fro appveyor to pass this PR is crazy, as nothing changed that affects those builds.

PR #1242, opened 2019-05-04
Use floskell from hackage

bubba:
Thanks to @ennocramer for the upload https://github.com/ennocramer/floskell/issues/6#issuecomment-489230007


alanz:
@bubba did you follow the process outlined in the top of `.gitmodules` to remove the submodule?

bubba:
@alanz I've deleted and checked in submodules/floskell folder, but was not able to find/delete it in `.git/modules`, although this is inside a different worktree, whilst I also have master checked at the same time.

PR #1245, opened 2019-05-05
Use new cabal-helper

bubba:
This removes cabal-helper-wrapper, and so will open up the way for static binaries until hie-bios lands. 
Also noticeably, it removes the build plugin since it relied on the old cabal-helper API: I plan on removing this in #1105 anyway as it is only used by the old JSON transport. 

Please dogfood this a bit!


bubba:
@alanz this is all using your work on new-cabal-helper with ghc-mod, should I create a PR to merge this into your fork?

test/functional/FunctionalBadProjectSpec.hs
```diff
@@ -32,7 +32,7 @@ spec = describe "behaviour on malformed projects" $ do
           d ^. code `shouldBe` Nothing
           d ^. source `shouldBe` Just "ghcmod"
           d ^. message `shouldBe`
-            (T.pack "readCreateProcess: stack \"build\" \"--only-configure\" \".\" (exit 1): failed\n")
+            (T.pack "callProcessStderr: stack --stack-yaml=/root/build/test/testdata/badProjects/cabal/stack.yaml build --only-configure . (exit 1): failed\n")
```
alanz: For the HaRe tests I copy certain projects to $TEMPDIR/testproj, and then work with that.  This looks like a very brittle hard coded path.

bubba: I never even noticed the path, I just copied and pasted it verbatim. Will do a fuzzier expectation

fendor:
While the stability seems to be improved in general, it reopens #1264, probably since the patch for it was submitted after this pr had been opened.
Patch was in https://github.com/alanz/ghc-mod/pull/20/files

fendor:
When you open `haskell-ide-engine` repository, you no longer get diagnostics for modules in the sub-package `hie-plugin-api`. Instead, it fails with the error message, when I open `TypeMap.hs`:
```
/home/user/Documents/haskell/bubba-hie/hie-plugin-api/stack.yaml: getFileStatus: does not exist (No such file or directory)
```

NickSeagull:
What's missing for this to be merged? I see that the cabal job has passed, but the others are failing with some `allow-newer` errors. I guess this is due to the version bump of `cabal-helper`?

I'd love to help in order to get this merged

bubba:
The cabal job is passing just because the cabal job doesn’t run any tests 😬
The first thing would be to probably get the remaining tests passing locally, and from what I can remember there were some new problems when working with cabal projects nested within each other, like @fendor pointed out above with the hie-plugin-api sub-project in haskell-ide-engine itself.
PRs to this branch are more than welcome! :)

bubba:
This is done in hie-bios

PR #1246, opened 2019-05-07
hie sends invalid message on hover

Hogeyama:
When there is no information available against a hover request, hie returns a response like

```json
{"result":{"contents":null},"jsonrpc":"2.0","id":1}
```

which does not respect the Language Server Protocol in which `contents` cannot be `null`:

```typescript
interface Hover {
    contents: MarkedString | MarkedString[] | MarkupContent;
    range?: Range;
}
```

This PR includes a temporal measure for this problem.
It would be better to fix `haskell-lsp` so as not to allow `null` in `HoverContents`.
I've opened the [issue](https://github.com/alanz/haskell-lsp/issues/162) on `haskell-lsp` too.


alanz:
See discussion at https://github.com/alanz/haskell-lsp/issues/162

PR #1247, opened 2019-05-07
Adapting to new haskell-lsp

alanz:
- It uses `rope-utf16-splay` instead of `yi-rope`
- Moving some functionality and types from `haskell-ide-engine` to `haskell-lsp`, to benefit other LSP servers based on it
- It is a step toward bringing in the `hie-bios` being worked on by @mpickering 

This is WIP pending merges of 
- https://github.com/alanz/haskell-lsp/pull/164
- https://github.com/alanz/lsp-test/tree/az-hie-bios

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,9 @@
 	# url = https://github.com/bubba/ghc-mod.git
 	url = https://github.com/alanz/ghc-mod.git
 
+[submodule "submodules/haskell-lsp"]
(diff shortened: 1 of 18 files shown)


alanz:
I am planning on merging this when CI passes, so I can fix https://github.com/alanz/haskell-lsp/issues/162 before publishing a new `haskell-lsp`.

PR #1255, opened 2019-05-11
Narrow ghc mod core

alanz:
Since ghc-mod is going to be deprecated soon, and replaced by some new shiny thing as part of @DanielG 's GSOC, this patch attempts to help the process by slimming down and documenting the actual parts of ghc-mod currently used by hie.


src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -134,6 +123,7 @@ typeCmd :: CommandFunc TypeParams [(Range,T.Text)]
 typeCmd = CmdSync $ \(TP _bool uri pos) ->
   liftToGhc $ newTypeCmd pos uri
 
+-- AZ: currently only used in tests, but
```
fendor: but what?

PR #1265, opened 2019-05-22
Bump version for hsimport to 0.10.0

### LINKED ISSUE #1228: HsImport inserts module between function and Haddock documentation.
Has been reported upstream ~[hsimport/#16](https://github.com/dan-t/hsimport/issues/16)~. We have to wait for a new release, then this issue can be closed.

Example is roughly taken from `stack new test haskeleton`
Example.hs
```haskell
-- | An example module.
module Example (main) where

-- | An example function.
main :: IO ()
main = return ()

foo :: IO ()
foo = do
    _ <- async $ return ()
    return ()
```
```bash
> hsimport -m "Control.Concurrent.Async" Example.hs
```

Output: 
```haskell
-- | An example module.
module Example (main) where

-- | An example function.
import Control.Concurrent.Async
main :: IO ()
main = return ()

foo :: IO ()
foo = do
    _ <- async $ return ()
    return ()
```

Expected:
```haskell
-- | An example module.
module Example (main) where

import Control.Concurrent.Async
-- | An example function.
main :: IO ()
main = return ()

foo :: IO ()
foo = do
    _ <- async $ return ()
    return ()
```

Happens when using the code action to import a module.

fendor:
closes #1228


fendor:
I actually want to add a test for that version bump, since it solves the issues described in #1228.

PR #1267, opened 2019-05-23
Support infix completions

### LINKED ISSUE #1259: Completions should be aware of infix backticks
Doing completion on

```haskell
"a" `should
```
gives the option `shouldBe`, but the completion/snippet does not have the closing backtick, and is presented as requiring two more params.

fendor:
closes #1259

* ~[ ] More Documentation~
* [x] More Tests.

### DIFF
--- a/test/functional/CompletionSpec.hs
+++ b/test/functional/CompletionSpec.hs
@@ -257,13 +257,13 @@ spec = describe "completions" $ do
       let te = TextEdit (Range (Position 5 7) (Position 5 24)) "even `filte"
       _ <- applyEdit doc te
 
-      compls <- getCompletions doc (Position 5 17)
+      compls <- getCompletions doc (Position 5 18)
       let item = head $ filter ((== "filter") . (^. label)) compls
       liftIO $ do
         item ^. label `shouldBe` "filter"
         item ^. kind `shouldBe` Just CiFunction
         item ^. insertTextFormat `shouldBe` Just Snippet
-        item ^. insertText `shouldBe` Just "`filter`"
+        item ^. insertText `shouldBe` Just "filter`"
 
     it "work for infix functions in backticks" $ runSession hieCommand fullCaps "test/testdata/completion" $ do
       doc <- openDoc "Completion.hs" "haskell"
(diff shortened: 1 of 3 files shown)


alanz:
Looking good.

I think as time passes we are going to end up with ever more tightly focused contexts, so the completion gets more reliable.

test/functional/CompletionSpec.hs
```diff
@@ -250,6 +250,67 @@ spec = describe "completions" $ do
...
+      let te = TextEdit (Range (Position 5 7) (Position 5 24)) "even `filte"
+      _ <- applyEdit doc te
+
+      compls <- getCompletions doc (Position 5 18)
+      let item = head $ filter ((== "filter") . (^. label)) compls
+      liftIO $ do
+        item ^. label `shouldBe` "filter"
+        item ^. kind `shouldBe` Just CiFunction
+        item ^. insertTextFormat `shouldBe` Just Snippet
+        item ^. insertText `shouldBe` Just "filter`"
```
lorenzo: Can we add some assertions about the placeholders the function is suggesting after completing? It would be painful if the placeholders we suggest contain the left hand side of the function

fendor: The placeholders are the expected type? e.g. in the case of filter: `filter' {2:[a]}`?

bubba: With backticks the placeholders should just drop the first element right? Because we can assume that the left hand side will supply the first argument

alanz:
I think we should wait for the more complete tests before merging this.

fendor:
> Asides from what @lorenzo said this looks good! By the way have you been able to get the tests running locally yet?

Yes, with some workaround. First I dont use stack to get the hie executable, I just use the one on the path and then I can use `cabal new-run :func-test`

test/functional/CompletionSpec.hs
fendor: I dont know. Is this really wanted? I find myself often writing such functions in exisiting expressions, normally I dont want these placeholders...

alanz: This is something that has crossed my mind too, but I think it needs a more general solution, to include in the context if there is anything to the right already, and in that case not offering placeholders

fendor: Maybe I am also misunderstanding the question. This particular completion omits the placeholders completely at the moment.

alanz: So you are proposing that we keep it like that?  Which could work. If anyone feels strongly about it we can add them again. I do think that infix operations tend to be binary, so the missing placeholder(s) is probable not an issue.

fendor: Yes, I think it would be better and we can add it anytime.

alanz:
Ok, LGTM, will merge after the pending monthly release.

Issue #20, opened 2015-11-05
Rework the Console


alanz:
At the moment the [Console](https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/Console.hs) is incredibly basic.

It should be reworked to provide the following basics

- [x]  History. Pressing up/down should scroll through prior history, allow editing

- [ ]  Better command parsing
       At the moment the command parsing is as simple as possible, for the POC. This should be extended in some way, perhaps by generating parsers from plugin command descriptors [UiCommand](https://github.com/haskell/haskell-ide-engine/blob/master/haskell-ide-plugin-api/Haskell/Ide/Engine/PluginDescriptor.hs#L42)

- [ ] Better representation of output

- [ ]  Built in commands to browse the plugins, and get help on them. This should just be wrappers around the functions exposed via the [base plugin](https://github.com/haskell/haskell-ide-engine/blob/wip-az/src/Haskell/Ide/Engine/BasePlugin.hs#L24)

- [ ]  Anything else

alanz:
Probably makes sense to use https://hackage.haskell.org/package/haskeline-0.7.2.1/docs/System-Console-Haskeline.html

An example use of this library for a simple read-eval-print loop (REPL) is the following:

```haskell
import System.Console.Haskeline

main :: IO ()
main = runInputT defaultSettings loop
   where 
       loop :: InputT IO ()
       loop = do
           minput <- getInputLine "% "
           case minput of
               Nothing -> return ()
               Just "quit" -> return ()
               Just input -> do outputStrLn $ "Input was: " ++ input
                                loop
```

alanz:
Ok, item 1 is done, Haskeline in place. Lots more available though

rvion:
@alanz tip: in the issue description, if you use 

- [ ] aaa (`- [ ] aaa`)
and 
- [X] aaa (`- [X] aaa`)

the number of items to be done and the progression will be understood by github and integrated in various places 
exemple:
![image](https://cloud.githubusercontent.com/assets/2150990/10982765/802f821c-840f-11e5-9d47-e9bc4020b05a.png)

Also, you'll directly be able to click on checkboxes to mark them as done
other ex: https://github.com/neovim/neovim/pull/243
![image](https://cloud.githubusercontent.com/assets/2150990/10982790/a9d2d3ee-840f-11e5-9263-e7c372c8d9a9.png)

gracjan:
[...]

PR #1269, opened 2019-05-23
Use ghc-mod which loads GHC plugins specified in Cabal file

alanz:
It does not load ones specified with a GHC pragma in the source file.

Thanks @fendor


fendor:
Does ghc expose the `initializePlugins` before ghc 8.6.1? It seems like plugins were introduced way earlier.

fendor:
should be closed in favour of [#20](https://github.com/alanz/pulls/20).
Will also fix the compile errors by using CPP instructions.

PR #1270, opened 2019-05-25
Hlint 2.1.22

alanz:
Bring in hlint-2.1.22, which uses `ghc-lib-parser` when it can, instead of `haskell-src-exts`.

It does not support GHC < 8.4, so those still use the older `hlint`.


test.sh
```diff
@@ -0,0 +1,17 @@
...
+# stack test haskell-ide-engine:func-test --test-arguments "--failure-report .hspec-failures --rerun"
+# stack test haskell-ide-engine:func-test --test-arguments "--match \"completes ghc options pragma values\""
+#stack test haskell-ide-engine:func-test --test-arguments "--match \"behaviour on malformed projects\""
+
+# stack --stack-yaml=stack-8.6.4.yaml test haskell-ide-engine:func-test --test-arguments "--match \"completes imports\""
+# stack --stack-yaml=stack.yaml test haskell-ide-engine:func-test --test-arguments "--match \"completes pragmas no close\""
+
+# stack --stack-yaml=stack-8.6.4.yaml test haskell-ide-engine:unit-test --test-arguments "--match \"ApplyRefactPlugin\""
+
+# stack --stack-yaml=stack.yaml test haskell-ide-engine:func-test --test-arguments "--match \"/FunctionalCodeActions/code actions/hlint suggestions/provides 3.8 code actions/\""
```
bubba: Did you mean to add this file?

alanz: I added it to help the `ghc-lib-parser` guys track down the problem. I guess it can be removed again.

PR #1273, opened 2019-05-30
Avoid legacy warning

### LINKED ISSUE #1201: v1 style warnings during stack ./install.hs
`Warning: The update command is a part of the legacy v1 style of cabal usage.` and I think a few different ones later.

fendor:
closes #1201

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -247,7 +247,7 @@ installCabal = do
   -- install `cabal-install` if not already installed
   when (isNothing cabalExe) $
     execStackShake_ ["install", "cabal-install"]
-  execCabal_ ["update"]
+  execCabal_ ["v1-update"]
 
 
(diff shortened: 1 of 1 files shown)


install.hs
```diff
@@ -247,7 +247,7 @@ installCabal = do
   -- install `cabal-install` if not already installed
   when (isNothing cabalExe) $
     execStackShake_ ["install", "cabal-install"]
-  execCabal_ ["update"]
+  execCabal_ ["v1-update"]
```
alanz: I think we had this before, and it broke installs for people using earlier versions of cabal.

fendor: Probably true, but we demand that the cabal version is at least `2.4.1.0`. this is checked before that.

PR #1274, opened 2019-05-30
Bump resolver for ghc 8.6.3 due to lts mutation

### LINKED ISSUE #1236: LTS-13.11 is mutated back to ghc-8.6.3
https://github.com/commercialhaskell/lts-haskell/commit/ebafeb5402c9bb52bdf236b699490c28fb907972

So we can bump the snapshot for hie-8.6.3

fendor:
Closes #1236.
Caveat: On my system this still tries to use ghc 8.6.4? Although it should use ghc 8.6.3.

### DIFF
--- a/stack-8.6.3.yaml
+++ b/stack-8.6.3.yaml
@@ -1,4 +1,4 @@
-resolver: lts-13.10 # Last GHC 8.6.3
+resolver: lts-13.11 # Last GHC 8.6.3
 packages:
 - .
 - hie-plugin-api


fendor:
I think, if the resolver has already been downloaded, it will not be updated and use ghc 8.6.4 nevertheless, although it should use ghc 8.6.3.
It does not seem to be worth the hassle and I would suggest closing this PR and closing the regarding issue.

alanz:
I agree with @fendor, I don't believe the hassle it will cause out there is worth the difference between the two resolver versions.  See https://www.stackage.org/diff/lts-13.10/lts-13.11

`haskell-lsp` has changed, but we specify it in the stack files anyway to provide the current version.

PR #1279, opened 2019-06-01
Drop GHC 8.2.1 support

### LINKED ISSUE #1249: Drop 8.2.1 support?
We're supporting quite a lot of builds right now, and 8.2.1 has been around for a while. If there is not a lot of interest in keeping it on then we should remove it make CI easier.
It'll also allow us to clean up all those CPP pragmas where we need to explicitly import Data.Semigroup

alanz:
Closes #1249


lorenzo:
Don’t we need to drop the circleci configuration too?

alanz:
Good point.

alanz:
The appveyor build overview does not show an 8.2.1 job, so merge should be ok.

Issue #1191, opened 2019-04-23
Add code action for hiding imports


expipiplus1:
It would be nice if there was a code action for hiding imports which are unused and being shadowed by local declarations.

This is complicated because it requires some detailed information on the provenance of these names in order to hide them (for example if they are record fields).

An example:

Given the following warning:
```
triangle/Main.hs:83:26: warning: [-Wname-shadowing]
    This binding for ‘window’ shadows the existing bindings
      imported from ‘Graphics.Vulkan’ at triangle/Main.hs:(29,1)-(45,59)
      (and originally defined in ‘Graphics.Vulkan.Extensions.VK_NN_vi_surface’)
      imported from ‘Graphics.Vulkan’ at triangle/Main.hs:(29,1)-(45,59)
      (and originally defined in ‘Graphics.Vulkan.Extensions.VK_KHR_xlib_surface’)
      imported from ‘Graphics.Vulkan’ at triangle/Main.hs:(29,1)-(45,59)
      (and originally defined in ‘Graphics.Vulkan.Extensions.VK_KHR_xcb_surface’)
      imported from ‘Graphics.Vulkan’ at triangle/Main.hs:(29,1)-(45,59)
      (and originally defined in ‘Graphics.Vulkan.Extensions.VK_KHR_android_surface’)
```

To the hiding list of `Graphics.Vulkan` add

```
  XlibSurfaceCreateInfoKHR(window)
, XcbSurfaceCreateInfoKHR(window)
, ViSurfaceCreateInfoNN(window)
, AndroidSurfaceCreateInfoKHR(window)
```

At the moment this is a pain, as one has to visit each of those modules and find out which record is exposing that name.

lorenzo:
An alternate code action could be to make the shadowed function come from a qualified module instead. I often choose to fully qualify things when getting name conflicts

fendor:
Is there a tool to do it already? E.g. could `hsimport` do that?

fendor:
Hsimport does not seem to be able to do that, maybe we can add that to hsimport? It seems at least to be a feature they might want.

lorenzo:
@fendor that sounds like a great idea!

fendor:
I am implementing the feature upstream in [#18](https://github.com/dan-t/hsimport/pull/18)

expipiplus1:
Thanks!

fendor:
[...]

PR #1284, opened 2019-06-03
Improve import action of hsimport

fendor:
Current implementation of the hsimport plugin can not deal with constructors, e.g. trying to import `Null` from `Data.Aeson` will wrongly try to import `Data.Aeson (Null)` instead of the correct `Data.Aeson (Value(Null))`. This pr aims to change that, while also preparing the implementation of #1191. 
Clones the API of HsImport with small adaptations for our use case, e.g. currently no support for qualified imports.
This pr also fixes the import of operators.

* [x] Tests
* [x] Cleanup code

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/Hoogle.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Hoogle.hs
@@ -152,7 +152,19 @@ renderTarget t = T.intercalate "\n" $
 -- If an error occurs, such as no hoogle database has been found,
 -- or the search term has no match, an empty list will be returned.
 searchModules :: T.Text -> IdeM [T.Text]
-searchModules = fmap (nub . take 5) . searchTargets (fmap (T.pack . fst) . targetModule)
+searchModules = fmap (map fst) . searchModules'
+
+-- | Just like 'searchModules', but includes the signature of the search term
+-- that has been found in the module.
+searchModules' :: T.Text -> IdeM [(T.Text, T.Text)]
+searchModules' = fmap (nub . take 5)
+  . searchTargets
+    (\target
+     -> (\modTarget -> (T.pack $ fst modTarget, normaliseItem . T.pack $ targetItem target))
+     <$> targetModule target)
+     where
+      normaliseItem :: T.Text -> T.Text
+      normaliseItem  = innerText . parseTags
 
 -- | Search for packages that satisfy the given search text.
 -- Will return at most five, unique results.
--- a/src/Haskell/Ide/Engine/Plugin/Hoogle.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Hoogle.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE CPP               #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Haskell.Ide.Engine.Plugin.Hoogle where
@@ -5,6 +6,7 @@ module Haskell.Ide.Engine.Plugin.Hoogle where
 import           Control.Monad.IO.Class
 import           Control.Monad (join)
 import           Control.Exception
+import           Control.Applicative (liftA2)
 import           Data.Aeson
 import           Data.Bifunctor
 import           Data.Maybe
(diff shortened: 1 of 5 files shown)


src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -153,6 +204,37 @@ importModule uri importList modName =
                     $ IdeResultOk (J.WorkspaceEdit newChanges newDocChanges)
             else return $ IdeResultOk (J.WorkspaceEdit mChanges mDocChanges)
 
+-- | Convert the import style arguments into HsImport arguments.
+-- Takes an input and an output file as well as a module name.
+importStyleToHsImportArgs
+  :: FilePath -> FilePath -> ModuleName -> ImportStyle -> HsImport.HsImportArgs
```
lorenzo: Would you mind using a `newtype` for `input` and `output` to distinguish them. It wouldd make the type signature more readable and safer too

fendor: sure, will do!

fendor:
Example of what will be possible now:
```haskell
{-# LANGUAGE NoImplicitPrelude #-}
import           System.IO (IO)
-- | Main entry point to the program
main :: IO ()
main =
    when True
        $ hPutStrLn stdout
        $ fromMaybe "Good night, World!" (Just "Hello, World!")
```
Only using Code Actions, this can be turned into:
```haskell
{-# LANGUAGE NoImplicitPrelude #-}
import           System.IO (IO, hPutStrLn, stdout)
import           Prelude (Bool(..))
import           Control.Monad (when)
import           Data.Maybe (fromMaybe, Maybe(Just))
import           Data.Function (($))
-- | Main entry point to the program
main :: IO ()
main =
    when True
        $ hPutStrLn stdout
        $ fromMaybe "Good night, World!" (Just "Hello, World!")
```
Of course, this will be documented via a test.

fendor:
@bubba, would you mind taking a look at the tests again? I can run them, but I dont know how to solve the issue. The test case worked when executing each code action manually.

src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -224,60 +323,173 @@ codeActionProvider plId docId _ context = do
...
+  let extractedTerm = asum
+        [ (\name -> (name, Import Symbol))
+            <$> T.stripPrefix "Variable not in scope: " importMsg
+        , (\name -> (T.init name, Import Type))
+            <$> T.stripPrefix
+              "Not in scope: type constructor or class ‘"
+              importMsg
+        , (\name -> (name, Import Constructor))
+            <$> T.stripPrefix "Data constructor not in scope: " importMsg]
+  in do
```
bubba: Nitpick: can you move the do to the start of the function so we don't need the `in do`?

bubba:
@fendor sure thing! This looks really cool, so I was planning on trying it out anyway :)

src/Haskell/Ide/Engine/Plugin/Hoogle.hs
```diff
@@ -152,7 +152,19 @@ renderTarget t = T.intercalate "\n" $
 -- If an error occurs, such as no hoogle database has been found,
 -- or the search term has no match, an empty list will be returned.
 searchModules :: T.Text -> IdeM [T.Text]
-searchModules = fmap (nub . take 5) . searchTargets (fmap (T.pack . fst) . targetModule)
+searchModules = fmap (map fst) . searchModules'
+
+-- | Just like 'searchModules', but includes the signature of the search term
+-- that has been found in the module.
+searchModules' :: T.Text -> IdeM [(T.Text, T.Text)]
+searchModules' = fmap (nub . take 5)
```
lorenzo: shouldn't this be `take 5 . nub` ? We should first filter duplicates out and then take 5 survivors

src/Haskell/Ide/Engine/Plugin/Hoogle.hs
```diff
@@ -152,7 +152,19 @@ renderTarget t = T.intercalate "\n" $
...
+  . searchTargets
+    (\target
+     -> (\modTarget -> (T.pack $ fst modTarget, normaliseItem . T.pack $ targetItem target))
```
lorenzo: can we maybe take this lambda out into the `where` clause as a 2 argument function? I think it would make it easier to read

src/Haskell/Ide/Engine/Plugin/Hoogle.hs
fendor: Possibly, similar functions use this this, though. I can edit it for these functions as well

src/Haskell/Ide/Engine/Plugin/Hoogle.hs
fendor: you are right

fendor:
Currently, it seems like import actions for tpyes do not work at all. I would like to take a look at it, before this is merged.

src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -224,61 +323,174 @@ codeActionProvider plId docId _ context = do
...
-  msg =
-    head
-        -- Get rid of the rename suggestion parts
+extractImportableTerm :: T.Text -> Maybe (T.Text, SymbolImport SymbolType)
+extractImportableTerm dirtyMsg = do
+  (n, s) <- extractedTerm
+  let n' = T.strip n
+  return (n', s)
+  where
+    importMsg = head
```
lorenzo: I’m always uncomfortable when I see partial functions. Can you use Safe.headMay?

fendor: I thought the same, but it seems to never have exploded onto someone. I will change it.

alanz: @fendor I understand we are waiting for you to change this before we can merge?

fendor: Done now.

PR #1290, opened 2019-06-14
Render completion documentation to markdown

Avi-D-coder:


src/Haskell/Ide/Engine/Plugin/Hoogle.hs
```diff
@@ -103,7 +103,14 @@ infoCmd' expr = do
       if null res then
         Left NoResults
       else
-        return $ T.pack $ targetInfo $ head res
```
lorenzo: can we use pattern matching here? Again I'm uncomfortable with using `head` even if it is obvious that the list here is not empty.

```haskell
liftIO $ runHoogleQuery mdb expr $ \res ->
	      case res of
	        [] -> Left NoResults
	        x:_ -> return $ T.pack $ targetInfo x
```

Avi-D-coder: Thanks

PR #1291, opened 2019-06-15
Stack 2.1.1

alanz:
Down to two problems, both in `unit-test`

```
  # To rerun use: --match "/GhcModPlugin/ghc-mod plugin/ghc-mod plugin commands(old plugin api)/runs the check command/"
  # To rerun use: --match "/Liquid/liquid haskell plugin/runs the liquid haskell exe/"
```

The liquid haskell one comes from installing `liquid` via `cabal new-install liquidhaskell --with-compiler=ghc-8.2.2` and having the package path now set by stack, so the exe does not run properly.

The ghc-mod one has an extra diagnostic returned of the form `["Loaded package environment from /tmp/stack14358/test-ghc-env"])` where the exact directory changes with each test run. So we need to somehow get that path from stack while running the test, so we can plug it into the expected result.


bubba:
> So we need to somehow get that path from stack while running the test, so we can plug it into the expected result.
Can we get away with just checking the prefix of the diagnostic?

alanz:
> Can we get away with just checking the prefix of the diagnostic?

I don't think so. It gets checked as part of a utility function, that calls the old and the new API, and does the one test by comparing generated JSON instances.

The other option is to filter out those reports in ghc-mod or its equivalent, they are incredibly annoying.

alanz:
I think the solution to the liquid haskell test is to move it into func tests, as it does work properly there.

alanz:
One really beneficial effect of the new stack is it caches the entire snapshot. extra deps and all.

So blowing away `.stack-work` only causes the submodules and related to rebuild. This is especially great given the long compile times on `ghc-lib-parser`.

Issue #806, opened 2018-09-06
Have a watchdog restarting hie on high memory usage


lorenzo:
A common complaint about hie is its unbounded memory consumption. This can be due to many different factors, but probably mostly to the famously known memory leak in GHCi, which has mostly been solved in the upcoming GHC 8.6.

For many the road to using 8.6 is far away in the future, so we should mitigate the issue with some other measures. One strategy can be having a separate thread in hie checking on the memory usage and making the server restart when it exceed a certain threshold.

Before restart, hie can send a `window/ShowMessage` explaining what it is doing.

alanz:
This will need some kind of command line parameter, or config setting to decide on the criteria for restart,

lorenzo:
I tried to implement this but found a couple roadblocks:

* The vscode plugin does not seem to be handling restarts successfully. After a restart, opening new files will not get them type checked and will get no diagnostics at all.
* The neovim client does not restart the server when it dies

Given those 2 issues, the experience of restarting the server will be significantly worse than consuming too much memory, so canning this for the time being.

kfigiela:
@lorenzo, the workaround I posted some time ago in the other issue does partially what you want: 
https://github.com/haskell/haskell-ide-engine/issues/412#issuecomment-385033847 OTOH, this does not solve the issues you mentioned.

> Given those 2 issues, the experience of restarting the server will be significantly worse than consuming too much memory
I don't agree with that. Having OS unresponsive is very bad. I am working on not so small 35k LOC project and as soon as HIE uses more than ~4 GB memory my OS (macOS) starts to get sluggish and unresponsive. This is not out of memory condition, but HIE and kernel task are using all CPU (probably this is related to some memory allocation routines).

lorenzo:
@kfigiela I'm not saying the issue should be ignored, only that after trying a restart in vscode and neovim, in both cases I had to restart the editor. That's no different to having the user just do that.

unode:
If using `hie-wrapper` through vscode I get reliable restarts of `hie-x.y`.
If `hie-wrapper` gets killed it's a different story.

Tools like [earlyoom](https://github.com/rfjakob/earlyoom) are a big cannon but may help alleviating the symptom.
You can customize it to prefer killing `hie-x.y` over something else.

yousefvand:
I experience high memory consumption (> 6GB) even with a `hello world` project for a few minutes which brings the system to its knees :cry: 
[...]

Issue #665, opened 2018-06-28
Is it normal for hie to take ~4GB of memory per instance?


CaptJakk:
I currently have a workflow where I usually have a couple different projects open, each of them has the VSCode integration with hie. I find that each editor has it's own hie process and that process takes about 4G of RAM. Is this typical/expected or is something in my setup wrong/different than others?

Bubba:
I believe this is caused by a memory leak in ghci, #580 and https://ghc.haskell.org/trac/ghc/ticket/15111

CaptJakk:
So should I read that as "Yes, but it shouldn't be". If so, is this something that we generally just have to deal with for now or is there a known workaround?

alanz:
We have not been focusing on the memory footprint, more on functionality.

So I have no idea of what savings can be made by optimising things.  But I would prefer to do that once we know that GHC itself is not leaking as much.

CaptJakk:
Understood.

JasonMFry:
Having this issue, too, HIE is taking over 7GBs.

kfigiela:
Possibly duplicate of https://github.com/haskell/haskell-ide-engine/issues/412

codygman:
If I'm understanding #580 correctly, memory leak issues should be fixed if you use GHC 8.6 or higher due to ghcMod's usage of `loadInterface` taking advantage of `dontLeakTheHPT`.

JasonMFry:
Also there's this ugly workaround for now, https://dimjasevic.net/marko/2018/08/15/haskell-ide-the-memory-hog-engine/, thanks @codygman for the link.

PR #1292, opened 2019-06-16
Fix UriCaches being leaked

### LINKED ISSUE #412: Memory characteristics questions
I have a stack project, which defines a bunch of simple executables (~20). I use atom with atom-ide-ui as a client to hie. Now hie's memory usage starts out quite normal (~150MB), but as I open more of the executables source files the memory usage shows an alarming growth rate reaching over 2GB of usage when most of them are opened and after a while it can more than double that again (What activities increase the memory usage is a little unclear, but the usage increases over time (not indefinitely, but enough to force me to allocate more memory for my vm)). If I close some of the open files the usage drops again.

It isn't a problem per se (The benefits are worth the cost), but it does make me wonder:

  - Does the structure of a project affect the memory usage? My executables share a bunch of dependencies, but this does not seem to help hie, since each file adds a whole lot of memory, compared to the amount of information in it.
  - What causes the memory usage to increase like that over time? I know that I am not adding enough new information to the code to warrant it.

bubba:
This fixes UriCaches from being leaked via GhcModuleCaches. This is the result of 72 hours at ZuriHac between 3 people and a lot of time spent in gdb.

Fixes #412, #806, #665
Blog post coming soon

Co-Authored-By: Matthew Pickering <matthewtpickering@gmail.com>
Co-Authored-By: Daniel Gröber <dxld@darkboxed.org>

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
@@ -455,7 +455,7 @@ withIndefiniteProgress t c f = do
     Just wp -> control $ \run -> wp t c (run f)
 
 data IdeState = IdeState
-  { moduleCache :: GhcModuleCache
+  { moduleCache :: !GhcModuleCache
   -- | A queue of requests to be performed once a module is loaded
   , requestQueue :: Map.Map FilePath [UriCacheResult -> IdeM ()]
   , extensibleState :: !(Map.Map TypeRep Dynamic)


alanz:
This leaves me wondering how we strayed from best practice of making all data structure fields strict, unless there is a good reason not to.

bubba:
@wz1000 Running with an lsp-test test-scenario that makes 10 edits:

# Before 🚰
memleak

# After
memfixed

bubba:
@wz1000 I will tidy up the test executable and make a separate PR for it

lorenzo:
Should we just add `StrictData` as default extension to the project?

alanz:
> Should we just add StrictData as default extension to the project?

So long as there is a way to mark fields as lazy, if required.

alanz:
And it is great to have the ability to profile in the project knowledge base.  Well done.

lorenzo:
> So long as there is a way to mark fields as lazy, if required.

Yes it is possible, you can make any field lazy by prepending `~` to it. It has the opposite effect of `!`

lorenzo:
@bubba Would you mind profiling again after adding `StrictData` by default?

bubba:
@lorenzo sure! I’ll see if I can convert my test case into a benchmark actually

Issue #1286, opened 2019-06-09
When Errors are present goto goes to a empty temporary file


Avi-D-coder:
If an error is present in `app/Main.hs` any goto on definition in the same file) will send me to empty file like `/tmp/ghc-mod15112/Main15110-415110-22.hs`

I'm using an update hie with https://github.com/alanz/ghc-mod/pull/20.

fendor:
Does this happen only on this branch? Or did you only notice it on this branch?
Because, iirc, there is an issue about that somewhere.
Also, is there something to reproduce it?

Avi-D-coder:
I noticed it on this branch, I have not had time to reproduce it on master.

Avi-D-coder:
The issue exists on master

wz1000:
This occurs when we have a parsed module, but not a typechecked module, and is due to the update of reverse file map/rfm at https://github.com/haskell/haskell-ide-engine/blob/master/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs#L228

We update the reverse file mapping according to the state of the mapped files when the parsed module is generated. However, the old typechecked module that is retained still contains `SrcSpan`s generated according to the previous state of the mapped files. So, when we use information from the typechecked module to answer queries, we are not able to map the temporary file to the real file.

Hopefully this class of issues will be solved once and forall in GHC 8.8+ by @DanielG fixing GHCs support for memory mapped files. Until then, I am pushing a fix.

PR #1295, opened 2019-06-16
Fix file mapping state when we have a parsed module but not a typechecked module

wz1000:
Fixes https://github.com/haskell/haskell-ide-engine/issues/1286


alanz:
LGTM

mpickering:
I suppose composing these two functions works because if a file isn't mapped then the function acts as the identity. It made me think for a second whether it was correct.

PR #1297, opened 2019-06-18
install.hs: Make all available GHCs in PATH buildable

maoe:
This PR extends the GHC search logic so that `install.hs` can find all available `ghc`s in `$PATH`.

The motivation is that at our company we have our own build of GHCs installed in some directories with custom versioning (e.g. `ghc-8.6.4.0`) and make them available by modifying `$PATH` variable. Currently `stack install.hs cabal-ghcs` cannot find them because it only tries to find GHCs whose version matches one of the versions returned by `getHieVersions`.

The new search logic does the following:
* Find all known GHCs in `$PATH`
    * This is the original behavior.
* Find all available `ghc`s in `$PATH`
* Merge the two sets of paths.
    * If stack-provided GHCs are included, exclude them.
    * If both sets have the same version, the known GHC takes precedence.

This can also be useful to try `hie` with GHC alpha/HEAD etc. You can make the GHC available in `$PATH` and type `stack install.hs cabal-ghc-8.7.20190618` for example.

### DIFF
--- a/install.hs
+++ b/install.hs
@@ -206,10 +206,10 @@ findInstalledGhcs = do
     (reverse hieVersions)
   availableGhcs <- getGhcPaths
   return
-    -- filter out stack provided GHCs
-    $ filter (not . isInfixOf ".stack" . snd)
     -- nub by version. knownGhcs takes precedence.
-    $ nubBy ((==) `on` fst) (knownGhcs ++ availableGhcs)
+    $ nubBy ((==) `on` fst)
+    -- filter out stack provided GHCs
+    $ filter (not . isInfixOf ".stack" . snd) (knownGhcs ++ availableGhcs)
 
 cabalBuildHie :: VersionNumber -> Action ()
 cabalBuildHie versionNumber = do
--- a/install.hs
+++ b/install.hs
@@ -19,6 +19,7 @@ import           Control.Monad.Extra                      ( unlessM
                                                           )
 import           Data.Maybe                               ( isJust )
 import           System.Directory                         ( findExecutable
+                                                          , findExecutables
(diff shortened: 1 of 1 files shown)


install.hs
```diff
@@ -193,12 +198,18 @@ configureCabal versionNumber = do
...
+  knownGhcs <- mapMaybeM
+    (\version -> getGhcPathOf version >>= \case
       Nothing -> return Nothing
       Just p  -> return $ Just (version, p)
     )
     (reverse hieVersions)
+  availableGhcs <- getGhcPaths
+  return
+    -- filter out stack provided GHCs
+    $ filter (not . isInfixOf ".stack" . snd)
```
fendor: Why dont we want to have the possibility to use a ghc installed by stack?
I suppose, it contradicts your use-case, e.g. you can never install a custom ghc 8.6.4 if the stack-ghc is always on the path, right?

maoe: > Why dont we want to have the possibility to use a ghc installed by stack?

Because this GHC finding logic is used only for cabal-related targets. If you'd like to use a stack-provided GHC, you'd use a hie-x.x.x target.

> I suppose, it contradicts your use-case, e.g. you can never install a custom ghc 8.6.4 if the stack-ghc is always on the path, right?

That's right. And I found a bug in my code. We should filter stack-provided GHCs _before_ applying `nub`. Otherwise a custom GHC 8.6.4 is always filtered out. I'll push a fix shortly.

maoe: Pushed fdaf48b. We should squash it when merging this PR.

fendor:
waiting for approval of @power-fungus

PR #1299, opened 2019-06-18
Use ghc-mod without memory leak

alanz:
As fixed by @bubba in https://github.com/alanz/ghc-mod/pull/21


bubba:
I’d like to change the putMVar calls to non blocking ones just in case before this is merged, will do this tonight!

alanz:
> I’d like to change the putMVar calls to non blocking ones just in case before this is merged, will do this tonight!

Why? There should be no contention for the MVar, if there is, a hangup gives us some kind of warning.

lorenzo:
I'll merge as this is already an improvement!

PR #1304, opened 2019-06-20
Fix build for Windows 7

jneira:
* Actually the project cant be built in windows 7 due a bug in the package `unix-time` (via `hoogle` -> `wai-logger` -> `fast-logger`)
* The issue was resolved in the package: see https://github.com/kazu-yamamoto/unix-time/issues/51
* The first version of the package with the fix was `0.4.7`

This pr update the resolver for `stack-yaml` and `stack-8.6.5.yaml` to get one with the patched version and add `unix-time-0.4.7` as exta dependency for the rest of stack config files.

### DIFF
--- a/stack-8.2.2.yaml
+++ b/stack-8.2.2.yaml
@@ -33,6 +33,8 @@ extra-deps:
 - rope-utf16-splay-0.3.1.0
 - sorted-list-0.2.1.0
 - syz-0.2.0.0
+# To make build work in windows 7
+- unix-time-0.4.7
 
 flags:
   haskell-ide-engine:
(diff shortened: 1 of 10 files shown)


alanz:
In order to minimise rebuilding for people using hie, we try to only bump resolvers once a month.

Can we maybe just add the extra dep for GHC 8.6.5 and the nightly stack, until they get bumped at the end of the month?

jneira:
@alanz of course! cache is very valuable with so many deps

alanz:
Thanks, LGTM. We can merge this when CI passes

alanz:
And it just struck me we should add

```
unix-time >= 0.4.7
```
to the cabal file somewhere.

jneira:
@alanz maybe it would be not neccesary (other than doing a `cabal update`) cause:
* wai-logger has not upper bound on fast-logger
* fast-logger has not upper bound on unix-time

it doesnt follow pvp but you have the last patched version without hackage revisions in exchange.
I havent tested it though.

alanz:
Fair enough, but I think I will be happier to actively constrain a lower bound on it, as per #1306

Issue #1296, opened 2019-06-17
modifyCache is the source of a space leak


mpickering:
https://github.com/haskell/haskell-ide-engine/blob/ad0d21f14a2b8f770110660942f46a1977b5f42e/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs#L56

Observe here that  the result of applying `f` is not forced so repeated applications will
result in a chain of thunks. 

One way to fix this is:

```diff
commit c3bdae0f7df63c8bdb4f360285feeb7eadff8966 (HEAD -> hie-bios, mpickering/hie-bios)
Author: Matthew Pickering <matthewtpickering@gmail.com>
Date:   Mon Jun 17 23:58:05 2019 +0100

    Fix module leaks

diff --git a/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs b/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
index 3bf08aa5..e40ef193 100644
--- a/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
@@ -61,7 +61,8 @@ import System.IO
 modifyCache :: (HasGhcModuleCache m) => (GhcModuleCache -> GhcModuleCache) -> m ()
 modifyCache f = do
   mc <- getModuleCache
-  setModuleCache (f mc)
+  let x = (f mc)
+  x `seq` setModuleCache x
```

PR #1305, opened 2019-06-20
Fix a memory leak found by mpickering

lorenzo:
closes https://github.com/haskell/haskell-ide-engine/issues/1296

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -54,8 +53,7 @@ import           Haskell.Ide.Engine.PluginsIdeMonads
 modifyCache :: (HasGhcModuleCache m) => (GhcModuleCache -> GhcModuleCache) -> m ()
 modifyCache f = do
   mc <- getModuleCache
-  let !cached = f mc -- Avoid a space leak by forcing the cache calculation
-  setModuleCache cached
+  setModuleCache (f mc)
 
 -- ---------------------------------------------------------------------
 -- | Runs an IdeM action with the given Cradle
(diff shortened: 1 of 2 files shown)


hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -53,7 +54,8 @@ import           Haskell.Ide.Engine.PluginsIdeMonads
 modifyCache :: (HasGhcModuleCache m) => (GhcModuleCache -> GhcModuleCache) -> m ()
 modifyCache f = do
   mc <- getModuleCache
-  setModuleCache (f mc)
+  let !cached = f mc -- Avoid a space leak by forcing the cache calculation
+  setModuleCache cached
```
alanz: Should we not make `setModuleCache` itself strict in its argument?  Otherwise we have to do this in all places it might get used.

lorenzo: Sounds good, I'll do that instead

lorenzo:
@alanz implemented your suggestion

mpickering:
Thanks @lorenzo

tomjaguarpaw:
Hello, I came here as a consequence of reading https://lukelau.me/haskell/posts/leak/.

I would be very interested to know, if the information is available, whether any space improvement was actually seen as a result of merging this PR.  [@bubba's PR](https://github.com/haskell/haskell-ide-engine/pull/1292) had already been merged the day before and should make this PR redundant.  In particular, `modifyTVar'` forces its argument before returning, and subsequent to @bubba's PR forcing the `IdeState` forces the `GhcModuleCache` it contains (that's what making a field strict means).

I have spent, many, many, days investigating the causes of various space leaks.  I'm trying to come up with a set of principles that make it easy to write space-leak-free code.  I would like to test my principles against the space leak that you have discovered.  If this PR did indeed remove a space leak then my principles are incorrect and i need to go back to the drawing board!

tomjaguarpaw:
(Oh, and I got to Luke Lau's post via https://mpickering.github.io/ide/posts/2020-05-08-state-of-haskell-ide.html)

fendor:
The code-base has changed signifitcantly since this PR has been merged.

tomjaguarpaw:
That may be so, but I'm wondering whether there is any residual evidence either way that this PR helped.

fendor:
I am sorry, I dont think we have any hard numbers anymore. Doesnt the blog post state that the leak has been solved?

tomjaguarpaw:
Sure, I've no doubt the space leak has been solved. My question is whether it was by this PR or by [@bubba's earlier PR](https://github.com/haskell/haskell-ide-engine/pull/1292).

bubba:
@tomjaguarpaw these PRs might have been created in parallel - it does seem strange that we enforce strictness here as well. You might want to checkout to both of these revisions and try running the script from the blog post to see if it does make a difference. Otherwise its probably out of an abundance of caution

PR #1316, opened 2019-07-04
Upgrade to haskell-lsp 0.15

lorenzo:


hie-plugin-api/Haskell/Ide/Engine/Ghc.hs
```diff
@@ -45,10 +47,23 @@ import           GHC
...
 
-type Diagnostics = Map.Map Uri (Set.Set Diagnostic)
+newtype Diagnostics = Diagnostics (Map.Map NormalizedUri (Set.Set Diagnostic))
+  deriving (Show, Eq)
+
+instance Semigroup Diagnostics where
+  Diagnostics d1 <> Diagnostics d2 = Diagnostics (d1 <> d2)
+
+instance Monoid Diagnostics where
+  mempty = Diagnostics mempty
```
bubba: Can these be derived?

wz1000: The default monoid instance for Map is probably not what we want. This should be Map.Union with Set.union. the default semi group operation is leftbiased union.

bubba: In any case I think `ToJSON` should be derivable. Although it looks like `NormalizedUri` has no instance for `ToJSON/FromJSON`. This should probably be added in haskell-lsp.

lorenzo:
I'll merge now, please let me know if you wanted me to take any extra actions and I'll open another pull request.

Issue #1309, opened 2019-06-27
unable to build with cabal


power-fungus:
as part of #1221, I need to build `hie` with cabal, which worked fine up to now, but I am getting the following error:

```
$ cabal new-build -w ghc-8.6.5
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: HaRe-0.8.4.1 (user goal)
[__1] rejecting: HaRe:!test (constraint from config file, command line flag,
or user target requires opposite flag selection)
[__1] trying: HaRe:*test
[__2] trying: turtle-1.5.14 (dependency of HaRe *test)
[__3] trying: temporary-1.3 (dependency of turtle)
[__4] next goal: cabal-helper (user goal)
[__4] rejecting: cabal-helper-0.9.0.0 (conflict: temporary==1.3, cabal-helper
=> temporary<1.3 && >=1.2.1)
[__4] rejecting: cabal-helper-0.8.2.0, cabal-helper-0.8.1.2,
cabal-helper-0.8.0.2, cabal-helper-0.8.0.1, cabal-helper-0.8.0.0,
cabal-helper-0.7.3.0, cabal-helper-0.7.2.0, cabal-helper-0.7.1.0,
cabal-helper-0.6.3.1, cabal-helper-0.6.3.0, cabal-helper-0.6.2.0,
cabal-helper-0.6.1.0, cabal-helper-0.6.0.0, cabal-helper-0.5.3.0,
cabal-helper-0.5.1.0, cabal-helper-0.5.0.0, cabal-helper-0.4.0.0,
cabal-helper-0.3.9.0, cabal-helper-0.3.8.0, cabal-helper-0.3.7.0,
cabal-helper-0.3.6.0, cabal-helper-0.3.5.0, cabal-helper-0.3.4.0,
cabal-helper-0.3.3.0, cabal-helper-0.3.2.1, cabal-helper-0.3.2.0,
cabal-helper-0.3.1.0, cabal-helper-0.3.0.0, cabal-helper-0.2.0.0,
cabal-helper-0.1.0.1, cabal-helper-0.1.0.0, cabal-helper-0.8.1.1,
cabal-helper-0.8.1.0, cabal-helper-0.7.0.1, cabal-helper-0.5.2.0 (constraint
from user target requires ==0.9.0.0)
[__4] fail (backjumping, conflict set: cabal-helper, temporary)
Backjump limit reached (currently 2000, change with --max-backjumps or try to
[...]

PR #1317, opened 2019-07-04
Clear out pattern matching and error message of executeCodeActionByName

jneira:
It seems the message had become obsolete, the unique error case is the empty list.
I've added the list ot actions candidates to be run.


jneira:
The cabal build in ci is failing due to #1309

PR #1326, opened 2019-07-08
Remove unnecesary extra-dep unix-time

jneira:
The version with the fix for windows 7 already is in both resolvers.


jneira:
Appveyor build failed due a 443 trying to download `/hackage.fpcomplete.com/package/parallel-3.2.1.1.tar.gz` with resolver `8.4.2`

PR #1334, opened 2019-07-20
Simpler completion

wz1000:
After @mpickering pointed this out yesterday, I noticed that completion was a lot more complicated than it needed to be. Instead of the convoluted wrangling we do with import declarations, all the information we need is in the `GlobalRdrEnv`.

In addition to this, I've separated out the step of getting the type information/snippets for the completion into the LSP `CompletionItemResolve` request. Type information for local variables which
is readily at hand in the `TypecheckedModule` itself is still reported in the original completion request.

A potential performance issue I've tried to fix is that looking up the types of all the possible completions would result in GHC loading a bunch of interface files it didn't need to.

For example, consider a module that imports `Prelude`, but doesn't use any functions from `GHC.Enum`(like `enumFrom`). Normally, GHC will not load the interface file for `GHC.Enum`. However, when we try to get the type of the `enumFrom` completion, this forces GHC to go and actually load the interface file for that module. 

~~I've implemented the function we used to look up the type information to just return `Nothing` in this case and not try to load the `GHC.Enum` interface file. You can see this in action in the `"does not pull in unnecessary modules until needed"` test case I've added.~~

~~The above behaviour is up for discussion, and I will re-implement the old/default behaviour for `getTypeForName` depending on the consensus.~~ 

The interface file will now be pulled in when the client requests to "resolve" the completion.

The behavioural changes due to this PR are:

- Completion items need to be resolved to get their type information/snippets.
- ~~The types/snippets for some functions from modules that haven't been used previously may not be available, as described above.~~

### DIFF
--- a/src/Haskell/Ide/Engine/Support/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Support/HieExtras.hs
@@ -41,7 +41,6 @@ import           Data.IORef
 import qualified Data.List                                    as List
 import qualified Data.Map                                     as Map
 import           Data.Maybe
-import           Data.Monoid                                  ( (<>) )
 import qualified Data.Text                                    as T
 import qualified Data.Text.IO                                 as T
 import           Data.Typeable
--- a/src/Haskell/Ide/Engine/Support/HieExtras.hs
+++ b/src/Haskell/Ide/Engine/Support/HieExtras.hs
@@ -8,6 +8,7 @@ module Haskell.Ide.Engine.Support.HieExtras
   ( getDynFlags
   , WithSnippets(..)
   , getCompletions
+  , resolveCompletion
   , getTypeForName
   , getSymbolsAtPoint
   , getReferencesInDoc
@@ -26,7 +27,7 @@ module Haskell.Ide.Engine.Support.HieExtras
   ) where
 
 import           ConLike
-import           Control.Lens.Operators                       ( (^?), (?~), (&) )
+import           Control.Lens.Operators                       ( (.~), (^.), (^?), (?~), (&) )
 import           Control.Lens.Prism                           ( _Just )
 import           Control.Lens.Setter ((%~))
 import           Control.Lens.Traversal (traverseOf)
@@ -62,7 +63,8 @@ import           Haskell.Ide.Engine.Context
 import           Haskell.Ide.Engine.MonadFunctions
 import           Haskell.Ide.Engine.MonadTypes
 import           Haskell.Ide.Engine.PluginUtils
-import qualified Haskell.Ide.Engine.Support.Fuzzy              as Fuzzy
+import qualified Haskell.Ide.Engine.Support.Fuzzy             as Fuzzy
+import qualified Haskell.Ide.Engine.Plugin.Hoogle             as Hoogle
 import           HscTypes
 import qualified Language.Haskell.LSP.Types                   as J
 import qualified Language.Haskell.LSP.Types.Lens              as J
@@ -74,9 +76,11 @@ import           Outputable                                   (Outputable)
 import qualified Outputable                                   as GHC
 import           Packages
 import           SrcLoc
-import           TcEnv
 import           Type
 import           Var
+import           Unique
+import           UniqFM
+import           Module hiding (getModule)
 
 -- ---------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


src/Haskell/Ide/Engine/Support/HieExtras.hs
```diff
@@ -194,16 +273,25 @@ safeTyThingId (AConLike (RealDataCon dc)) = Just $ dataConWrapId dc
 safeTyThingId _                           = Nothing
 
 -- Associates a module's qualifier with its members
-type QualCompls = Map.Map T.Text [CompItem]
+newtype QualCompls = QualCompls { getQualCompls :: Map.Map T.Text [CompItem] }
+
+instance Semigroup QualCompls where
+  (QualCompls a) <> (QualCompls b) = QualCompls $ Map.unionWith (++) a b
+
+instance Monoid QualCompls where
```
lorenzo: Needs a mappend implementation for ghc 8.2

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -649,22 +648,11 @@ reactor inp diagIn = do
...
+              callback res = do
+                let rspMsg = Core.makeResponseMessage req $ res
                 reactorSend $ RspCompletionItemResolve rspMsg
-              hreq = IReq tn (req ^. J.id) callback $ runIdeResultT $ case mquery of
-                        Nothing -> return Nothing
-                        Just query -> do
-                          result <- lift $ lift $ Hoogle.infoCmd' query
-                          case result of
-                            Right x -> return $ Just x
-                            _ -> return Nothing
```
lorenzo: Nice cleanup!

src/Haskell/Ide/Engine/Support/HieExtras.hs
lorenzo: The solution is to simply add here
`mappend = (<>)`

src/Haskell/Ide/Engine/Support/HieExtras.hs
```diff
@@ -194,16 +273,25 @@ safeTyThingId (AConLike (RealDataCon dc)) = Just $ dataConWrapId dc
...
   { allModNamesAsNS :: [T.Text]
   , unqualCompls :: [CompItem]
   , qualCompls :: QualCompls
   , importableModules :: [T.Text]
-  , cachedExtensions :: [T.Text]
   } deriving (Typeable)
 
+-- The supported languages and extensions
+languagesAndExts :: [T.Text]
+languagesAndExts = map T.pack GHC.supportedLanguagesAndExtensions
```
bubba: This is great, was this previously cached for a reason? Helpful reading by the way https://stackoverflow.com/a/3951092/1320911

bubba:
I agree with @Lorenzo, I think the completion logic is significant enough to split out

src/Haskell/Ide/Engine/Support/HieExtras.hs
lorenzo: This basically achieve the same as caching it. When I added it to the cache I just wanted to avoid the text packing on each use.

wz1000:
@bubba, @lorenzo any comments on the type lookup behaviour? If I restore the old behaviour, because looking up the type is anyway done in the `resolveCompletion` step, the interface file would only be loaded in when the user goes over the relevant completion (or whenever the client decides to resolve the completion).

lorenzo:
@wz1000 I don’t think I fully understand the consequences of changing the behavior, superficially this looks ok to me. What do you think the consequences of not having the old behavior might be?

wz1000:
This is how the new behaviour manifests itself in vscode:

Before GHC has loaded the GHC.Enum interface file, the completion item looks like this:

before

After I add a usage of `enumFrom`, GHC is forced to load the `GHC.Enum` interface file, we get the type for `fromEnum` also. When the completion is selected, we will also a "snippet" for the argument.

after

If I revert to the old behaviour, just having the `fromEnum` entry selected will make GHC load the interface file for `GHC.Enum`, and we will get the type of `fromEnum` in the completion window even if we don't have a usage of `enumFrom` elsewhere in the file.

alanz:
I think doing the expensive work in the resolve step is the right way to go. As I understand it, this is precisely why the resolve step exists.

Completion needs to be a fast/unintrusive operation, so reducing the work done is a good step.

alanz:
I am happy with this approach, and agree we should move the stuff into its own file.

But for traceability it would be better to land this one as a change to HieExtras, so we can see the diff, and then add a second administrative change to move the code around.

bubba:
@wz1000 relying on Hoogle for type information felt a bit dodgey, so I am happy that this is the correct way, even if it is a bit slower to resolve, which happens "asynchronously" so I am not concerned.

But in your example posted, does this mean that completion won't show any types until that interface file is loaded by means of it being used inside the document? Or is a completion request alone able to force it to load the interface file?

wz1000:
@bubba 
> But in your example posted, does this mean that completion won't show any types until that interface file is loaded by means of it being used inside the document?

This. But I guess it won't be too bad to load the interface file on a completion resolve request.

Btw, for a comparison of memory usage before and after this patch, see [BEFORE](https://htmlpreview.github.io/?https://github.com/wz1000/haskell-ide-engine/blob/simpler-completion-mem/hie.eventlog.old.html) and [AFTER](https://htmlpreview.github.io/?https://github.com/wz1000/haskell-ide-engine/blob/simpler-completion-mem/hie.eventlog.html)

The test is loading the file 20 times, then loading it 20 times asking for completions every time. The bump in the middle in BEFORE is when we start asking for completions.

wz1000:
I've changed it so that the interface file is loaded in when the client asks to resolve the completion, and the type/snippet is available for every symbol in the completion list.
This should be ready to merge now.

PR #1338, opened 2019-07-22
lift required stack-version to 2.1.1

### LINKED ISSUE #1328: Thread blocked in an STM transaction
Hi, I'm an Arch user (relevant neofetch + glibc? https://owo.whats-th.is/5zu65ff.png) and I've installed cabal using the bootstrap script. I then used cabal to install stack, and stack is now trying to install HIE. However, when it gets around to unpacking ghc-tinfo6-8.6.4 (and once even got that and went on to installing but failed with the same error), it errs with `thread blocked indefinitely in an STM transaction` and error code 1. This is done with `stack install.hs`. It always manages to download ghc-tinfo6, but it never manages to actually use it.
I'm new to Haskell as a whole and for sure don't know much about the ecosystem, so I might be missing something but I'm fairly sure I've been doing it correctly so far.

power-fungus:
Closes #1328. 

Is lifting the stack-version to 2.1.1 ok? Do we need still to support stack 1.9 ?

### DIFF
--- a/install/src/Stack.hs
+++ b/install/src/Stack.hs
@@ -86,7 +86,7 @@ stackExeIsOldFailMsg stackVersion =
     ++ "Please run `stack upgrade` to upgrade your stack installation"
 
 requiredStackVersion :: RequiredVersion
-requiredStackVersion = [1, 9, 3]
+requiredStackVersion = [2, 1, 1]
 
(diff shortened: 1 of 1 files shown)


power-fungus:
I would like to merge this, if there are no objections.

PR #1346, opened 2019-07-27
Towards testing with cabal

DanielG:
Currently the tests force using `stack`, but HIE should be an equal opportunity build system employer :)

Remember to read commit messages for details.

---

TODO:
- [ ] Also test against cabal cradle

### DIFF
--- a/test/utils/TestUtils.hs
+++ b/test/utils/TestUtils.hs
@@ -160,7 +160,7 @@ stackYaml =
 #endif
 
 logFilePath :: String
-logFilePath = "functional-hie-" ++ stackYaml ++ ".log"
+logFilePath = "hie-" ++ stackYaml ++ ".log"
 
 -- | The command to execute the version of hie for the current compiler.
 --
--- a/test/utils/TestUtils.hs
+++ b/test/utils/TestUtils.hs
@@ -163,13 +163,12 @@ logFilePath :: String
 logFilePath = "functional-hie-" ++ stackYaml ++ ".log"
 
 -- | The command to execute the version of hie for the current compiler.
--- Make sure to disable the STACK_EXE and GHC_PACKAGE_PATH environment
--- variables or else it messes up -- ghc-mod.
--- We also need to unset STACK_EXE manually inside the tests if they are
--- run with `stack test`
+--
+-- Both @stack test@ and @cabal new-test@ setup the environment so @hie@ is
+-- on PATH. Cabal seems to respond to @build-tool-depends@ specifically while
+-- stack just puts all project executables on PATH.
 hieCommand :: String
-hieCommand = "stack exec --no-stack-exe --no-ghc-package-path --stack-yaml=" ++ stackYaml ++
-             " hie -- -d -l test-logs/" ++ logFilePath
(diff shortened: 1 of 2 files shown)


test/utils/TestUtils.hs
```diff
@@ -159,9 +159,6 @@ stackYaml =
   "stack-8.0.2.yaml"
 #endif
 
-logFilePath :: String
-logFilePath = "functional-hie-" ++ stackYaml ++ ".log"
```
power-fungus: This function is required by some tests. See ci-build

DanielG: A grep over the entire source doesn't show any other users of this definition. I don't know what you mean?

DanielG: Ugh, nvm. I'm an idiot :)

power-fungus:
I actually tried to run the tests with cabal and a large number of them failed. See [my azure-ci build](https://dev.azure.com/haskell-ide-engine/haskell-ide-engine-power-fungus/_build/results?buildId=140) for a comparison between the test-runs using `cabal` and `stack`.

I would say that currently executing the tests with stack is unsupported as well. @jneira, @fendor and I tried our best in the last months to get the tests working using a repeatable script that starts from a clean os-image. We did not succeed so far.

DanielG:
I ran all the tests with cabal on my machine and they all worked perfectly after all the necessary `hoogle` setup and what have you at least. But I agree the tests are very fiddly at the moment.

DanielG:
Looks like your CI builds with cabal were all failing because of caba-helper $libexec stuff. Not sure why that was working for me..

power-fungus:
The tests work in specialized environments that nobody knows how to create. The azure-ci scripts help to find a way to create such an environment reliably.

Do the tests still work on your machine after cleaning `~/.cabal`, running `git clean -Xdf` in the hie-repo and removing all `hie` and `cabal-helper` executables, unsetting ?

DanielG:
Indeed it looks like I had an hie executable in `~/.local/bin` from running `stack install` still. I don't understand why it makes a difference as new-test does put the cabal installed hie executable on PATH before `~/.local` though. Strange.

power-fungus:
> Strange.

This describes the `hie` runtime-dependencies perfectly well. To this date, I still have no idea what is really needed and why `hie` fails on certain projects or setups. The azure-ci and `install.hs` projects were efforts towards a deeper understanding of the issue. I have learned some rules, but I am no less confused.

The [curent build](https://dev.azure.com/haskell-ide-engine/haskell-ide-engine-power-fungus/_build/results?buildId=140) shows that the tests also fail with `stack`. Some environment is no quite right. :cry:

fendor:
So, this looks pretty nice now, right? I would be so glad if we could remove stack as a test dependency :)

jneira:
@fendor If i have ubderstood correctly tests started to fail in the last @DanielG, right?

DanielG:
The tests did fail once they were using the `hie` executable built by cabal. I still think we should merge this. It's at least making some progress towards getting cabal working.

I was going to include more stuff here but got side-tracked a bit by cabal-helper work so I think we should just merge this now.

PR #1355, opened 2019-08-05
Made hlint dependency properly depend on version of ghc.

LinuxUser404:
hie does not compile with earlier versions of hlint, so I bumped the version in the cabal file instead of stack file. I did not touch stack-8.2.2.yaml though it seems that build is broken anyway. As a bonus the whole thing builds with cabal v2-build.

### DIFF
--- a/stack-8.4.2.yaml
+++ b/stack-8.4.2.yaml
@@ -10,8 +10,8 @@ extra-deps:
 - ./submodules/ghc-mod/core
 - ./submodules/ghc-mod/ghc-project-types
 
-- brittany-0.12.0.0
 - base-compat-0.9.3
+- brittany-0.12.0.0
 - cabal-plan-0.3.0.0
 - constrained-dynamic-0.1.0.0
 - floskell-0.10.0
@@ -23,6 +23,7 @@ extra-deps:
 - haskell-lsp-types-0.15.0.0
 - haskell-src-exts-1.21.0
 - haskell-src-exts-util-0.2.5
+- hlint-2.2.2
 - hoogle-5.0.17.9
 - hsimport-0.10.0
 - lsp-test-0.6.0.0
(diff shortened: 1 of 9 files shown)


LinuxUser404:
Hm, looks like I am missing something...

fendor:
I am not sure this change can succeed. If you force the version in `haskell-ide-engine.cabal` to be `==2.2.2`, how could hie-8.2.2 still be built, since it uses the same cabal file and doesnt work with `hlint == 2.2.2`?
Thus, the version range of the cabal file.

LinuxUser404:
My change was a bit premature indeed... I initially thought the ghc-8.2 would fail regardless because
https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
imports HLint4. I see now that it is conditional depending on the version of ghc. My cabal build was failing with 2.1.22 because of the same issue. Okay, so the problem I was trying to solve is failing builds with ghc-8.4 and above using old hlint. I think my last next commit should properly solve it.

stack-8.4.2.yaml
```diff
@@ -10,8 +10,8 @@ extra-deps:
 - ./submodules/ghc-mod/core
 - ./submodules/ghc-mod/ghc-project-types
 
-- brittany-0.12.0.0
 - base-compat-0.9.3
+- brittany-0.12.0.0
```
fendor: Can you maybe revert this change? Seems not necessary.

stack-8.6.5.yaml
```diff
@@ -21,8 +21,8 @@ extra-deps:
 - haskell-lsp-types-0.15.0.0
 - haskell-src-exts-1.21.0
 - hlint-2.2.2
-- hsimport-0.10.0
 - hoogle-5.0.17.9
+- hsimport-0.10.0
```
fendor: Same here

LinuxUser404:
Done

Issue #1327, opened 2019-07-09
`stack install.hs build-all` installs cabal in $HOME/.local/bin


alexanderkjeldaas:
I was a bit surprised that `stack install.sh build-all` installs a new `cabal` into `$HOME/.local/bin`.  That seems to go a bit beyond what that command is supposed to do.

fendor:
Would  you prefer that the command fails if no appropriate cabal version can be found?

power-fungus:
Hopefully, this can be removed in the future, but currently the cabal Binary is a runtime dependency of hie

fendor:
Still, instead of automatically installing a cabal version we need, it might be better to show an error message and let the user decide how to resolve the problem with cabal.

jneira:
Another approach could be (optionally) download runtime dependencies at runtime, when precompiled binaries are available and the tool is not present in path or custom location (like stack does with ghc)

power-fungus:
I am against downloading dependencies at runtime since this increases startup-latency even further.

The current state is not pretty, but the current plugins do not allow a better solution. For this reason, I would like to close this issue.

jneira:
> I am against downloading dependencies at runtime since this increases startup-latency even further.

Mmm, totally agree with that, but if we want at some point[ offer hie as a precompiled binary](https://github.com/haskell/haskell-ide-engine/issues/1068) itself (imho essential to increase the user base) we have to solve this anyway.

I think initiated haskell devs will have already `cabal`, usually in `$PATH`, so they will not have any latency. If we let customize cabal location, users could choose their fav cabal version suited for hie. Download cabal would be the last resort and should be notified to user, in a second plane and fail gracefully with no internet connection.

It actually is 40-70 Mb so i think the delay will not be terrible, right?

fendor:
Well, on a 1Mbit/s connection that is more than a minute. However, the bigger problem I see is, if you use an older cabal and a newer one on the same project, a lot of stuff breaks. E.g. the user uses the pre shipped with their distro cabal version `2.4` or something like that on their project with `cabal v2-build`. However, hie makes use of the latest and greatest development tools in haskell, thus it installs `cabal 3.0.0.0`, assuming it had been released already. The build - caches will have different contents and formats and ultimately cause a lot of package rebuilds and unforeseen breakage. 
Possible solutions are to force every user of hie to use cabal 3.0.0.0 or be aware of already installed cabal versions and act accordingly.

power-fungus:
[...]

PR #1368, opened 2019-08-24
Remove cabal check from stack builds

ollef:
Stack should provide its own cabal, so I don't think the system cabal is
relevant here. This worked for me despite the check failing.

### DIFF
--- a/install/src/Cabal.hs
+++ b/install/src/Cabal.hs
@@ -66,7 +66,7 @@ installCabal = do
   -- install `cabal-install` if not already installed
   unless cabalExeOk $ execStackShake_ ["install", "cabal-install"]
 
--- | check `stack` has the required version
+-- | check `cabal` has the required version
 checkCabal :: Action ()
 checkCabal = do
   cabalVersion <- getCabalVersion
@@ -92,7 +92,7 @@ cabalInstallNotSuportedFailMsg =
     ++ "If this system has been falsely identified, please open an issue at:\n\thttps://github.com/haskell/haskell-ide-engine\n"
 
 
--- | Error message when the `stack` binary is an older version
+-- | Error message when the `cabal` binary is an older version
 cabalInstallIsOldFailMsg :: String -> String
 cabalInstallIsOldFailMsg cabalVersion =
   "The `cabal` executable is outdated.\n"
--- a/install/src/HieInstall.hs
+++ b/install/src/HieInstall.hs
@@ -102,7 +102,6 @@ defaultMain = do
       (\version -> phony ("stack-hie-" ++ version) $ do
         need ["submodules"]
         need ["check-stack"]
-        need ["cabal"]
         stackBuildHie version
         stackInstallHie version
       )


power-fungus:
Cabal is regarded a runtime dependency of hie and that has nothing to do with the functionality of stack.

I can try to reproduce the issue when installing hie without having a cabal binary. Until then, this pr should not me merged.

power-fungus:
Cabal is regarded a runtime dependency of hie and that has nothing to do with the functionality of stack.

I can try to reproduce the issue when installing hie without having a cabal binary. Until then, this pr should not me merged.

ollef:
If so, it seems like it makes more sense to perform that check at runtime, since the installed version of cabal could change after installing hie, and since you shouldn't need it to compile hie with stack.

The problem that I ran into that this fixed for me was that hie reported my system's cabal version, 2.4.0.0, as insufficient.

jneira:
Mmm, some context: that was added in #1344 to close #1327, if i am not wrong. Before that, installing hie triggered installing globally the cabal version it needed, overwriting the old one.

In #1327 we discussed that maybe we could remove the cabal check/install when installing hie, or at least make hie use a private version of cabal. But i am afraid that we can't simply remove it without some additional work.

@ollef have you considered upgrade your default cabal to `2.4.1.0`?

ollef:
Cheers for providing context. Is cabal necessary for using hie with stack projects? It seems to be working here without the correct version.

Again, I think runtime dependencies, especially when optional, are better checked at runtime.

I could upgrade, yeah.

power-fungus:
The hie runtime is very tricky. I work on getting the correct checks that hie is able to perform a self test. However, I had little success so far. 

As far as I know, cabal is indeed reqired, even for stack based projects.

jneira:
Well, i've tested to open a stack based project *removing* cabal entirely from path and hie *seems* to work (open the project without errors), emitting a warning about not finding the cabal executable.
But i've not tested that the main features continue working. 

@ollef have you tested that main features continue working with your cabal version? I guess that being a minor version mismatch (`2.4.1.0`vs `2.4.0.0`) they'll probably work but there is no the general case.

maybe we could add a flag to avoid/force the cabal installation (`--install-cabal` and `--no-install-cabal` or something alike), keeping the install by default for now? I would keep the check as a warning (like at runtime) that hie needs cabal (with a specific version) to work in all cases.

ollef:
Ah, interesting, so there's _something_ looking for cabal even in stack based projects. I didn't notice any warnings about my specific version of cabal, however.

As far as I can tell hie works with my cabal version, but I've only tested some of the basics.

A warning at compile-time on encountering the wrong version seems like a sensible approach to me.

mpickering:
I have no idea why `hie` depends on `cabal`. It seems unecessary.. my branch doesn't depend on it, unless you are already using `cabal`.

jneira:
I've tried stack without cabal in path a little bit more and hie seems to behave identically for stack based projects.
For cabal projects it shows this log:
```
info: Found Cabal project at: D:\dev\ws\haskell\cabal-test
Using hie version: Version 0.12.0.0, Git revision ff9a5b2e85a0910205cb781bf630503efe6bd227 (2995 commits) x86_64 ghc-8.6.5
info: 'cabal' executable wasn't found, trying next project type
info: Found no other project type, falling back to plain GHC project
```
I guess that some features will be not present.
 
So i am for mimic the runtime behavior in the build and keep the warn in the build  if cabal is not present or the version is not adequate. At least there would be a flag to avoid the install.

power-fungus:
If others have no objection to this, I'm fine with dropping the check and merge this PR.

PR #1369, opened 2019-08-25
Split out completion from HieExtras

### LINKED ISSUE #1340: Split out completion logic
As per https://github.com/haskell/haskell-ide-engine/pull/1334 , the completion logic in HieExtras is getting pretty big. This should be split out into its own module

bubba:
Closes #1340


alanz:
If this is just a rearrangement of code I am fine with it.

src/Haskell/Ide/Engine/Transport/LspStdio.hs
```diff
@@ -41,6 +41,7 @@ import qualified GhcModCore               as GM ( loadMappedFileSource, getMMapp
 import           Haskell.Ide.Engine.Config
 import qualified Haskell.Ide.Engine.Ghc   as HIE
 import           Haskell.Ide.Engine.LSP.CodeActions
+import           Haskell.Ide.Engine.LSP.Completions
```
lorenzo: It would be nice if you imported this qualified. I personally find it easier to read when using the functions

alanz: @lorenzo this should be *only* a code move.  If we want to change the code that moved, that should be a separate PR.

Otherwise it is too hard to follow the trail of what happened. In my opinion.

bubba:
Argh appveyor failed because of a chocolatey error. I'm going to ignore this one

src/Haskell/Ide/Engine/Transport/LspStdio.hs
lorenzo: Well, the code was being used qualified before this move, so I think keeping the style is desirable.

bubba: Added it back!

PR #1372, opened 2019-08-27
Do not traverse into Generated bindings when creating TypeMap

fendor:
Backported from https://github.com/mpickering/haskell-ide-engine/commit/13daf07124770a2643b63f8d581f88cacdea7ce1 by cherry-picking.


mpickering:
Looks good to me. Even if the patch isn't perfect, at least for now this reduces memory usage a lot of modules with a lot of derived code.

PR #1375, opened 2019-08-29
Add different Contexts for Module, import etc...

fendor:
Add more Contexts for the function `getContext`.
Goal: Provide information with the Contexts that otherwise need to be reparsed or are simply not available for completions.
So far, this pr provides context for the module header `module Foo.Bar where`, export list from the module and imports.

Currently, the performance may take a hit when there are a lot of imports, but I dont know what 'a lot' might be. 
 
Also, adds tests for the different contexts.

PR aims to not change the existing completion system but rather expand available contexts.
Completion system should be changed in a different PR.

Contexts that are planned to be implemented here:
* [x] Module Declaration, extracts module name
* [x] Module Declaration, export list
* [x] Imports, import name
* [x] Imports, import list, etc...  
* ~[x] Class declaration~
* ~[x] Instance declaration~
* ~[ ] Language pragma (Seems impossible?)~

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Context.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Context.hs
@@ -1,6 +1,8 @@
+{-# LANGUAGE LambdaCase #-}
 module Haskell.Ide.Engine.Context where
 
 import Data.Generics
+import Data.List (find)
 import Language.Haskell.LSP.Types
 import GHC
 import qualified GhcModCore as GM (GhcPs) -- for GHC 8.2.2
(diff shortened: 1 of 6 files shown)


hie-plugin-api/Haskell/Ide/Engine/Context.hs
```diff
@@ -13,27 +15,86 @@ import Haskell.Ide.Engine.PluginUtils
 -- smarter code completion
 data Context = TypeContext
              | ValueContext
+             | ModuleContext String
+             | ImportContext String
+             | ImportListContext String
```
fendor: What other information is relevant? List of imported symbols? 
Do we have to be careful to not clone the GHC/ghc-lib/haskell-src-exts API?

hie-plugin-api/Haskell/Ide/Engine/Context.hs
```diff
@@ -13,27 +15,86 @@ import Haskell.Ide.Engine.PluginUtils
...
+  = Just (ModuleContext (moduleNameString modName))
+
+  | Just (L (RealSrcSpan r) _) <- exportList
+  , pos `isInsideRange` r
+  = Just ExportContext
+
+  | Just ctx <- everything (<|>) (Nothing `mkQ` go `extQ` goInline) decl
+  = Just ctx
+
+  | Just ctx <- asum $ map importGo imports
```
fendor: I restrained myself to use `everything` here, since we do not wish to traverse every info in import lists, this seemed to be more efficient to me, but I do not understand `Data` enough, if it actually does what I would expect it to.

hie-plugin-api/Haskell/Ide/Engine/Context.hs
```diff
@@ -13,27 +15,86 @@ import Haskell.Ide.Engine.PluginUtils
...
+        go (L (GHC.RealSrcSpan r) GHC.ValD {})
           | pos `isInsideRange` r = Just ValueContext
           | otherwise = Nothing
+        go (L (GHC.RealSrcSpan r) GHC.InstD {})
+          | pos `isInsideRange` r = Just InstanceContext
+          | otherwise = Nothing
+        go (L (GHC.RealSrcSpan r) GHC.DerivD {})
+          | pos `isInsideRange` r = Just DerivingContext
+          | otherwise = Nothing
+        go (L (GHC.RealSrcSpan r) (GHC.TyClD _ GHC.ClassDecl {}))
```
fendor: How should we handle GHC AST version differences?
A quick work around would be, to add another view pattern to `Compat`.
However, it seems like this would grow out of control quickly. Should we rather use something more version independent, such as ghc-lib or haskell-src-exts?

hie-plugin-api/Haskell/Ide/Engine/Context.hs
fendor: Now uses `something` is the same as `everything (<|>)` afaik.

hie-plugin-api/Haskell/Ide/Engine/Context.hs
fendor: Instead of trying to anticipate every possible use-case, I would leave it like that and required context information should be added as needed.

fendor:
Only keep contexts that I have specific use-case in mind for.
E.g. ClassContext and InstanceContext might overshadow the more specific TypeContexts that are currently more helpful.
In a follow-up PR, I aim to add Completions for Import Lists and tidying up `Completions.hs`. Especially the backtick aware completions are a mess.

Issue #1376, opened 2019-08-31
Support Cabal 3.0.0.0


Infinisil:
Recently [Cabal 3.0.0.0](https://hackage.haskell.org/package/Cabal-3.0.0.0/changelog) and [cabal-install 3.0.0.0](https://hackage.haskell.org/package/cabal-install-3.0.0.0/changelog) was released, bringing along the following relevant changes I could find:
- Deprecation of old symbols in https://github.com/haskell/cabal/pull/6022
  This change is relevant because [the version of `cabal-helper` HIE is using](https://github.com/alanz/cabal-helper/tree/eafed5e8c1d82b8daa35775b52361132f2e70261) still references these old symbols. [upstream `cabal-helper`](https://github.com/DanielG/cabal-helper) doesn't reference them anymore.
- New-style builds are now the default, meaning `cabal configure` is not using old-style builds anymore. This is a problem because both `cabal-helper` and `ghc-mod` have the assumption that old-style builds are the default in that they use `cabal configure` and such, and not `cabal v1-configure`. 
  Relevant code in `ghc-mod`: https://github.com/alanz/ghc-mod/blob/910887b2c33237b703417ec07f35ca8bbf35d729/core/GhcMod/CabalHelper.hs#L192-L197.
  Relevant code in `cabal-helper`: https://github.com/alanz/cabal-helper/blob/eafed5e8c1d82b8daa35775b52361132f2e70261/src/CabalHelper/Compiletime/Compile.hs#L452

As far as I can see, this results in HIE being broken for everybody using Cabal 3.0.0.0

### Possible solutions

- Patch `ghc-mod` and `cabal-helper` to use `v1-` commands explicitly
- One of
  - Patch `cabal-helper` to not use the deprecated symbols anymore, by looking at what upstream `cabal-helper` did to remove them. This might be simple or it might not be simple at all
  - Update `cabal-helper` to the upstream version, this is what https://github.com/haskell/haskell-ide-engine/pull/1245 tries to do, but there seems to be some problems
  - I also think [hie-bios](https://github.com/haskell/haskell-ide-engine/pull/1126) by @mpickering would solve this, though this doesn't look like it's going to be merged quickly

Ping @fendor @bubba @alanz @mpickering @DanielG 

Related issues:
- https://github.com/Infinisil/all-hies/issues/22
- https://github.com/haskell/haskell-ide-engine/issues/1373
- https://github.com/NixOS/nixpkgs/issues/67761

[...]

PR #1379, opened 2019-09-04
Add support for building with cabal-3.0.0.0

### LINKED ISSUE #1250: Cannot do cabal 3.0 build for 8.6.5
Trying to install `cabal-hie-8.6.5` gives me the following error instead:

```
Warning: Failed to build documentation for haddock-api-2.22.0 (which is
required by test:wrapper-test from haskell-ide-engine-0.9.0.0, test:unit-test
from haskell-ide-engine-0.9.0.0 and others).
# stack (for cabal-hie-8.6.5)
# cabal (for cabal-hie-8.6.5)
cabal: unrecognized 'new-install' option
`--symlink-bindir=/home/annaia/.local/bin'
  at src/Development/Shake/Internal/Args.hs:81:69-79:
* Depends on: cabal-hie-8.6.5
* Raised the exception:
user error (Development.Shake.command_, system command failed
Command: cabal new-install --write-ghc-environment-files=never --symlink-bindir=/home/annaia/.local/bin exe:hie --overwrite-policy=always
Exit code: 1
Stderr:
cabal: unrecognized 'new-install' option
`--symlink-bindir=/home/annaia/.local/bin'
)
```

I have ghc-8.6.5 and cabal 3.0 installed locally so I'm presuming this is because somehow the stack temp cabal isn't the right version?

Anyway to fix this or is there something I'm missing?

jneira:
* Use `symlink-bindir` or `install-dir` depending on cabal version
* Add support for building in windows, using `install-method=copy`
* Use `v2-` prefix instead `new-` (available since `cabal-2.4.1.0`)

Fixes #1250

### DIFF
--- a/install/src/Cabal.hs
+++ b/install/src/Cabal.hs
@@ -17,7 +17,6 @@ import           Print
 import           Env
 import           Stack
 
-
 execCabal :: CmdResult r => [String] -> Action r
 execCabal = command [] "cabal"
 
@@ -26,8 +25,8 @@ execCabal_ = command_ [] "cabal"
 
 cabalBuildData :: Action ()
 cabalBuildData = do
-  execCabal_ ["new-build", "hoogle"]
-  execCabal_ ["new-exec", "hoogle", "generate"]
+  execCabal_ ["v2-build", "hoogle"]
+  execCabal_ ["v2-exec", "hoogle", "generate"]
 
 cabalBuildHie :: VersionNumber -> Action ()
 cabalBuildHie versionNumber = do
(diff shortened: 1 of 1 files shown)


jneira:
Oh, i've disabled tests when building `hie`, it makes the installation takes longer to finish and afaik it is not needed for install it.

install/src/Cabal.hs
```diff
@@ -79,20 +87,24 @@ getCabalVersion = trimmedStdout <$> execCabal ["--numeric-version"]
 
 -- TODO: this restriction will be gone in the next release of cabal
```
fendor: Remove or edit todo

power-fungus:
This seems nice! Thanks

install/src/Cabal.hs
```diff
@@ -37,18 +36,27 @@ cabalBuildHie versionNumber = do
...
   localBin <- getLocalBin
-  execCabal_
-    [ "new-install"
+  cabalVersion <- getCabalVersion
+  
+  let isCabal3 = checkVersion [3,0,0,0] cabalVersion
+      installDirOpt | isCabal3 = "--installdir"
+                    | otherwise = "--symlink-bindir"
+      installMethod | isWindowsSystem && isCabal3 = ["--install-method=copy"]
+                    | otherwise = []
```
fendor: maybe more explicit `["--install-method=symlink"]`?

install/src/Cabal.hs
jneira: Ooops, you are right

install/src/Cabal.hs
jneira: Mmm, i think we shouldn't force users of other os's use symlinking cause maybe some of them will prefer copy them instead (modifying the cabal global config)

masaeedu:
Does this fix https://github.com/haskell/haskell-ide-engine/issues/1376 as well?

fendor:
Nope, this is just for building the executables

masaeedu:
Ah, my bad.

Issue #1391, opened 2019-09-21
Error in windows detected by the functional test suite


jneira:
The test suite fails in my windows 10 msys2 console with those errors:

```
Failures:

  test/functional\\FunctionalCodeActionsSpec.hs:110:16:
  1) FunctionalCodeActions, code actions, rename suggestions, works
       expected: "main = putStrLn \"hello\""
        but got: "main =  \"hello\""

  To rerun use: --match "/FunctionalCodeActions/code actions/rename suggestions/works/"

  test/functional\\FunctionalCodeActionsSpec.hs:127:18:
  2) FunctionalCodeActions, code actions, rename suggestions, doesn't give both documentChanges and changes
       expected: "foo = putStrLn \"world\""
        but got: "foo =  \"world\""

  To rerun use: --match "/FunctionalCodeActions/code actions/rename suggestions/doesn't give both documentChanges and changes/"

  test/functional\\FunctionalCodeActionsSpec.hs:729:11:
  3) FunctionalCodeActions, code actions, import suggestions, formats with brittany, Execute HsImport with formatter brittany, multiple import-list formats
       uncaught exception: ErrorCall
       No action found to be executed!
        Actual actions titles: ["Replace with "]
        Expected actions titles: ["Import module System.IO (hPutStrLn)","Import module System.IO (stdout)","Import module Control.Monad (when)","Import module Data.Maybe (fromMaybe)"]
       CallStack (from HasCallStack):
         error, called at test/functional\\FunctionalCodeActionsSpec.hs:729:11 in main:FunctionalCodeActionsSpec

  To rerun use: --match "/FunctionalCodeActions/code actions/import suggestions/formats with brittany/Execute HsImport with formatter brittany/multiple import-list formats/"

  test/functional\\FunctionalCodeActionsSpec.hs:729:11:
  4) FunctionalCodeActions, code actions, import suggestions, formats with brittany, Execute HsImport with formatter brittany, respects format config, multiple import-list
       uncaught exception: ErrorCall
       No action found to be executed!
        Actual actions titles: ["Replace with "]
        Expected actions titles: ["Import module System.IO (hPutStrLn)","Import module System.IO (stdout)","Import module Control.Monad (when)","Import module Data.Maybe (fromMaybe)"]
       CallStack (from HasCallStack):
         error, called at test/functional\\FunctionalCodeActionsSpec.hs:729:11 in main:FunctionalCodeActionsSpec

[...]

PR #1392, opened 2019-09-24
Fix code renaming in windows

jneira:
*  Taking in account terms between backquote and single quote
* In my `hie.log` i'be observed that the quickfix request is
```
{
  "jsonrpc": "2.0",
  "id": 38,
  "method": "textDocument/codeAction",
  "params": {
    "textDocument": {
      "uri": "file:///d%3A/ws/haskell/haskell-ide-engine/test/testdata/CodeActionRename.hs"
    },
    "range": {
      "start": {
        "line": 0,
        "character": 11
      },
      "end": {
        "line": 0,
        "character": 11
      }
    },
    "context": {
      "diagnostics": [
        {
          "range": {
            "start": {
              "line": 0,
              "character": 7
            },
            "end": {
              "line": 0,
              "character": 15
            }
          },
          "message": "* Variable not in scope: butStrLn :: [Char] -> t\n* Perhaps you meant `putStrLn' (imported from Prelude)",
          "severity": 1,
          "source": "ghcmod"
        }
      ]
    }
  }
}
```

* So the suggestion `putStrLn` is between `` ` `` and `` ' `` instead `` ‘ `` and `` ’ `` (the delimiters expected by code)
* Not sure why it is different in windows but the code now pass the tests 1 and 2 of #1391 .

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
+++ b/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
@@ -314,7 +314,7 @@ extractRenamableTerms msg
                      . T.dropWhileEnd (== e)
                      . T.dropAround (\c -> c /= b && c /= e)
     getEnclosed txt = case getEnclosed' '‘' '’' txt of
-                        ""  -> getEnclosed' '`' '\'' txt
+                        ""  -> getEnclosed' '`' '\'' txt -- Needed for windows
                         enc -> enc
 
 extractRedundantImport :: T.Text -> Maybe T.Text
--- a/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
+++ b/src/Haskell/Ide/Engine/Plugin/GhcMod.hs
@@ -310,9 +310,12 @@ extractRenamableTerms msg
                        . T.lines
     singleSuggestions = T.splitOn "), " -- Each suggestion is comma delimited
     isKnownSymbol t = " (imported from" `T.isInfixOf` t  || " (line " `T.isInfixOf` t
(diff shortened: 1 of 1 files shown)


src/Haskell/Ide/Engine/Plugin/GhcMod.hs
```diff
@@ -310,9 +310,12 @@ extractRenamableTerms msg
                        . T.lines
     singleSuggestions = T.splitOn "), " -- Each suggestion is comma delimited
     isKnownSymbol t = " (imported from" `T.isInfixOf` t  || " (line " `T.isInfixOf` t
-    getEnclosed = T.dropWhile (== '‘')
-                . T.dropWhileEnd (== '’')
-                . T.dropAround (\c -> c /= '‘' && c /= '’')
+    getEnclosed' b e = T.dropWhile (== b)
+                     . T.dropWhileEnd (== e)
+                     . T.dropAround (\c -> c /= b && c /= e)
+    getEnclosed txt = case getEnclosed' '‘' '’' txt of
```
fendor: I think a comment why we do this would be very helpful for the future!

jneira: I've just was thinking in what message would be adequate! 😄 
But i wanted to investigate a little bit why delimiters are different in windows and linux, i is soooo strange.

jneira: I've found https://gitlab.haskell.org//ghc/ghc/issues/2507#note_79667 (closed)
I deduce that it is related with utf support, anyway i'll simply comment that it is needed for windows.

PR #1399, opened 2019-09-30
Fix more code actions in windows

jneira:
* Follow up of #1392 
* It turns out that there were more places where the code assumed ``‘`` and ``’`` as term delimiters
  * And it assumed that the symbol to start a list option was `•` (in windows is `*`)
* It fixes almost all failed tests listed in #1391 (all except 11, 12 and 13) :smile: 
* The duplicated function `extractTerm` could be in common module but i am not sure where would be: maybe `Haskell.Ide.Engine.PluginUtils` in `hie-plugin-api`?
* An alternative impl for `extractTerm`would be `T.init $ T.tail $ T.dropWhileEnd (/= b) $ T.dropWhile (/= e)`, do you think that it would be better?

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/Package.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Package.hs
@@ -338,11 +338,19 @@ codeActionProvider plId docId _ context = do
 -- | Extract a module name from an error message.
 extractModuleName :: T.Text -> Maybe Package
 extractModuleName msg
-  | T.isPrefixOf "Could not find module " msg = Just $ T.tail $ T.init nameAndQuotes
-  | T.isPrefixOf "Could not load module " msg = Just $ T.tail $ T.init nameAndQuotes
+  | T.isPrefixOf "Could not find module " msg = Just $ extractTerm line
+  | T.isPrefixOf "Could not load module " msg = Just $ extractTerm line
   | otherwise = Nothing
   where line = head $ T.lines msg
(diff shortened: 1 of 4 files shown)


src/Haskell/Ide/Engine/Plugin/Package.hs
```diff
@@ -338,11 +338,19 @@ codeActionProvider plId docId _ context = do
...
 extractModuleName msg
-  | T.isPrefixOf "Could not find module " msg = Just $ T.tail $ T.init nameAndQuotes
-  | T.isPrefixOf "Could not load module " msg = Just $ T.tail $ T.init nameAndQuotes
+  | T.isPrefixOf "Could not find module " msg = Just $ extractTerm line
+  | T.isPrefixOf "Could not load module " msg = Just $ extractTerm line
   | otherwise = Nothing
   where line = head $ T.lines msg
-        nameAndQuotes = T.dropWhileEnd (/= '’') $ T.dropWhile (/= '‘') line
+
+extractTerm :: T.Text -> T.Text
```
fendor: Can we somehow reduce this code duplication?

jneira: You mean moving `extractTerm`to a common module? Yeah, that is my intention but i am not sure where to place it (`askell.Ide.Engine.PluginUtils`?)

fendor: I would rather go for HieExtras.

PR #1405, opened 2019-10-08
Add stack-install-cabal target and confirmation messages

jneira:
* For stack users that want to install auto cabal with the required minimum version to work with hie
* As bonus, i've added some confirmation messages when installing cabal and when installing hie using cabal

### DIFF
--- a/install/src/HieInstall.hs
+++ b/install/src/HieInstall.hs
@@ -91,6 +91,7 @@ defaultMain = do
       )
 
     -- stack specific targets
+    when isRunFromStack (phony "stack-install-cabal" (need ["cabal"]))
     phony "stack-build"     (need (reverse $ map ("stack-hie-" ++) hieVersions))
     phony "stack-build-all" (need ["build-data", "build"])
     phony "stack-build-data" $ do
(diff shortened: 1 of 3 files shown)


jneira:
The pr should fix the behaviour described in https://github.com/alanz/vscode-hie-server/issues/131#issuecomment-539478925

power-fungus:
Btw, is cabal still reqired for stack projects when using the hie bios?
Cc @fendor

fendor:
I dont think so. Stack should depend only on `stack`.

fendor:
However, stack cradle depends on `ghc`. @power-fungus

install/src/HieInstall.hs
```diff
@@ -91,6 +91,7 @@ defaultMain = do
       )
 
     -- stack specific targets
+    when isRunFromStack (phony "stack-install-cabal" (need ["cabal"]))
```
fendor: "stack-install-cabal" does not fit to the help-message, I think? Or am I overlooking something?

jneira: The `stack-` prefix is added by `stackTarget` (it could be added to cabal targets and it would be `cabal-install-cabal` 😆)

fendor: I see, thanks for the clarification!

PR #1406, opened 2019-10-10
Find and run cabal in user original $PATH

jneira:
* We've found that the `cabal`executable used within `stack install.hs` executions after a first `stack install.hs stack-install-cabal` is the cached in `$STACK_ROOT/snapshots/{hash}/bin`. The cause is `stack` prepends that directory to the `$PATH` environment variable. That supposes:
  * If you delete the cabal executable in `$PATH` after running `stack install.hs stack-install-cabal`, another invocation doesn't copy again the cabal executable, cause the version cached by stack is found. (thanks @jchia for detected it)
  * All `stack install.hs cabal-*` use that cached version and no the original one, so if you upgrade your cabal in `$PATH` to `3.0.0.0` stack will continue using its cached version, (currently `2.4.1.0`)
    * So in windows you could not use those goals

To avoid it i had to remove manually `$STACK_ROOT/snapshots/{hash}/bin` from `$PATH`, an ugly ugly hack, but i didint find better way to do it (via stack flags or alternative env vars)

I think this one would fix https://github.com/alanz/vscode-hie-server/issues/131

### DIFF
--- a/install/hie-install.cabal
+++ b/install/hie-install.cabal
@@ -21,6 +21,7 @@ library
   build-depends:       base >= 4.9 && < 5
                      , shake == 0.17.8
                      , directory
+                     , filepath
                      , extra
                      , text
   default-extensions: LambdaCase
--- a/install/src/Cabal.hs
+++ b/install/src/Cabal.hs
@@ -24,7 +24,7 @@ execCabal_ :: [String] -> Action ()
 execCabal_ = execCabalWithOriginalPath
 
 execCabalWithOriginalPath :: CmdResult r => [String] -> Action r
-execCabalWithOriginalPath = withOriginalPath . (command [] "cabal")
+execCabalWithOriginalPath = withoutStackCachedBinaries . (command [] "cabal")
 
 cabalBuildData :: Action ()
 cabalBuildData = do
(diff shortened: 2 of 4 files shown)


power-fungus:
I like the idea, but the approach is very hacky indeed.

Is it a possible alternative to use [`findExecutables`](https://hackage.haskell.org/package/directory-1.3.4.0/docs/System-Directory.html#v:findExecutables) and select the appropriate executable from that list?

jneira:
I thought in use that at first but unfortunately in windows `findExecutables` only returns the first one (as it states in the comments).

Imho the right one would be to have in `stack` a way (flag or whatever) to ignore its modifications over the $PATH variable... :worried:

alanz:
I am having issues on Debian with tests failing when I build locally, but passing fine on CI, e.g. https://circleci.com/workflow-run/57c08f4c-eca4-46af-bedd-97c3ae0d0835.

I wonder if this cached cabal could be the issue for me.

jneira:
i've opened an issue in `stack` to try to remove or make less britle the hack: https://github.com/commercialhaskell/stack/issues/5055

jneira:
I've updated the pr:
* change the hack function name to `withoutStackCachedBinaries`: an ugly hack deserves a precise name
* Remove just `${snapshot-install-root}/bin` and `${local-install-root}\bin`

This could be merged, as the stack issue has not feedback and there is no better alternative

power-fungus:
Im ok with this. @fendor should we merge this now?

PR #1418, opened 2019-10-20
Update haskell-lsp to 0.17

alanz:
Together with haskell-lsp-types and lsp-test

This now supports LSP spec 3.15


bubba:
Not all of 3.15, but the new progress API anyway. And it’s nice to see that by abstracting away the mechanism in haskell-lsp, we didn’t need to change any code at all!

PR #1427, opened 2019-10-27
Register rename and implementation provider

bubba:
These need to be passed to haskell-lsp in order to declare it in the server capabilities.


alanz:
This feels like a code smell in haskell-lsp.  It should pick it up from the presence of a handler, like most of the others.

bubba:
@alanz I mentioned this to @fendor on IRC earlier. It seems fairly arbitrary what capabilities are inferred and what need to be explicitly declared. I personally think haskell-lsp should infer them all, and `Core.Options` should only contain non-capability specific options, such as completion prefix character or command ids.

alanz:
I agree. Saying a thing is supported but without having a handler cannot work.

PR #1430, opened 2019-10-29
Drop stack support for GHC 8.2.2

fendor:
This is necessary, because when we upgrade cabal-helper,
and transivitely cabal-plan to  version 0.5.0.0 in #1126,
the minimum cabal-version is 2.2.
However, each stack snapshot ships with its own version
of cabal-install that can not be changed.

The latest lts for ghc-8.2.2 is 11.18, which ships with
a cabal-install version of 2.0.0.1, which is too old
for cabal-plan 0.5.0.0.

However, in general, there is nothing that forces us to
drop support for GHC 8.2.2. Overriding the compiler version
should be possible, e.g. taking the snapshort 14.8 and
overriding the GHC version, but this requires us
to do manual constraint solving. Could be done if needed.

closes #1430


jneira:
I assume that most users could use a ghc version greater than 8.2, but `cabal` is still able to build with `ghc-8.2.2`, no? ([this is](https://gist.github.com/jneira/f8c28a6f682b2b854018fe5b62912a2b) the `cabal.project.freeze` for `cabal freeze -w ghc-8.2.2`).  stack can't do it due its more reproducible but rigid snapshots.
So build with cabal could be still a workaround for users that have to use it (if we dont remove those cpp's conditions).
Until hie could be used with ghc-8.8 we will support only two main versions of ghc... :thinking:

alanz:
> Until hie could be used with ghc-8.8 we will support only two main versions of ghc... thinking

@jneira we will just have to put up with that, people using the older GHC can still use the last hie that supported it.

And I think `cabal-plan-0.5.0.0` cannot be built with the older cabal, and won't build with the newer one and GHC 8.2.  I remember trying to do that, and coming to the conclusion that we had to drop it.

It has been a second class citizen for some time, running with older versions of things like hlint.

fendor:
One more review, now deletes all legacy code that was necessary for compatibility with GHC 8.2.2. Including tests, dependencies in .cabal files and CI.

hie-plugin-api/Haskell/Ide/Engine/Compat.hs
```diff
@@ -37,79 +37,68 @@ isExtensionOf ext         = isSuffixOf ('.':ext) . takeExtensions
...
-#endif
 
 pattern HsOverLitType :: Type.Type -> GHC.HsExpr GhcTc
 pattern HsOverLitType t <-
 #if MIN_VERSION_ghc(8, 6, 0)
     GHC.HsOverLit _ (GHC.overLitType -> t)
-#elif MIN_VERSION_ghc(8, 4, 0)
-    GHC.HsOverLit (GHC.overLitType -> t)
 #else
+    -- elif MIN_VERSION_ghc(8, 4, 0)
```
alanz: This comment should probably come out

hie-plugin-api/Haskell/Ide/Engine/Compat.hs
```diff
@@ -37,79 +37,68 @@ isExtensionOf ext         = isSuffixOf ('.':ext) . takeExtensions
...
 #endif
 
 pattern HsLamCaseType :: Type.Type -> GHC.HsExpr GhcTc
 pattern HsLamCaseType t <-
 #if MIN_VERSION_ghc(8, 6, 0)
     GHC.HsLamCase _ ((\(GHC.MG { GHC.mg_ext = groupTy }) -> matchGroupType groupTy) -> t)
-#elif MIN_VERSION_ghc(8, 4, 0)
-    GHC.HsLamCase (\GHC.MG { GHC.mg_res_ty = res, GHC.mg_arg_tys = args } -> Type.mkFunTys args res -> t)
 #else
+    -- elif MIN_VERSION_ghc(8, 4, 0)
```
alanz: And ditto for all the others.

hie-plugin-api/Haskell/Ide/Engine/Compat.hs
fendor: I thought, it may be helpful to have them, but you are right, it is unnecessary.

fendor:
HieWrapper tests are simple tests that check that the correct ghc version is discovered, no change necessary.

alanz:
I am happy for this to land when CI passes

PR #1435, opened 2019-11-05
Try out GitHub actions

bubba:


jneira:
hi @bubba ! github actions looks nice and to have a specifical config for haskell project is really great. Do you think it could replace azure as default future ci env for the project?
In that case maybe we should not invest much more time in azure config...

bubba:
@jneira this is mostly experimental to see if it's viable or not. So far there's definitely a lot of kinks with the setup-haskell action, mainly that it doesn't support macOS or Windows so I would keep the azure config around!

bubba:
I am closing this for now as the [cache limit is too small](https://github.com/actions/cache/issues/6) to be useful

PR #1437, opened 2019-11-06
Ignore the PATH fix if it is not set

### LINKED ISSUE #1436: stack ./install.hs stack-install-cabal failed
```
$ stack ./install.hs stack-install-cabal
run from: stack
# stack (for cabal)
# stack (for cabal)
  at want, called at src/Development/Shake/Internal/Args.hs:81:69 in shake-0.17.8-42j6WLpuW2RLDjaIMjbO7Y:Development.Shake.Internal.Args
* Depends on: stack-install-cabal
  at need, called at src/HieInstall.hs:94:55 in hie-install-0.8.0.0-217T3gMG2hYDJU1sGHGe5h:HieInstall
* Depends on: cabal
* Raised the exception:
PATH: getEnv: does not exist (no environment variable)
```

jneira:
* If it is not set the fix makes no sense
* Fixes #1436

### DIFF
--- a/install/src/Stack.hs
+++ b/install/src/Stack.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE ScopedTypeVariables #-} 
 module Stack where
 
 import           Development.Shake
--- a/install/src/Stack.hs
+++ b/install/src/Stack.hs
@@ -106,7 +106,7 @@ withoutStackCachedBinaries :: Action a -> Action a
 withoutStackCachedBinaries action = do
   mbPath <- liftIO (catch
                     (lookupEnv "PATH")
-                    (\(_ :: SomeException) -> return Nothing))
+                    (\(_ :: IOException) -> return Nothing))
 
   case (mbPath, isRunFromStack) of
 
--- a/install/src/Stack.hs
+++ b/install/src/Stack.hs
@@ -10,6 +10,7 @@ import           Data.List
 import           System.Directory                         ( copyFile )
 import           System.FilePath                          ( searchPathSeparator, (</>) )
 import           System.Environment                       ( lookupEnv, setEnv, getEnvironment )
+import           System.IO.Error                          ( isDoesNotExistError )
 import           BuildSystem
 import           Version
 import           Print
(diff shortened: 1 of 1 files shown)


install/src/Stack.hs
```diff
@@ -102,7 +104,9 @@ stackBuildFailMsg =
 -- |Run actions without the stack cached binaries 
 withoutStackCachedBinaries :: Action a -> Action a
 withoutStackCachedBinaries action = do
-  mbPath <- liftIO (lookupEnv "PATH")
+  mbPath <- liftIO (catch
+                    (lookupEnv "PATH")
+                    (\(_ :: SomeException) -> return Nothing))
```
fendor: `SomeException` also catches Async Exceptions. In this case, this may be safe, but it is a bad idea in general.
Please just catch `IOException` (assuming that is the actual exception that is thrown.)

jneira: Oh, right, thanks for pointed it out

fendor:
In https://github.com/haskell/haskell-ide-engine/issues/1436#issuecomment-550301084 it is said that the path is set. Maybe dont merge this until it is clear why this issue arises?

jneira:
> In #1436 (comment) it is said that the path is set. Maybe dont merge this until it is clear why this issue arises?

Sure, but i think the actual code does what i expected in the fisrt place (return `Nothing` if the var is not informed)

> Is catching an exception the best way to detect if the environment variable is set? This seems a bit hacky to me, but I have no experience with the System.Environment module

It was a surprise to me that `lookupEnv` throws an Exception instead `Nothing` if the variable is not set (why return `Maybe String` then?)

The [docs ](http://hackage.haskell.org/package/base-4.12.0.0/docs/System-Environment.html#v:lookupEnv) for `lookupEnv` states:
> For POSIX users, this is equivalent to getEnv.

...and `getenv` trhows an IOException if the variable is not set  🤦‍♂

fendor:
>
>
> ...and getenv trhows an IOException if the variable is not set 🤦‍♂

then the patch makes sense. Maybe we should only catch exceptions that satisfy the predicate `isDoesNotExistError`? Otherwise, this PR is good to go.

fendor:
@jneira Merge when you are ready!

PR #1440, opened 2019-11-07
Adapt liquid haskell tests for version 0.8.6.2

alanz:
This needs https://github.com/alanz/haskell-dockerfiles/pull/2 for CI,
to update the base image GHC and liquid haskell version.


jneira:
Great! i think we will able to fix liquid tests easily in other ci's with that

alanz:
It seems to have stabilised now on just having the one failure

```
  To rerun use: --match "/FunctionalCodeActions/code actions/add package suggestions/adds to .cabal files/"
```

Which I also see locally.

alanz:
I think this is ready for review, and land

PR #1450, opened 2019-11-11
[WIP] Generate an "Apply all hints" hlint code action

alanz:
This commit generates it all the time, without checking if we actually have any hints.

This is a possible alternative to https://github.com/alanz/vscode-hie-server/pull/168

But it may be better to update the `CodeActionProvider` interface to provide all the diagnostics for the given Uri/plugin combo, which can then be checked.  Except that it may be too much processing, given the frequency of request.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
+++ b/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
@@ -307,7 +307,10 @@ codeActionProvider plId docId _ context = IdeResultOk <$> hlintActions
   where
 
     hlintActions :: IdeM [LSP.CodeAction]
-    hlintActions = catMaybes <$> mapM mkHlintAction (filter validCommand diags)
+    hlintActions = do
+      actions <- mapM mkHlintAction (filter validCommand diags)
+      applyAll <- mkApplyAllAction
+      return (catMaybes $ applyAll:actions)
 
     -- |Some hints do not have an associated refactoring
     validCommand (LSP.Diagnostic _ _ (Just (LSP.StringValue code)) (Just "hlint") _ _) =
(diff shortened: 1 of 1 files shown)


bubba:
This seems to be affecting stand alone hints? Just from looking at the diff it’s not obvious how exactly though

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -307,7 +307,10 @@ codeActionProvider plId docId _ context = IdeResultOk <$> hlintActions
   where
 
     hlintActions :: IdeM [LSP.CodeAction]
-    hlintActions = catMaybes <$> mapM mkHlintAction (filter validCommand diags)
+    hlintActions = do
+      actions <- mapM mkHlintAction (filter validCommand diags)
+      applyAll <- mkApplyAllAction
```
bubba: We should only return this hint if there is a hint available to apply
```suggestion
      applyAll <- if null actions then return Nothing else mkApplyAllAction
```

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -307,7 +307,10 @@ codeActionProvider plId docId _ context = IdeResultOk <$> hlintActions
...
+      return (catMaybes $ applyAll:actions)
```
bubba: Some clients like coc.vim have keybindings to apply the first code action, so would it make sense to return this last?
```suggestion
      return (catMaybes $ actions ++ [applyAll])
```

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
alanz: The way code actions works is that the server sends diagnostics for the file to the client.  The client then sends back any diagnostics that apply to the source where the cursor is currently located in the file.

So if your cursor is near the end of the file, and there is a hint near the beginning, then the actions will be empty.

Hence my remark in the original description that we should be able to quickly check if there are in fact diagnostics for a given file/diagnostics source combination, anywhere in the file.

PR #1452, opened 2019-11-14
Various improvements and fixes over shake install

### LINKED ISSUE #1436: stack ./install.hs stack-install-cabal failed

### LINKED ISSUE #1439: liquidhaskell 0.8.6.2 output is noisy
See https://github.com/ucsd-progsys/liquidhaskell/issues/1562

We must now ignore all lines until we see `^RESULT$`.

### LINKED ISSUE #1445: Install.hs: cabal-ghcs shows unsupported GHC versions
Based on https://github.com/haskell/haskell-ide-engine/issues/1436#issuecomment-552214342.

Example:
```bash
$ stack install.hs cabal-ghcs
run from: stack

******************************
Found the following GHC paths:
ghc-8.8.1: /opt/ghc/bin/ghc
ghc-8.0.1: /usr/bin/ghc

******************************

Build completed in 0.00s
```

However, since we do not support these versions, they should not be detected, since these are transitively used to show the help-message. 
Bug-fix could be to intersect the found ghc versions with the supported ghc versions, which are obtained by parsing all `stack-*.yaml` files in the directory.

jneira:
* Improvements 
  * ~~Use explicit list of ghc supported versions to filter out targets (stack and cabal).~~
  * Use the list of stack-*.yaml files to filter out unsupported ghc versions of available ones.  Fixes #1445
  * Add new target `build-lastest`to build hie with the lastest supported and available ghc
  * *Change the default `build` target to build the lastest version and data files instead `build-all`*
  * ~~Prints a warning about consumed resources and time when executing `build-all` and when printing the help~~
  * Remove the `build-all` target (see https://github.com/haskell/haskell-ide-engine/pull/1452#discussion_r346941695)
* Bug fixes
  * make `cabal-ghcs` work in windows
  * fix the compiler used in the `cabal v2-install` phase of `cabal-hie-<version>`target
  * fix the `stack-install-cabal` target for posix envs. Hopefully it fixes #1436
    * Thanks @fendor for detecting the offending code (introduced by myself with https://github.com/haskell/haskell-ide-engine/pull/1406/files#diff-707cf6d4db4ff1099736896c9d96a6f8R132)
  * sync docs with recent changes (included this one)

* Unrelated changes
  * Fix the builds of macos in azure

### DIFF
--- a/docs/Build.md
+++ b/docs/Build.md
@@ -28,10 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
 The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
 
 * `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
-* `build-lastest`: builds ad installs `hie` for the lastest available and supported `ghc` version.
+* `build-latest`: builds ad installs `hie` for the latest available and supported `ghc` version.
 * `build-all`: builds and installs `hie` binaries for all supported `ghc` versions. This option may take a long time and computer resources so use it with caution.
 * `build-data`: builds the hoogle-db required by `hie`
-* `build`:  builds ad installs `hie` for the lastest supported `ghc` version (like `build-lastest`) and the hoogle-db (like `build-data`)
+* `build`:  builds ad installs `hie` for the latest supported `ghc` version (like `build-latest`) and the hoogle-db (like `build-data`)
 * `cabal-*`: execute the same task as the original target, but with `cabal` instead of `stack`
 
 Each `stack-*.yaml` contains references to packages in the submodules. Calling `stack` with one of those causes the build to fail if the submodules have not been initialized already. The file `shake.yaml` solves this issue invoking the `git` binary itself to update the submodules. Moreover, it specifies the correct version of `shake` and is used for installing all run-time dependencies such as `cabal` and `hoogle` if necessary.
--- a/docs/Build.md
+++ b/docs/Build.md
@@ -13,7 +13,6 @@ The design of the build system has the following main goals:
     - `stack`
     - `git`
 * is completely functional right after a simple `git clone` and after every `git pull`
-* one-stop-shop for building and naming all executables required for using `hie` in IDEs.
 * prevents certain build failures by either identifying a failed precondition (such as wrong `stack` version) or by performing the necessary steps so users can't forget them (such as invoking `git` to update submodules)
 
 
@@ -29,7 +28,6 @@ The build script `install.hs` defines several targets using the `shake` build sy
 
 * `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
 * `build-latest`: builds and installs `hie` for the latest available and supported `ghc` version.
-* `build-all`: builds and installs `hie` binaries for all supported `ghc` versions. This option may take a long time and computer resources so use it with caution.
 * `build-data`: builds the hoogle-db required by `hie`
 * `build`:  builds and installs `hie` for the latest supported `ghc` version (like `build-latest`) and the hoogle-db (like `build-data`)
 * `cabal-*`: execute the same task as the original target, but with `cabal` instead of `stack`
--- a/docs/Build.md
+++ b/docs/Build.md
@@ -28,8 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
 The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
 
 * `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
-* `build`: builds and installs `hie` binaries for all supported `ghc` versions.
+* `build-lastest`: builds ad installs `hie` for the last available and supported `ghc` version.
+* `build-all`: builds and installs `hie` binaries for all supported `ghc` versions. This option may take a long time and computer resources so use it with caution.
 * `build-data`: builds the hoogle-db required by `hie`
+* `build`:  builds ad installs `hie` for the last supported `ghc` version (like `build-lastest`) and the hoogle-db (like `build-data`)
 * `cabal-*`: execute the same task as the original target, but with `cabal` instead of `stack`
 
 Each `stack-*.yaml` contains references to packages in the submodules. Calling `stack` with one of those causes the build to fail if the submodules have not been initialized already. The file `shake.yaml` solves this issue invoking the `git` binary itself to update the submodules. Moreover, it specifies the correct version of `shake` and is used for installing all run-time dependencies such as `cabal` and `hoogle` if necessary.
@@ -38,7 +40,7 @@ Each `stack-*.yaml` contains references to packages in the submodules. Calling `
 
 `hie` depends on a correct environment in order to function properly:
 
-* `cabal-install`: If no `cabal` executable can be found or has an outdated version, `cabal-install` is installed via `stack`.
+* `cabal-install`: This dependency is required by `hie` to handle correctly projects that are not `stack` based (without `stack.yaml`). You can install an appropiate version using `stack` with the `stack-install-cabal` target.
 * The `hoogle` database: `hoogle generate` needs to be called with the most-recent `hoogle` version.
 
 ### Steps to build `hie`
@@ -47,10 +49,9 @@ Installing `hie` is a multi-step process:
 
 1. `git submodule sync && git submodule update --init`
 2. `hoogle generate` (`hoogle>=5.0.17` to be safe)
-3. ensure that `cabal-install` is installed in the correct version
-4. `stack --stack-yaml=stack-<X>.yaml install` or `cabal new-install -w ghc-<X>`
-5. rename `hie` binary to `hie-<X>` in `$HOME/.local/bin`, where `<X>` is the GHC version used
-6. repeat step 4 and 5 for all desired GHC versions
+3. `stack --stack-yaml=stack-<X>.yaml install` or `cabal new-install -w ghc-<X>`
+4. rename `hie` binary to `hie-<X>` in `$HOME/.local/bin`, where `<X>` is the GHC version used
+5. repeat step 4 and 5 for all desired GHC versions
 
 This ensures that a complete install is always possible after each `git pull` or a `git clone`.
 
@@ -90,19 +91,17 @@ The final step is to configure the `hie` client to use a custom `hie-wrapper` sc
 The `install.hs` script performs some checks to ensure that a correct installation is possible and provide meaningful error messages for known issues.
 
 * `stack` needs to be up-to-date. Version `1.9.3` is required
+* `cabal` needs to be up-to-date. Version `3.0.0.0` is required for windows systems and `2.4.1.0` for other ones.
 * `ghc-8.6.3` is broken on windows. Trying to install `hie-8.6.3` on windows is not possible.
-* `cabal new-build` does not work on windows at the moment. All `cabal-*` targets exit with an error message about that.
 * When the build fails, an error message, that suggests to remove `.stack-work` directory, is displayed.
 
 ### Tradeoffs
 
 #### `stack` is a build dependency
 
-Currently, it is not possible to build all `hie-*` executables automatically without `stack`, since the `install.hs` script is executed by `stack`.
+Currently, `stack` is needed even if you run the script with `cabal` to get the path where install the binaries but there are plans to remove that dependency (see #1380).
 
-We are open to suggestions of other build systems that honor the requirements above, but are executable without `stack`.
-
-#### `install.hs` installs a GHC before running
+#### run `install.hs` with `stack` installs a GHC before running
 
 Before the code in `install.hs` can be executed, `stack` installs a `GHC`, depending on the `resolver` field in `shake.yaml`. This is necessary if `install.hs` should be completely functional right after a fresh `git clone` without further configuration.
 
--- a/docs/Build.md
+++ b/docs/Build.md
@@ -28,10 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
 The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
 
 * `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
-* `build-latest`: builds ad installs `hie` for the latest available and supported `ghc` version.
+* `build-latest`: builds and installs `hie` for the latest available and supported `ghc` version.
 * `build-all`: builds and installs `hie` binaries for all supported `ghc` versions. This option may take a long time and computer resources so use it with caution.
 * `build-data`: builds the hoogle-db required by `hie`
-* `build`:  builds ad installs `hie` for the latest supported `ghc` version (like `build-latest`) and the hoogle-db (like `build-data`)
+* `build`:  builds and installs `hie` for the latest supported `ghc` version (like `build-latest`) and the hoogle-db (like `build-data`)
 * `cabal-*`: execute the same task as the original target, but with `cabal` instead of `stack`
 
 Each `stack-*.yaml` contains references to packages in the submodules. Calling `stack` with one of those causes the build to fail if the submodules have not been initialized already. The file `shake.yaml` solves this issue invoking the `git` binary itself to update the submodules. Moreover, it specifies the correct version of `shake` and is used for installing all run-time dependencies such as `cabal` and `hoogle` if necessary.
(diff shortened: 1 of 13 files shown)


docs/Build.md
```diff
@@ -28,8 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
 The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
 
 * `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
-* `build`: builds and installs `hie` binaries for all supported `ghc` versions.
+* `build-lastest`: builds ad installs `hie` for the last available and supported `ghc` version.
```
fendor: Typo `latest`

fendor: typo `ad`

fendor: s/last/latest/ maybe?

docs/Build.md
```diff
@@ -28,8 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
...
+* `build-all`: builds and installs `hie` binaries for all supported `ghc` versions. This option may take a long time and computer resources so use it with caution.
```
fendor: s/computer resources so use it with caution/a lot of cpu resources, so use with caution/ maybe?

docs/Build.md
```diff
@@ -28,8 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
...
 * `build-data`: builds the hoogle-db required by `hie`
+* `build`:  builds ad installs `hie` for the last supported `ghc` version (like `build-lastest`) and the hoogle-db (like `build-data`)
```
fendor: s/last/latest/ again maybe?

docs/Build.md
```diff
@@ -38,7 +40,7 @@ Each `stack-*.yaml` contains references to packages in the submodules. Calling `
 
 `hie` depends on a correct environment in order to function properly:
 
-* `cabal-install`: If no `cabal` executable can be found or has an outdated version, `cabal-install` is installed via `stack`.
+* `cabal-install`: This dependency is required by `hie` to handle correctly projects that are not `stack` based (without `stack.yaml`). You can install an appropiate version using `stack` with the `stack-install-cabal` target.
```
fendor: strictly speaking, hie may still work if you exclusively work on stack projects. Dont know if we should say that, though.

docs/Build.md
```diff
@@ -47,10 +49,9 @@ Installing `hie` is a multi-step process:
 
 1. `git submodule sync && git submodule update --init`
 2. `hoogle generate` (`hoogle>=5.0.17` to be safe)
-3. ensure that `cabal-install` is installed in the correct version
-4. `stack --stack-yaml=stack-<X>.yaml install` or `cabal new-install -w ghc-<X>`
-5. rename `hie` binary to `hie-<X>` in `$HOME/.local/bin`, where `<X>` is the GHC version used
-6. repeat step 4 and 5 for all desired GHC versions
+3. `stack --stack-yaml=stack-<X>.yaml install` or `cabal new-install -w ghc-<X>`
```
fendor: why no `cabal-install` step anymore?

docs/Build.md
```diff
@@ -47,10 +49,9 @@ Installing `hie` is a multi-step process:
...
+4. rename `hie` binary to `hie-<X>` in `$HOME/.local/bin`, where `<X>` is the GHC version used
+5. repeat step 4 and 5 for all desired GHC versions
```
fendor: should be 3 and 4 if step three has changed

docs/Build.md
```diff
@@ -90,19 +91,17 @@ The final step is to configure the `hie` client to use a custom `hie-wrapper` sc
 The `install.hs` script performs some checks to ensure that a correct installation is possible and provide meaningful error messages for known issues.
 
 * `stack` needs to be up-to-date. Version `1.9.3` is required
+* `cabal` needs to be up-to-date. Version `3.0.0.0` is required for windows systems and `2.4.1.0` for other ones.
```
fendor: We are not supporting cabal 3.0.0.0 before #1126 has landed, right?

install/src/Env.hs
```diff
@@ -40,20 +42,32 @@ existsExecutable executable = liftIO $ isJust <$> findExecutable executable
 isWindowsSystem :: Bool
 isWindowsSystem = os `elem` ["mingw32", "win32"]
 
+-- | Defines all different hie versions that are supported.
+-- On windows, `8.6.3` is excluded as this version of ghc does not work there
+supportedGhcVersions :: [VersionNumber]
+supportedGhcVersions = sort (commonVersions ++ osVersions)
+  where commonVersions = ["8.4.2", "8.4.3", "8.4.4", "8.6.1", "8.6.2", "8.6.4", "8.6.5"]
```
fendor: hard coding makes it more complex to update the range of supported ghc versions. Currently, supported ghc versions are determined by `stack-*.yaml` files. This makes it easy to add or remove supported versions, no change of install.hs necessary. Only OS dependant exceptions have to be specified.

install/src/Env.hs
```diff
@@ -63,9 +77,11 @@ findInstalledGhcs = do
 -- command fits to the desired version.
 getGhcPathOf :: MonadIO m => VersionNumber -> m (Maybe GhcPath)
 getGhcPathOf ghcVersion =
-  liftIO $ findExecutable ("ghc-" ++ ghcVersion) >>= \case
+  liftIO $ findExecutable ("ghc-" ++ ghcVersion <.> exe) >>= \case
     Nothing -> lookup ghcVersion <$> getGhcPaths
     path -> return path
+  where exe | isWindowsSystem = "exe"
```
fendor: Shake already provides this function, iirc.

install/src/Help.hs
```diff
@@ -12,13 +12,23 @@ import           Version
 import           BuildSystem
 import           Cabal
 
+stackCommand :: TargetDescription -> String
+stackCommand target = "stack install.hs " ++ fst target
+
+cabalCommand :: TargetDescription -> String
+cabalCommand target = "cabal new-run install.hs --project-file install/shake.project " ++ fst target
```
fendor: should we promote `v2-` since this is forward and backwards compatible?

install/src/Help.hs
```diff
@@ -114,17 +127,31 @@ hieTarget version =
...
-  ("build-all", "Builds hie for all installed GHC versions and the data files")
+  ( "build-all"
+  , "Builds hie for all installed GHC versions and the data files. "
+  ++ buildAllWarning)
+
+buildAllWarning :: String
+buildAllWarning = "WARNING: This command may take a long time and computer resources"
+
+buildAllWarningAlt :: String
+buildAllWarningAlt = "Consider build only the needed ghc versions using:\n"
```
fendor: Consider building only the ghc versions  you need using:

install/src/Help.hs
```diff
@@ -114,17 +127,31 @@ hieTarget version =
...
 
 buildAllTarget :: TargetDescription
 buildAllTarget =
-  ("build-all", "Builds hie for all installed GHC versions and the data files")
+  ( "build-all"
+  , "Builds hie for all installed GHC versions and the data files. "
+  ++ buildAllWarning)
+
+buildAllWarning :: String
+buildAllWarning = "WARNING: This command may take a long time and computer resources"
```
fendor: s/computer resources/a lot of cpu resources/ maybe?

docs/Build.md
jneira: it consumes a lot of memory too :thinking:

docs/Build.md
jneira: lastest could be correct too, no? (to use the same name that the target)

fendor: lastest is wrong, imo. Either `last` or `latest`.

jneira: Ok i will change all references to lastest to latest (including the target name)

jneira: done!

install/src/Help.hs
jneira: Yeah, cause we are supporting a cabal version that already has the `v2` prefixes

jneira: done

docs/Build.md
jneira: We support it for *build* hie but no for *use* hie. It is not consistent and i would stick to `2.4.1.0` but it can be used to build hie in windows

install/src/Env.hs
jneira: Well, i like in general to have the constraints explicitly in the code to being able to generate/validate/check conventions (like stack yaml file names).
This way we could validate that the stack yaml files names are correct (or even if the compiler version used by the resolver is).
The pr doesnt validate the stack yaml files but it filters out the stack files which versions are not in the explicit list (so if we dont remember to delete it, it would not be used in the script)

That said, if we found that, in this case, the possible benefits of a explicit list doesnt worth the duplication with the stack yaml files i would remove it

fendor: Good point. Let's wait for other opinions, e.g. @alanz, @power-fungus and @bubba.

install/src/Help.hs
jneira: Done

docs/Build.md
```diff
@@ -28,10 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
 The build script `install.hs` defines several targets using the `shake` build system. The targets are roughly:
 
 * `hie-*`: builds and installs the `hie` binaries. Also renames the binaries to contain the correct version-number.
-* `build-lastest`: builds ad installs `hie` for the lastest available and supported `ghc` version.
+* `build-latest`: builds ad installs `hie` for the latest available and supported `ghc` version.
```
Anrock: Typo: `ad` -> `and`

docs/Build.md
Anrock: Do we really need this target? It's confusing for newcomers and they're complaining about "hie takes ages to build and uses dozens of gigabytes of disk space". It's still a thing I'm seeing in reddit discussions.

docs/Build.md
```diff
@@ -28,10 +28,10 @@ See the project's `README` for detailed information about installing `hie`.
...
 * `build-all`: builds and installs `hie` binaries for all supported `ghc` versions. This option may take a long time and computer resources so use it with caution.
 * `build-data`: builds the hoogle-db required by `hie`
-* `build`:  builds ad installs `hie` for the lastest supported `ghc` version (like `build-lastest`) and the hoogle-db (like `build-data`)
+* `build`:  builds ad installs `hie` for the latest supported `ghc` version (like `build-latest`) and the hoogle-db (like `build-data`)
```
Anrock: Typo: `ad` -> `and`

docs/Build.md
jneira: Before #1368 cabal was always installed independently of the target. Now it is only installed by stack if you executes explicitly the target `stack-install-cabal` or executes a `cabal-` target with stack.

docs/Build.md
jneira: done

docs/Build.md
fendor: Sometimes you want it. But it is discouraged by the script from now on, meaning it is no longer the default "build" and displays a warning, iirc. Is it still confusing with these changes?

docs/Build.md
jneira: Thanks for catching them,.. done!

docs/Build.md
Anrock: Not sure. What are the use-cases? I imagine it's mostly for CI and binary packaging, right? If so, maybe add additional note like "Used for CI / development / packaging. You probably want a hie-latest or hie-X.Y.Z for your specific GHC"?

fendor: Fair. I would actually assume that there currently is no real use-case. Except when someone *really* wants to install all supported ghc versions. We can probably remove it?

install/src/Env.hs
jneira: Yeah, good catch, done

docs/Build.md
Anrock: > Except when someone really wants to install all supported ghc versions

Welp, that someone can make a trivial script like `./install.hs hie-8.4.4 hie-8.6.5 ...` 🤔 , so not much of a loss in case of removing this target. So looks like there is no concrete pros for this target, but we know at least one con.

jneira: Maybe the main case would be ours: to test dev changes compiles and are functionally correct for all supported ghcs.
I dont know if someone is already using it (in ci or whatever) and their motivation, maybe we could add a deprecation note and remove it in the next version if nobody complains meanwhile?.

jneira: Or change the target name to `build-all-available-ghcs` (as i did originally) and remove it in the next release

fendor: We do not use it in CI, the arch AUR uses a custom build script, I really think nobody is using it.
Locally installing all versions also seems infeasible. What I normally do is let the CI verify that everything works for every version and if there is a bug for a specific GHC version, I install only that.

jneira: ok, we can take the other way: remove it now to avoid unwanted full builds and add later if someone bring us a legitimate case use

install/src/Env.hs
alanz: FWIW I don't have a strong opinion on this. I do think capturing this info in one place is worthwhile, but if you are doing a stack install you need the appropriate stack file, so that does seem to work.

The only problem is if it leaves cabal out i the cold.

Another option might be to do it something like what [haskell-ci](https://github.com/haskell-CI/haskell-ci) does, taking the versions from the cabal file.

jneira: Mmm, use an existing field (`tested-with`) of a standard (in the haskell world) config file sounds good. 
However `haskell-ci` depends on `Cabal` to parse the `.cabal` file. We will need to add it or use an ad-hoc parser to extract the info of `tested-with` and i dont fully like neither of both

docs/Build.md
fendor: Yeah, I agree

install/src/Env.hs
Anrock: I don't have a strong opinion on this too, but explicit list sounds more preferable.

docs/Build.md
jneira: removed

docs/Build.md
fendor: That is troublesome.

install/src/Stack.hs
```diff
@@ -135,5 +132,5 @@ withoutStackCachedBinaries action = do
         splitPaths s =
           case dropWhile (== searchPathSeparator) s of
                       "" -> []
-                      s' -> w : words s''
+                      s' -> w : splitPaths s''
                             where (w, s'') = break (== searchPathSeparator) s'
```
fendor: The intention of this is to split the PATH into multiple paths separated by either `:`, or `;` right? If yes, then the function: https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Windows.html#v:splitSearchPath looks helpful

jneira: wow, just what i needed, many thanks!

docs/Build.md
jneira: Well, hie-bios are going to land really soon 😉
 i'll update the phrase to make ir clear.

install/src/Help.hs
jneira: No more build-all!

docs/Build.md
jneira: @fendor i've rephrased the requirements for cabal in the docs

install/src/Env.hs
nponeccop: In case my opinion counts, I'm against duplication so enumerating *.yml files sounds more preferable. We already have too many build-related reports.

And if we can forget something - then it's a good idea to check that with CI. So if you decide to leave the explicit list you should check during CI that the list is consistent with the `stack-*.yml` files.

jneira: Ok, as i dont have neither an strong opnion on this i will use the list of `stack-<ghc-version>.yaml`

fendor: @nponeccop Of course, your opinion counts! thank you for your input!

jneira: done!

jneira:
thanks all for the feedback! [azure ci](https://dev.azure.com/jneira/haskell-ide-engine/_build/results?buildId=297) seems to be happy too so merging

Issue #1229, opened 2019-05-01
Travis fails with "Could not find module: TestUtils" on MacOS


nponeccop:

fendor:
For which test-suite? Or is that the whole error message?

nponeccop:
See Travis logs: https://travis-ci.org/haskell/haskell-ide-engine/jobs/526823736

To find this log by hand do the following:

- go to https://github.com/haskell/haskell-ide-engine/commits/master
- see all those X which mean failed CI
- click on any X to see a popup
- click `details` link in a failed Travis build (the first line)

bubba:
Stack 1.7.1 is outdated: there's a [bug with internal libraries](https://github.com/commercialhaskell/stack/issues/4265) that was fixed in 1.9.1.
I think there is a flag you can use to tell travis to run `brew update`

bubba:
We can close this now since there's been enough time for Stack 1.9.1

PR #1456, opened 2019-11-16
[WIP] fix Travis CI

peterbecich:
https://github.com/haskell/haskell-ide-engine/issues/1229

This is close to having Travis CI working for both Ubuntu and OS X.
I've removed the `deploy` stage; figured it had been replaced by now.  Should it remain?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,7 @@ jobs:
        - export PATH=$HOME/.local/bin:$PATH
        - ls .stack-work/install/*/*/*/bin/hie || true
        # Build some dependencies to offload the next stage from doing too much work
-       - stack --stack-yaml=stack-$GHC_VER.yaml build lens aeson alex conduit x509 tls haskell-lexer Strafunski-StrategyLib HaRe streaming-commons
+       - travis_retry stack --stack-yaml=stack-$GHC_VER.yaml build lens aeson alex conduit x509 tls haskell-lexer Strafunski-StrategyLib HaRe streaming-commons
 
    - stage: some_dependencies
      os: linux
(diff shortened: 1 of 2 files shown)


bubba:
Thanks! Looks like travis CI isn't running for PRs. I think we need a Haskell org admin to turn it on?

peterbecich:
I'm expecting this to pass: https://travis-ci.org/peterbecich/haskell-ide-engine/builds/613034877

mpickering:
Why do we need travis, circle, azure and github actions CI?

bubba:
GitHub actions is just experimental, don’t consider it. CircleCI was needed for Linux, Azure was needed for windows, Travis was needed for Mac. Now they all support all operating systems though.

lorenzo:
The idea of having Travis was to generate binaries for all platforms after each tag. That was before discovering that the binaries we produce can only work in the machine they were compiled at.

.travis.yml
```diff
@@ -22,104 +27,521 @@ cache:
...
+     os: osx
+     osx_image: xcode11.2
+     env: GHC_VER="8.6.4"
+     script: *test
+
+   - stage: test
+     os: osx
+     osx_image: xcode11.2
+     env: GHC_VER="8.6.5"
+     script: *test
```
bubba: Is there a way to specify a matrix for the GHC_VER env?

peterbecich: I don't think so.  Would have been nice.  Seems that stages and a matrix of environment variables cannot be used together:
https://github.com/peterbecich/haskell-ide-engine/blob/aadbea8eb9d1212dc3752ac5a843469d5f42cd94/.travis.yml
https://travis-ci.org/peterbecich/haskell-ide-engine/builds/618340235
I don't know why the Test stage expands differently than the other stages in this build.

To me this documentation implies it would work for something other than an environment variable: https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion

peterbecich:
I'm fixing the broken Test stage now

.travis.yml
```diff
@@ -22,104 +27,521 @@ cache:
   - $TRAVIS_BUILD_DIR/submodules/haskell-lsp/haskell-lsp-types/.stack-work
   - $TRAVIS_BUILD_DIR/submodules/cabal-helper/.stack-work
   - $TRAVIS_BUILD_DIR/hie-plugin-api/.stack-work
-  timeout: 800
+  timeout: 2000
```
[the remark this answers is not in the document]
peterbecich: seconds https://docs.travis-ci.com/user/caching/#setting-the-timeout

PR #1459, opened 2019-11-18
Update README.md

flip111:
The windows pre-requirements are not optional but an absolute must when cloning. Well if you already have the right settings it will be a no-op just like installing `sudo apt install libicu-dev libtinfo-dev libgmp-dev` on linux will be a no-op when you already have them.

Setting the windows path limitation was not enough. Git also needs to be configured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ sudo dnf install libicu-devel ncurses-devel
 In order to avoid problems with long paths on Windows you can do either one of the following:
 
 1. Clone the `haskell-ide-engine` to a short path, for example the root of your logical drive (e.g. to
-   `C:\hie`). If this doesn't work try the second option.
+   `C:\hie`). If this doesn't work or you want to use a longer path, try the second option.
 
 2. If the `Local Group Policy Editor` is available on your system, go to: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled`. If you don't have the policy editor you can use regedit by using the following instructions [here](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#enable-long-paths-in-windows-10-version-1607-and-later). You also need to configure git to allow longer paths by using unicode paths. To set this for all your git repositories use `git config --system core.longpaths true` (you probably need an administrative shell for this) or for just this one repository use `git config core.longpaths true`.
 
--- a/README.md
+++ b/README.md
@@ -176,11 +176,10 @@ sudo dnf install libicu-devel ncurses-devel
 
 In order to avoid problems with long paths on Windows you can do either one of the following:
 
-1. Clone the `haskell-ide-engine` to the root of your logical drive (e.g. to
+1. Clone the `haskell-ide-engine` to a short path, for example the root of your logical drive (e.g. to
    `C:\hie`). If this doesn't work try the second option.
 
-2. In the `Local Group Policy Editor`: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled` (Works
-   only for Windows 10). Also you need to configure git to allow longer paths by using unicode paths. To set this for all your git repositories use `git config --system core.longpaths true` (you probably need an administrative shell for this) or for just this one repository use `git config core.longpaths true`.
+2. If the `Local Group Policy Editor` is available on your system, go to: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled`. If you don't have the policy editor you can use regedit by using the following instructions [here](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#enable-long-paths-in-windows-10-version-1607-and-later). You also need to configure git to allow longer paths by using unicode paths. To set this for all your git repositories use `git config --system core.longpaths true` (you probably need an administrative shell for this) or for just this one repository use `git config core.longpaths true`.
 
 #### Download the source code
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ we talk to clients.__
     - [Installation from source](#installation-from-source)
       - [Common pre-requirements](#common-pre-requirements)
       - [Linux-specific pre-requirements](#linux-specific-pre-requirements)
-      - [Windows-specific pre-requirements (optional)](#windows-specific-pre-requirements-optional)
+      - [Windows-specific pre-requirements](#windows-specific-pre-requirements)
       - [Download the source code](#download-the-source-code)
       - [Building](#building)
         - [Install specific GHC Version](#install-specific-ghc-version)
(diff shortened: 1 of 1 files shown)


Anrock:
> The windows pre-requirements are not optional but an absolute must when cloning.

Objection. I never did any of described and never had any problems with paths.

flip111:
@Anrock like i said. I could say "objection i never had to install the requirements for linux" https://github.com/haskell/haskell-ide-engine#linux-specific-pre-requirements

because i already had them ?!

Anrock:
Gonna check if I really had this options enabled on my Windows box this week.

nponeccop:
> Setting the windows path limitation was not enough.

The Windows path limitation was optional last time I checked. It is configuration-dependent: the user should either ensure that there are no long paths or change a system-wide configuration. The latter is undesirable and/or impossible in controlled environments.

So we should test if it works when cloned into c:\a or if other workarounds such as `subst` work. If they don't then we should say it's mandatory and sorry for those who can't change. But as long as it works it's not a good idea to request long paths unconditionally.

jneira:
Hi, @flip111 thanks for your proposal, with which I agree in general. However i want to clarify some points:
* Agree in removing `optional` but the suggestion already was not technically precise: you can place the project in a path *short enough* to avoid the issue but it has not be the root. I myself have ` D:\dev\ws\haskell\haskell-ide-engine` without long paths enabled and it works fine. So there is a threshold and `optional` tried to capture badly that fact imo.

README.md
```diff
@@ -172,15 +172,15 @@ sudo apt install libicu-dev libtinfo-dev libgmp-dev
...
+In order to avoid problems with long paths on Windows you can do either one of the following:
 
-1. In the `Local Group Policy Editor`: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled` (Works
-   only for Windows 10).
+1. Clone the `haskell-ide-engine` to the root of your logical drive (e.g. to
+   `C:\hie`). If this doesn't work try the second option.
 
-2. Clone the `haskell-ide-engine` to the root of your logical drive (e.g. to
-   `C:\hie`)
+2. In the `Local Group Policy Editor`: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled` (Works
```
jneira: It seems `gpedir.msc` is not present in windows 10 home edition (my version!) so you have to edit directly the registry to enable long paths. Maybe a link to a official microsoft page with instructions (using regedit or gpedit.msc) would be useful: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#enable-long-paths-in-windows-10-version-1607-and-later

README.md
```diff
@@ -172,15 +172,15 @@ sudo apt install libicu-dev libtinfo-dev libgmp-dev
...
+   only for Windows 10). Also you need to configure git to allow longer paths by using unicode paths. To set this for all your git repositories use `git config --system core.longpaths true` (you probably need an administrative shell for this) or for just this one repository use `git config core.longpaths true`.
```
jneira: I guess we should add `Windows Server 2016` as another windows version whick long paths can be enabled

README.md
```diff
@@ -172,15 +172,15 @@ sudo apt install libicu-dev libtinfo-dev libgmp-dev
...
-#### Windows-specific pre-requirements (optional)
+#### Windows-specific pre-requirements
 
-In order to avoid problems with long paths on Windows you can do the following:
+In order to avoid problems with long paths on Windows you can do either one of the following:
 
-1. In the `Local Group Policy Editor`: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled` (Works
-   only for Windows 10).
+1. Clone the `haskell-ide-engine` to the root of your logical drive (e.g. to
+   `C:\hie`). If this doesn't work try the second option.
```
jneira: I would note explicitly that you can place it in a parent directory short enough to not trigger the long path issue.
We have take in account that cabal creates a deep path inside `dist-newstyle` to guess a maximum parent path length.
My longest path i've found withouth being exhaustive f.e. is:
`D:\ws\haskell\haskell-ide-engine\dist-newstyle\build\x86_64-windows\ghc-8.6.5\haskell-ide-engine-0.13.0.0\l\hie-test-utils\build\hie-test-utils\lib\HShaskell-ide-engine-0.13.0.0-inplace-hie-test-utils.a`
with 202 chars so i could have the project dir in a longer path (but no much! :grimacing: )
As names could change i would not use a exact number but something like `around x`

README.md
jneira: Re git the issue is less likely to being triggered cause `dist-newstyle` and `.stack-work` is ignored. So the limit could be even greater than the established by cabal. As you always will hit the compilation problem with a shorter path than the maximum for git i think we could remove the git section.

README.md
jneira: I readed more carefully the link above and [this blog post](https://social.msdn.microsoft.com/Forums/en-US/50ecad58-b8cc-411a-af50-08a67dde5fca/long-name-support-is-not-enabled-on-windows-10?forum=windowssdk) and it seems that you have to enable the option system wide *and* the application have to state in a manifest file that it wants to use the option.
As cabal/stack and ghc are cli applications compiled with ghc i dont know if they would have that option enabled. We should check that setting the option system wide is enoug to let those tools manage a project with a deep path.

flip111: > It seems `gpedit.msc` is not present in windows 10 home edition (my version!) so you have to edit directly the registry to enable long paths. Maybe a link to a official microsoft page with instructions (using `regedit` or `gpedit.msc`) would be useful: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#enable-long-paths-in-windows-10-version-1607-and-later
> 
> Moreover i would note that you have to open `gpedit.msc` to edit policies.

Good find !

jneira: > As cabal, stack and ghc are cli applications compiled with ghc i dont know if they will have that option enabled.

It seems that ghc itself dont use a manifest for windows and the long path issue is still opened so i'am afraid that it will not work :-(
See 
https://gitlab.haskell.org/ghc/ghc/issues/13373
https://gitlab.haskell.org/ghc/ghc/issues/13796

flip111: > so i'am afraid that it will not work :-(

It worked on my system regardless without this manifest

jneira:
@flip111 sorry for spread request changes which are concerns over the original text and no your concrete changes (they improve strictly the original text)
So if you want to merge it as is i'll approve it (we can later make it more precise). Of course if you want to address (or discuss or whatever) them it would be fine too.

README.md
```diff
@@ -172,15 +172,14 @@ sudo apt install libicu-dev libtinfo-dev libgmp-dev
...
-#### Windows-specific pre-requirements (optional)
+#### Windows-specific pre-requirements
 
-In order to avoid problems with long paths on Windows you can do the following:
+In order to avoid problems with long paths on Windows you can do either one of the following:
 
-1. In the `Local Group Policy Editor`: `Local Computer Policy -> Computer Configuration -> Administrative Templates -> System -> Filesystem` set `Enable Win32 long paths` to `Enabled` (Works
-   only for Windows 10).
+1. Clone the `haskell-ide-engine` to a short path, for example the root of your logical drive (e.g. to
+   `C:\hie`). If this doesn't work try the second option.
```
jneira:
```suggestion
   `C:\hie`). If this doesn't work or you want to use a longer path, try the second option.
```

flip111:
yes please approve & merge if you think it's an improvement to the current situation

jneira:
thanks! merging

PR #1464, opened 2019-11-20
Relax bounds of shake and use latest hie resolver

### LINKED ISSUE #1460: Loosen restriction on shake version
I changed this to LTS-14.14 https://github.com/haskell/haskell-ide-engine/blob/master/install/shake.yaml#L2

Then i got

```
$ stack ./install.hs help

Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for hie-install-0.8.0.0(+run-from-stack):
    shake-0.18.3 from stack configuration does not match ==0.17.8  (latest matching version is 0.17.8)
needed since hie-install is a build target.

Some different approaches to resolving this:

  * Set 'allow-newer: true' in C:\sr\config.yaml to ignore all version constraints and build anyway.

  * Recommended action: try adding the following to your extra-deps in C:\Users\philippe\haskell\haskell-ide-engine\install\shake.yaml:

- shake-0.17.8@sha256:d049e261ca6365b6378af6470cb2fca908ced8c5cc862fafc89f00d6baa2f35f,14842

Plan construction failed.
```

Which seems to be declared here https://github.com/haskell/haskell-ide-engine/blob/master/install/hie-install.cabal#L22

Is it absolutely required to have that exact version of the build system? Perhaps it could be changed to require some minimum version instead ??

jneira:
* The new bounds of shake let install it with all supported resolvers of hie so if users want to change `shake.yaml` to match the resolver of hie itself they will be able to do it
  * I still have kept the upper bound cause a new release of shake could break the script
* I've changed the actual resolver to match the latest hie cause i think it will be the most common scenario and this way run the script and install hie will use the same ghc.
* Tested with all `stack-*.yaml` resolvers and `lts-14.14`
* Fixes #1460

### DIFF
--- a/install/hie-install.cabal
+++ b/install/hie-install.cabal
@@ -19,7 +19,7 @@ library
                      , Env
                      , Help
   build-depends:       base >= 4.9 && < 5
-                     , shake == 0.17.8
+                     , shake >= 0.16.4 && < 0.19
                      , directory
                      , filepath
                      , extra
--- a/install/shake.yaml
+++ b/install/shake.yaml
@@ -1,5 +1,5 @@
 # Used to provide a different environment for the shake build script
-resolver: lts-13.18 # GHC 8.6.4
+resolver: lts-14.11 # GHC 8.6.5
 packages:
 - .
 


fendor:
If you match the install script lts snapshots, then installation is not idempotent because hie uses a time extra dep, but the install script uses the bundled to time package.
so, everytime you run the script, the time package will be unregistered and registered repeatedly.

jneira:
Mmm i am checking it but avoiding it changing the lts (and make users have to download two ghc's) doesnt seem the optimal solution. Moreover it will happen with the actual resolver doing a `stack install.hs hie-8.6.4`, right?
I dont know how exactly manage stack the snapshots under `.stack-work` and `.stack` but i think it reuses sometimes them between resolvers with the same ghc.

jneira:
I've just executed `stack install.hs hie-8.6.5` twice with this version and the second one didnt reinstall  any package (not sure if that test trigger the behavior you described)

jneira:
In fact they dont share any snaphot dir:
```
D:\ws\haskell\haskell-ide-engine>stack  --stack-yaml install\shake.yaml path --snapshot-install-root --local-install-root
snapshot-install-root: D:\sr\snapshots\848d866f
local-install-root: D:\ws\haskell\haskell-ide-engine\install\.stack-work\install\d89a2807

D:\ws\haskell\haskell-ide-engine>stack  --stack-yaml stack-8.6.5.yaml path --snapshot-install-root --local-install-root
snapshot-install-root: D:\sr\snapshots\1049e0b6
local-install-root: D:\ws\haskell\haskell-ide-engine\.stack-work\install\4aba8338
```

fendor:
Hm, maybe it has changed because hie-install is now a custom package.

jneira:
So you think it could be merged or there are still some concerns?

PR #1469, opened 2019-11-25
Fix typos; should be non-semantic

bwignall:


jneira:
@bwignall thanks for cleaning up

PR #1472, opened 2019-11-29
Update README.md

Shenmin-Z:
I guess in most cases `zlib` is installed by default, but in case it is not, build would failed as `digest` depends on it.


jneira:
thanks for completing docs!

PR #1476, opened 2019-12-04
README.md: do not hardcode ghc version in example command given in 'I…

smelc:
…nstall specific GHC Version' subsection, to be more beginner friendly

I'm a fresh haskeller and this is the only change needed to get haskell-ide-engine to work flawlessly using the given install procedure (I use stack). Maybe the $(...) command can be made more Windows friendly ? Or maybe you just wanna add a comment, instead of having this full-fledged command ? In any case, I think the concerned example can be enhanced.

And thanks for you hard work on haskell-ide-engine, greatly appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,14 +231,14 @@ stack ./install.hs stack-install-cabal
 Install **Nightly** (and hoogle docs):
 
 ```bash
-stack ./install.hs hie-8.6.4
+stack ./install.hs hie-$(stack exec ghc -- --version | sed 's/.*version //')
 stack ./install.hs build-data
 ```
 
 Install **LTS** (and hoogle docs):
 
 ```bash
-stack ./install.hs hie-8.4.4
+stack ./install.hs hie-$(stack exec ghc -- --version | sed 's/.*version //')
 stack ./install.hs build-data
 ```
 
--- a/README.md
+++ b/README.md
@@ -231,14 +231,14 @@ stack ./install.hs stack-install-cabal
 Install **Nightly** (and hoogle docs):
 
 ```bash
-stack ./install.hs hie-$(stack exec ghc -- --version | sed 's/.*version //')
+stack ./install.hs hie-$(stack exec ghc -- --numeric-version)
 stack ./install.hs build-data
 ```
 
 Install **LTS** (and hoogle docs):
 
 ```bash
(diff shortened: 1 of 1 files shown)


Anrock:
`ghc --numeric-version` can help get rid of sed. Still concerned about Windows friendliness.

fendor:
I also think that this might be wrong. 
When installing hie, you might want to install different versions, not `stack exec -- ghc --version`, which, btw, would default to nightly stack. Also, not applicable to windows.
Moreover, this makes installation instruction very hard to understand. These examples should showcase how to install hie.

smelc:
Following @Anrock's suggestion, I've updated the PR to use `ghc --numeric-version` to get the version instead of `ghc -- --version | sed 's/.*version //'`

I have no enhancement to propose to make the PR more Windows friendly ☹️

@fendor> I think the documentation makes clear already that multiple versions can be installed; experts will understand it. However, I believe it is worth making the given command more likely to work out the box for beginners.

fendor:
Up to that point, there is not a single mention of a correct and complete invocation to install hie.
There is a mention of the help message, thoughm which non-power-users have to read to get more concrete information.

You are not making the command more likely to work with this, now you are bash and unix specific. Even on unix if someone isnt using bash by default, the command fails.
I dont see how this is easier than to enter `hie-8.6.5` directly.

smelc:
@fendor > what is bash-specific in `$(stack exec ghc -- --numeric-version)`?

fendor:
the part `$(...)`. For example, I have fish shell, where this operator does not exist.

jneira:
Nowadays you dont have to set the ghc version!
From `stack install.hs help`:
```
Targets:
    help                   Show help message including all targets

    build                  Build hie with the latest available GHC and the data files
    build-latest           Build hie with the latest available GHC
....
```
For stack the latest ghc is determined by the `stack-*.yaml` files (so actually it will install `hie-8.6.5`). Stack will download the appropiate ghc is it is necessary.

So definitely the installation steps have to be updated but maybe we can remove the distintion between `Nightly` and `LTS` and simply include the default `build` target and an example of how to install an specific versionand the data files (`hie-8.4.4` for example)

fendor:
I think with the latest changes, this PR is not necessary. Feel free to reopen if anyone disagrees.

Issue #1410, opened 2019-10-13
How to support other formatters?


bolt12:
Right now I think hie only supports brittany and floskell. Is it possible to support other formatters via configuration?

fendor:
You have to add another Plugin, e.g. `Ormolu.hs` and register it as a formatter.
Take a look at `Floskell.hs` or `Brittany.hs` for what a formatter plugin needs to support.
Then you have to add the Plugin to basePlugins in `MainHie.hs`.

DavSanchez:
Hi @bolt12 , are you taking care of the issue? I'll be exploring it as well to see if I can contribute to HIE in any way. I'm not that proficient with Haskell, but I got to start somewhere! This looks easy enough at first glance 😃

bolt12:
@DavSanchez Please go ahead! As of now I could try and tackle this issue but it would take me a while to find time.

Maybe someone can assign you this task?

fendor:
@DavSanchez If you run into any problems, feel free to ask here or on irc. At the beginning the Code-base is a bit hard to explore, so if you need any explanations or pointers, please ask!

DavSanchez:
Thanks @fendor ! I'll try to add support for the recent Ormolu formatter, mentioned in your example.

DavSanchez:
Hi guys!

At the moment I have the following code to handle the formatting with `ormolu`, including the handling of a possible `OrmoluException`:

```haskell
provider :: FormattingProvider
provider contents uri typ _opts =
  pluginGetFile contents uri $ \file -> do
    let (range, selectedContents) = case typ of
          FormatText    -> (fullRange contents, contents)
          FormatRange r -> (r, extractRange r contents)

    result <- try @OrmoluException (ormolu defaultConfig file (T.unpack selectedContents))
    case result of
      Left  err -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: " ++ show err) Null)
      Right new -> return $ IdeResultOk [TextEdit range new]
```

I thought that when doing `result <- ...` the `m Text` content of `MonadIO m` get extracted (and, via the `try` function, encapsulated into an `Either OrmoluException Text` that I can use in the next lines to return `IdeResultFail` or `IdeResultOk`. However, the whole function definition gets a red underline and HIE complains (emphasis mine):

> **• Couldn't match type ‘IO’
>                  with ‘Control.Monad.Trans.Reader.ReaderT
[...]

PR #1481, opened 2019-12-17
Ormolu formatter support

DavSanchez:
Hi! This addresses https://github.com/haskell/haskell-ide-engine/issues/1410 for the [Ormolu formatter](https://hackage.haskell.org/package/ormolu)

I still have some work to do, that is sending a warning if the Format Selection option is used (in lines 31-33 of the committed `Ormolu.hs`), as this function is [not yet supported by Ormolu](https://github.com/tweag/ormolu/issues/480#issuecomment-562966292), but I don't know exactly the best way to proceed. I was pointed to `LspStdio.hs` and about throwing a warning on the right thread, but I'm still interpreting the file and how to make use of the functions in it.

Feel free to provide any suggestions.

Also, I think I should put up some tests in `FormatSpec.hs` when `Ormolu.hs` is done, to let everyone check that everything works 😃

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -33,7 +33,6 @@ library
                        Haskell.Ide.Engine.Plugin.HfaAlign
                        Haskell.Ide.Engine.Plugin.HsImport
                        Haskell.Ide.Engine.Plugin.Liquid
-                       Haskell.Ide.Engine.Plugin.Ormolu
                        Haskell.Ide.Engine.Plugin.Package
                        Haskell.Ide.Engine.Plugin.Package.Compat
                        Haskell.Ide.Engine.Plugin.Pragmas
@@ -47,6 +46,9 @@ library
                        Haskell.Ide.Engine.Server
                        Haskell.Ide.Engine.Types
                        Haskell.Ide.Engine.Version
+  if impl(ghc > 8.4)
+     exposed-modules: Haskell.Ide.Engine.Plugin.Ormolu
+
   other-modules:       Paths_haskell_ide_engine
   build-depends:       Cabal >= 1.22
                      , Diff
@@ -82,7 +84,6 @@ library
                      , monoid-subclasses > 0.4
                      , mtl
                      , optparse-simple >= 0.0.3
-                     , ormolu
                      , parsec
                      , process
                      , safe
@@ -101,6 +102,8 @@ library
                      , bytestring-trie
                      , unliftio
                      , hlint >= 2.2.2
+  if impl(ghc > 8.4)
+     build-depends: ormolu
 
   ghc-options:         -Wall -Wredundant-constraints
   if flag(pedantic)
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -33,6 +33,7 @@ library
                        Haskell.Ide.Engine.Plugin.HfaAlign
                        Haskell.Ide.Engine.Plugin.HsImport
                        Haskell.Ide.Engine.Plugin.Liquid
+                       Haskell.Ide.Engine.Plugin.Ormolu
                        Haskell.Ide.Engine.Plugin.Package
                        Haskell.Ide.Engine.Plugin.Package.Compat
                        Haskell.Ide.Engine.Plugin.Pragmas
@@ -81,6 +82,7 @@ library
                      , monoid-subclasses > 0.4
                      , mtl
                      , optparse-simple >= 0.0.3
+                     , ormolu
                      , parsec
                      , process
                      , safe
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -46,7 +46,7 @@ library
                        Haskell.Ide.Engine.Server
                        Haskell.Ide.Engine.Types
                        Haskell.Ide.Engine.Version
-  if impl(ghc > 8.4)
+  if impl(ghc >= 8.6)
      exposed-modules: Haskell.Ide.Engine.Plugin.Ormolu
 
   other-modules:       Paths_haskell_ide_engine
@@ -102,7 +102,7 @@ library
                      , bytestring-trie
                      , unliftio
                      , hlint >= 2.2.2
-  if impl(ghc > 8.4)
+  if impl(ghc >= 8.6)
      build-depends: ormolu
 
   ghc-options:         -Wall -Wredundant-constraints
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -33,6 +33,7 @@ library
                        Haskell.Ide.Engine.Plugin.HfaAlign
                        Haskell.Ide.Engine.Plugin.HsImport
                        Haskell.Ide.Engine.Plugin.Liquid
+                       Haskell.Ide.Engine.Plugin.Ormolu
                        Haskell.Ide.Engine.Plugin.Package
                        Haskell.Ide.Engine.Plugin.Package.Compat
                        Haskell.Ide.Engine.Plugin.Pragmas
@@ -46,8 +47,6 @@ library
                        Haskell.Ide.Engine.Server
                        Haskell.Ide.Engine.Types
                        Haskell.Ide.Engine.Version
-  if impl(ghc >= 8.6)
-     exposed-modules: Haskell.Ide.Engine.Plugin.Ormolu
 
   other-modules:       Paths_haskell_ide_engine
   build-depends:       Cabal >= 1.22
--- a/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
@@ -30,12 +30,8 @@ provider :: FormattingProvider
 provider contents uri typ _opts =
   case typ of 
     -- // TODO Adequately throw an error on the following line
-    FormatRange r -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: " ++ show r) Null)
+    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: Selection formatting not supported.") Null)
     FormatText -> pluginGetFile contents uri $ \file -> do
-        -- INFO: Selection Formatting currently not supported, the below comment is just for temporal information
-        -- let (range, selectedContents) = case typ of
-        --       FormatText    -> (fullRange contents, contents)
-        --       FormatRange r -> (r, extractRange r contents)
         result <- liftIO $ try @OrmoluException (ormolu defaultConfig file (T.unpack contents))
         case result of
           Left  err -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: " ++ show err) Null)
--- a/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
@@ -1,0 +1,42 @@
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Haskell.Ide.Engine.Plugin.Ormolu ( ormoluDescriptor ) where
+
+import Control.Monad.IO.Class ( liftIO, MonadIO(..) )
+import Control.Exception
+import Data.Aeson ( Value ( Null ) )
+import qualified Data.Text as T
+import Ormolu
+import Ormolu.Config (defaultConfig)
+import Ormolu.Exception (OrmoluException)
+import Haskell.Ide.Engine.MonadTypes
+import Haskell.Ide.Engine.PluginUtils
+
+ormoluDescriptor :: PluginId -> PluginDescriptor
+ormoluDescriptor plId = PluginDescriptor
+  { pluginId                 = plId
+  , pluginName               = "Ormolu"
+  , pluginDesc               = "A formatter for Haskell source code."
+  , pluginCommands           = []
+  , pluginCodeActionProvider = Nothing
+  , pluginDiagnosticProvider = Nothing
+  , pluginHoverProvider      = Nothing
+  , pluginSymbolProvider     = Nothing
+  , pluginFormattingProvider = Just provider
+  }
+
+provider :: FormattingProvider
+provider contents uri typ _opts =
+  case typ of 
+    -- // TODO Adequately throw an error on the following line
+    FormatRange r -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: " ++ show r) Null)
+    FormatText -> pluginGetFile contents uri $ \file -> do
+        -- INFO: Selection Formatting currently not supported, the below comment is just for temporal information
+        -- let (range, selectedContents) = case typ of
+        --       FormatText    -> (fullRange contents, contents)
+        --       FormatRange r -> (r, extractRange r contents)
+        result <- liftIO $ try @OrmoluException (ormolu defaultConfig file (T.unpack contents))
+        case result of
+          Left  err -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: " ++ show err) Null)
+          Right new -> return $ IdeResultOk [TextEdit (fullRange contents) new]
--- a/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
@@ -11,7 +11,6 @@ import Control.Exception
 import Control.Monad.IO.Class ( liftIO , MonadIO(..) )
 import Data.Aeson ( Value ( Null ) )
 import Data.Text
-import Haskell.Ide.Engine.MonadFunctions
 import Ormolu
 import Ormolu.Config (defaultConfig)
 import Ormolu.Exception (OrmoluException)
--- a/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
@@ -4,13 +4,12 @@
 
 module Haskell.Ide.Engine.Plugin.Ormolu ( ormoluDescriptor ) where
 
-import Data.Aeson ( Value ( Null ) )
-import Data.Text
-import Haskell.Ide.Engine.MonadTypes
-
 #if __GLASGOW_HASKELL__ >= 806
 import Control.Exception
 import Control.Monad.IO.Class ( liftIO , MonadIO(..) )
+import Data.Aeson ( Value ( Null ) )
+import Data.Text
+import Haskell.Ide.Engine.MonadTypes
 import Ormolu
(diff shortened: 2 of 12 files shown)


fendor:
First, thank you for this PR! It looks very promising.

Is it possible that ormolu does not work with GHC versions below 8.6?

DavSanchez:
Thanks @fendor! Mm yes it’s possible, currently I have only tested it in GHC 8.6.5... I’ll take a look on the Ormolu repo and the failing tests to see if I can make it work there.

fendor:
Yeah, with the base restriction, I dont think it is possible with GHC versions below 8.6.
Which is fine for now, rather focus on the test-suite, if possible!

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
```diff
@@ -0,0 +1,38 @@
...
+  , pluginHoverProvider      = Nothing
+  , pluginSymbolProvider     = Nothing
+  , pluginFormattingProvider = Just provider
+  }
+
+provider :: FormattingProvider
+provider contents uri typ _opts =
+  case typ of 
+    -- // TODO Adequately throw an error on the following line
+    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (T.pack $  "ormoluCmd: Selection formatting not supported.") Null)
```
bubba: Just from looking at the Ormolu API, I think we should be able to support range formatting! The file path is only used "to use it in parse error messages" according to the haddock, so we should be able to extract part of the text and feed it into `ormolu`
You should try the `extractRange` helper function that the other formatting providers use:

https://github.com/haskell/haskell-ide-engine/blob/b8c99d173f8062e633f95370423d7237956798ab/src/Haskell/Ide/Engine/Plugin/Floskell.hs#L37-L39

bubba: Just was pointed out that apparently ormolu doesn't support fragments this way. This is fine then! We can add this if ormolu ever decides to support it

bubba: Nitpick: this is a user facing error, we should probably make it a bit prettier
```suggestion
    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (T.pack $  "Selection formatting for Ormolu is not currently supported.") Null)
```

bubba:
@DavSanchez looks like you'll probably need to add ormolu to the `extra-deps` fields in the stack.yaml's of those failing CI runs

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
DavSanchez: Ah yes, it's a temporal way of handling it, more as a reminder for me while I work on it. The plan is throw a visible alert to the user.

DavSanchez:
> @DavSanchez looks like you'll probably need to add ormolu to the `extra-deps` fields in the stack.yaml's of those failing CI runs

I removed it because `ormolu` is not compatible with these GHC versions due to the versions of `base` used. Is there any course of action for these `stack-8.4.x.yaml` builds to ignore Ormolu for the CI?

fendor:
No, you can not disable the CI for certain tests. 
I *think* you have to add feature flags, e.g. guard the package import to only happen when the ghc version is greater than `8.4` and guard the plugin implementation, too, with `CPP` statements. My suggestion is to use a noop formatter if ormolu is not available and `Log.errorm` messages.

haskell-ide-engine.cabal
```diff
@@ -84,6 +85,7 @@ library
                      , monoid-subclasses > 0.4
                      , mtl
                      , optparse-simple >= 0.0.3
+                     , ormolu
```
fendor: Guard ormolu to only be included if the ghc version is greater than 8.4

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
```diff
@@ -0,0 +1,38 @@
...
+  , pluginHoverProvider      = Nothing
+  , pluginSymbolProvider     = Nothing
+  , pluginFormattingProvider = Just provider
+  }
+
+provider :: FormattingProvider
+provider contents uri typ _opts =
+  case typ of 
+    -- // TODO Adequately throw an error on the following line
+    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (T.pack $  "Selection formatting for Ormolu is not currently supported.") Null)
```
fendor: I dont think we can actually throw an error here. 
If `ormolu` is selected as the formatter, `hsimport` tries to use its `FormatRange` type to format only the import line.
See the test cases `hsimport` with formatter to see an example when the formatter is used.

bubba: We can just add a case for this in hsimport for now. But what we should really do is change the server capabilities on initialisation/configuration change to say that we don’t support range formatting anymore

bubba: Specifically, we should probably take advantage of https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#client_unregisterCapability

DavSanchez:
> No, you can not disable the CI for certain tests.
> I _think_ you have to add feature flags, e.g. guard the package import to only happen when the ghc version is greater than `8.4` and guard the plugin implementation, too, with `CPP` statements. My suggestion is to use a noop formatter if ormolu is not available and `Log.errorm` messages.

Ah, I didn't know you could insert C-like guards simply like that in Haskell code. I'll take a look and update the code accordingly. I'll also check `Log.errorm`.

As for the tests, yeah I don't have them at the moment, I'll add them during the holidays.

DavSanchez:
> I _think_ you have to add feature flags, e.g. guard the package import to only happen when the ghc version is greater than `8.4` and guard the plugin implementation, too, with `CPP` statements.

Hi @fendor, if I guard the package imports in `haskell-ide-engine.cabal` is it really needed to also guard `Ormolu.hs` itself? If I guard just the module import at `MainHie.hs` does the file get compiled anyway, even if it's not used by any target file in `haskell-ide-engine.cabal`?

> My suggestion is to use a noop formatter if ormolu is not available and `Log.errorm` messages.

At the moment my attempt just does `return $ IdeResultOk [TextEdit (fullRange contents) contents]`. Since this does an actual replacement (with exactly the same contents), is there any other way of doing a true `noop` without messing with the types or is that the correct way to go?

haskell-ide-engine.cabal
DavSanchez: Done!

fendor:
@DavSanchez I think it would be better if we guard the functions in Ormolu, since we have to add the available formatters to the lsp-client, seemingly. iirc https://github.com/alanz/vscode-hie-server/blob/954e0a9ecc7be4d96bc71d05c65a730e642b65a6/package.json#L87.
We probably have to add an entry to this list of formatters, but we dont know if the ormolu formatter is really available since this depends on the ghc version the lsp-server has been compiled with. That is why I *think*, guarding just the functions and the imports (also the dependency like you already did) in `Ormolu.hs` is better. We still want to inform the user somehow that this doesnt work. 

Maybe `return $ IdeResultOk []` is already enough? Then there is no text-edit.

DavSanchez:
Hi again! A few questions.

In the test suite (`FormatSpec.hs` file), how do I control what formatter is selected for the appropriate test? I see that it tests whether the formatting provider changes and also see `brittany`-specific test suite, but for these last test I don't see where `brittany` is selected. 

How can I assert that a warning or error was emitted? This is for checking that a format selection action triggers the correct message.

Also, should I make any tests for the noop formatter used when GHC < 8.6?

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
DavSanchez: Can I have some more assistance on what would I need to do here? I mean, about checking `hsimport` and adding a case to it.

fendor: I would say, dont worry about it. Register/unregister an capability is something to be done in haskell-lsp. Not sure if that is already implemented, but it is definitely out-of-scope for this pr. 
I would say, that a noop formatter and a log message is fine for this PR.

fendor:
@DavSanchez Sorry for the late response, I overlooked the notification.
To switch the formatter the following can be used: 
```haskell
sendNotification WorkspaceDidChangeConfiguration (DidChangeConfigurationParams (formatLspConfig "ormolu"))
formatDoc doc (FormattingOptions 2 True)
documentContents doc >>= liftIO . (`shouldBe` ormolu)
```
For example, see the test-case in https://github.com/haskell/haskell-ide-engine/blob/master/test/functional/FormatSpec.hs#L49.

You will test the noop formatter anyways, since the test-cases need to work across all CI jobs.
For example, you can check the ghc version via https://github.com/haskell/haskell-ide-engine/blob/master/test/utils/TestUtils.hs#L146 (this avoids using CPP instructions).
then just do:
````haskell
formatDoc doc (FormattingOptions 2 True)
docContent <- documentContents doc
case ghcVersion of 
  GHC86 -> liftIO $ docContent  `shouldBe` formattedOrmolu
  _ -> liftIO $ docContent  `shouldBe` unchangedDoc
```` 

Tests relevant for hsimport are here: https://github.com/haskell/haskell-ide-engine/blob/master/test/functional/FunctionalCodeActionsSpec.hs#L529
You can see an example call of this function here: https://github.com/haskell/haskell-ide-engine/blob/master/test/functional/FunctionalCodeActionsSpec.hs#L136
In this case, you wont need `ghcVersion` since we do not support that feature, anyways.
Calling the function `hsImportSpec` correctly, should be enough to test and document the `hsimport` interaction with ormolu we mentioned above.

hope this helps! (btw, did not try to compile any of the code snippets, might contain small errors but the idea is *hopefully* right)

**Comment by bubba** [comment]
@fendor

**Comment by DavSanchez** [comment]
Looks like the tests for `FormatSpec.hs` are passing. Just on time for 2020 at my place!

Happy new year everyone! 🎉

File: src/Haskell/Ide/Engine/Plugin/Ormolu.hs [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+          Right new -> return $ IdeResultOk [TextEdit (fullRange contents) new]
+#else
+-- Work in progress
+provider :: FormattingProvider
+provider _contents _uri typ _opts = do
+  errorm "This version of HIE does not support Ormolu as a formatter"
+  case typ of 
+    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (pack "Selection formatting for Ormolu is not currently supported.") Null)
+    FormatText -> return $ IdeResultOk []
+#endif
```
fendor: I think we did not mention explicitly, but usually you add the CPP statements to exactly the statements you want to guard. E.g.
```haskell
provider :: FormattingProvider
provider contents uri typ _opts =
  case typ of 

#if __GLASGOW_HASKELL__ >= 806
    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (pack "Selection formatting for Ormolu is not currently supported.") Null)
    FormatText -> pluginGetFile contents uri $ \file -> do
        result <- liftIO $ try @OrmoluException (ormolu defaultConfig file (unpack contents))
        case result of
          Left  err -> return $ IdeResultFail (IdeError PluginError (pack $  "ormoluCmd: " ++ show err) Null)
          Right new -> return $ IdeResultOk [TextEdit (fullRange contents) new]
#else 
    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (pack "Selection formatting for Ormolu is not currently supported.") Null)
    FormatText -> return $ IdeResultOk []
#endif

```

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
```diff
@@ -0,0 +1,53 @@
...
+        case result of
+          Left  err -> return $ IdeResultFail (IdeError PluginError (pack $  "ormoluCmd: " ++ show err) Null)
+          Right new -> return $ IdeResultOk [TextEdit (fullRange contents) new]
+#else
+-- Work in progress
+provider :: FormattingProvider
+provider _contents _uri typ _opts = do
+  errorm "This version of HIE does not support Ormolu as a formatter"
+  case typ of 
+    FormatRange _ -> return $ IdeResultFail (IdeError PluginError (pack "Selection formatting for Ormolu is not currently supported.") Null)
```
fendor: This should be like FormatText a noop if it is not supported.

DavSanchez: What about notifying about lack of support for GHC < 8.6? (The use of `errorm` included in this snippet.)

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
DavSanchez: Ok I’ll correct this!

DavSanchez:
Yeah, I’ll implement them in the next hours.

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
fendor: That should be done somewhere else imo, I am not sure where, yet, but not here. @bubba would be server.hs fine?

fendor:
Note, you will have problems with your implementation for format selection. I suggest, until @bubba disagrees, that format selection is a noop and that a warning is added to `Server.hs` when the explicit request for formatting a selection is received

DavSanchez:
Hi @fendor, I'm running into problems with some of the test cases for `HsImport`, specifically the ones where the file analysed has a Haddock comment in it. That's because of https://github.com/tweag/ormolu/issues/480

Should I update the files used by `hsImportSpec`, including a `module` declaration inside them, and fix the tests affected by the change? or maybe do something else?

fendor:
@DavSanchez This test exsits for exactly this reason, to find out about problems :) Please dont delete the haddock, we dont want to just remove failing test-cases.

This doesnt make sense to me, the test-case is a valid Haskell module with a main function in it. Does that mean ormolu does not work on the main module? 
EDIT: ah, ghc parse error, a ghc bug.

Yeah, ideally, we have both, assert somehow the parse error (maybe in FormatSpec.hs) and add another test-case with the module declaration.

However, how does that actually affect the hsimportSpec, since Omolu cant format only the import anyways, shouldnt the import just be added without a parse error?

DavSanchez:
Thanks @bubba!

@fendor, yeah I suppose it doesn’t affect `hsImportSpec`, it would only add the imports without formatting them the cases affected, I just thought that not asserting the parse error or something can somehow defeat the purpose of the test, but anyway I’ll, add the test case and then check out `Server.hs` to see where I can emit the appropriate warnings.

DavSanchez:
Alright, the tests for `hsImport` pass for all GHC >= 8.6, for versions lower than that they timeout. The message is the same for the 4 tests in all cases:

> Timed out waiting to receive a message from the server. Last message received: {     "tag": "RspExecuteCommand",     "contents": {         "result": {},         "jsonrpc": "2.0",         "id": 9     } }

I suppose the noop formatter is not working correctly? 😕 I guess I could CPP guard the test, but then it would be cheating 😄

alanz:
You could CPP guard it to something like `pendingWith "Need to revisit this"`, or something like that.

DavSanchez:
Yeah but, what is exactly happening here? I would assume that the test failed for the first two cases, since the expected results are formatted and Ormolu in GHC < 8.6 is a NOP, but why the timeout instead? Am I missing something?

Avi-D-coder:
We should pass `default-extensions` and `ghc-options` in ormolu `Config.cfgDynOptions`

DavSanchez:
> We should pass `default-extensions` and `ghc-options` in ormolu `Config.cfgDynOptions`

You mean adding them to the Cabal and/or Stack files for HIE or loading Ormolu as a submodule and doing it there? Why is it necessary?

fendor:
@DavSanchez while @Avi-D-coder is right, we will postpone this for now. None of the other formatters is doing this correctly at the moment, anyways, as far as I know. We can improve it iteratively. 
I will do a final review tomorrow and after addressing found issues (if there are any), I think we are ready for a merge.

Avi-D-coder:
@DavSanchez Ormolu formats code differently if certain extensions or ghc options are enabled. E.g `TypeApplications`. They can be enabled cradle wide in cabal and other build files, the project wide options need to be passed to the formatter so formatting does not change the codes semantics.

Avi-D-coder:
@fendor do we cache ghc flags anywhere or should we just call whatever function `hie-bios flags $filePath` does?

fendor:
@Avi-D-coder We are currently not caching the results of `hie-bios flags $filepath` and just calling this function would yield very slow performance.
But we are caching the `HscEnv` for each cradle and transitively, we have access to `DynFlags` which contain the exact flags used to compile the file. However, the `FormattingProvider` does not permit us to look up the cache at the moment, if I read that right. We need to refactor that, so that we have access to the `DynFlags`.

Avi-D-coder:
@fendor I'm not aware of a `DynFlags` to `[String]` conversion. Ormolu could be modified to provide a version that takes  `DynFlags` instead of  `[String]`.

At https://github.com/Avi-D-coder/haskell-ide-engine/commit/baa12ac464f5c5e0d663d02bf037bd57d4ce08c4 I call `findLocalCradle`; it's a bit slow, but it does the job.

fendor:
Using `findLocalCradle` is horribly slow since you are using either `cabal repl` or `stack repl` to load the project. I dont know how Ormolu works, but I can not imagine just passing in all the ghc-options makes sense, but rather parse out the Language Extensions we care about and passing them to Ormolu.

Avi-D-coder:
@fendor currently I'm filtering everything except plugins, preprocessor and extensions. 

Unless I missed something last time I took a look at getting strings out of  `DynFlags` it would require abunch of manually mappings. Perhaps we could cache the GHC option strings directly?

test/functional/FunctionalCodeActionsSpec.hs
```diff
@@ -218,6 +219,51 @@ spec = describe "code actions" $ do
...
+        , "    when True"
+        , "        $ hPutStrLn stderr"
+        , "        $ fromMaybe \"Good night, World!\" (Just \"Hello, World!\")"
+        ]
+      ]
+#else 
+    describe "formats with ormolu" $
+      it "is NOP formatter" $
+        pendingWith "Ormolu only supported by GHC >= 8.6. Need to restore this."
+#endif
```
fendor: I think we can avoid here CPP expression by reusing `ghcVersion`.

DavSanchez: Yeah, I thought the same. Solving it now!

fendor:
@Avi-D-coder Caching the ghc options directly is possible, too.

Avi-D-coder:
@alanz ping

DavSanchez:
Thanks for all the help guys! It’s been a great ride contributing with Haskell. A lot to learn yet...

jneira:
@DavSanchez thank you for your contribution 👌

PR #1485, opened 2019-12-21
Update Ubuntu dependency

leifmetcalf:
libtinfo-dev has been replaced by libncurses-dev since Ubuntu 18.10


alanz:
Do we need to keep both, and list the ubuntu versions for each?

leifmetcalf:
Ubuntu 18.04 is still the current LTS version, so I've added a section listing the build requirements for distros pre- Ubuntu 18.10/Debian 10.

PR #1489, opened 2019-12-22
Remove JSON transport

bubba:
A fresh take on #1105 

Stuff to do
- [x] Remove plugin descriptions, they're no longer needed


src/Haskell/Ide/Engine/Support/Hoogle.hs
```diff
@@ -22,27 +22,6 @@ import           Text.HTML.TagSoup.Tree
 
 -- ---------------------------------------------------------------------
 
-hoogleDescriptor :: PluginId -> PluginDescriptor
```
fendor: Hoogle Plugin does not exist anymore? Out of curiosity, what does that mean? Was it previously possible to invoke the PluginCommands "info" and "lookup" without a hover request? Or is this change unobservable for the lsp client?

src/Haskell/Ide/Engine/Support/Hoogle.hs
```diff
@@ -207,24 +179,14 @@ searchTargets f term = do
...
-    Right x -> IdeResultOk x
-
--- | Lookup the given Text in the local Hoogle database.
--- Takes the first `n` matches.
--- May fail with a HoogleError that can be shown to the user.
-lookupCmd' :: Int -> T.Text -> IdeM (Either HoogleError [T.Text])
-lookupCmd' n term = do
+-- | 'lookup' @n term@ looks up the given Text in the local Hoogle database.
+-- Takes the first @n@ matches.
+  -- May fail with a HoogleError that can be shown to the user.
```
fendor: Nitpick: Indentation a bit off

src/Haskell/Ide/Engine/Version.hs
```diff
@@ -1,93 +1,26 @@
...
+import qualified Data.Text                       as T
+import qualified Data.Versions                   as V
 import           System.Directory
 import           System.Info
 import           System.Process
-import qualified System.Log.Logger as L
-
--- ---------------------------------------------------------------------
-
-baseDescriptor :: PluginId -> PluginDescriptor
```
fendor: Same question as for hoogle plugin. What does this entail for lsp-clients?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -274,17 +278,16 @@ data PluginDescriptor =
...
-newtype CommandFunc a b = CmdSync (a -> IdeGhcM (IdeResult b))
+newtype CommandId = CommandId T.Text
+  deriving (Show, Read, Eq, Ord)
+instance IsString CommandId where
+  fromString = CommandId . T.pack
 
 data PluginCommand = forall a b. (FromJSON a, ToJSON b, Typeable b) =>
-  PluginCommand { commandName :: CommandName
-                , commandDesc :: T.Text
-                , commandFunc :: CommandFunc a b
```
fendor: Why is the description removed and the `CommandFunc`? Are these remnants from JsonStdio? Is the description displayed somewhere for lsp-clients?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -274,17 +278,16 @@ data PluginDescriptor =
                    } deriving (Generic)
 
 instance Show PluginCommand where
-  show (PluginCommand name _ _) = "PluginCommand { name = " ++ T.unpack name ++ " }"
-
-type PluginId = T.Text
-type CommandName = T.Text
+  show (PluginCommand i _) = "PluginCommand { name = " ++ show i ++ " }"
 
-newtype CommandFunc a b = CmdSync (a -> IdeGhcM (IdeResult b))
```
fendor: Why is this removed? Dont we need it anymore? Just for understanding the PR.

app/MainHie.hs
```diff
@@ -50,17 +47,14 @@ plugins includeExamples = pluginDescToIdePlugins allPlugins
...
       , haddockDescriptor     "haddock"
       -- , hareDescriptor        "hare"
-      , hoogleDescriptor      "hoogle"
       , hsimportDescriptor    "hsimport"
       , liquidDescriptor      "liquid"
       , packageDescriptor     "package"
       , pragmasDescriptor     "pragmas"
       , floskellDescriptor    "floskell"
-      , biosDescriptor        "bios"
       , genericDescriptor     "generic"
```
alanz: Why are we removing the base, hoogle and bios plugins?

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -41,12 +41,9 @@ type HintTitle = T.Text
 applyRefactDescriptor :: PluginId -> PluginDescriptor
 applyRefactDescriptor plId = PluginDescriptor
   { pluginId = plId
-  , pluginName = "ApplyRefact"
-  , pluginDesc = "apply-refact applies refactorings specified by the refact package. It is currently integrated into hlint to enable the automatic application of suggestions."
   , pluginCommands =
-      [ PluginCommand "applyOne" "Apply a single hint" applyOneCmd
-      , PluginCommand "applyAll" "Apply all hints to the file" applyAllCmd
-      , PluginCommand "lint" "Run hlint on the file to generate hints" lintCmd
```
alanz: Why is this being removed?

src/Haskell/Ide/Engine/Plugin/Package.hs
```diff
@@ -87,15 +85,8 @@ type Package = T.Text
 -- May fail if no project dependency specification can be found.
 -- Supported are `*.cabal` and `package.yaml` specifications.
 -- Moreover, may fail with an IOException in case of a filesystem problem.
-addCmd :: CommandFunc AddParams J.WorkspaceEdit
-addCmd = CmdSync addCmd'
-
--- | Add a package to the project's dependencies.
--- May fail if no project dependency specification can be found.
--- Supported are `*.cabal` and `package.yaml` specifications.
--- Moreover, may fail with an IOException in case of a filesystem problem.
```
alanz: Should this be removed?

src/Haskell/Ide/Engine/Support/Hoogle.hs
bubba: Through LSP no, it was for when the JSON transport could run plugin commands directly. The functionality is still there in the `lookup`/`info` cmds, just not registered as a command.

app/MainHie.hs
bubba: The base plugin was only used by the JSON transport to query existing plugins, LSP can't access them

```
      [ PluginCommand "version" "return HIE version" versionCmd
       , PluginCommand "plugins" "list available plugins" pluginsCmd
       , PluginCommand "commands" "list available commands for a given plugin" commandsCmd
       , PluginCommand "commandDetail" "list parameters required for a given command" commandDetailCmd
```

Same goes for Hoogle and Bios, they only exposed commands that LSP can't access, so it didn't make much sense to keep them around as plugins

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
bubba: Descriptions and name were only used for the plugin querying commands in base, which aren't accessible through LSP anymore. `PluginCommand` should now be isomorphic to the LSP commands

bubba: I removed `CommandFunc` just to simplify things. I think it's more illustrative to have the function type directly visible inside `PluginCommand` and it removes the redundant

```haskell
xCmd :: CmdFunc a b
```
```haskell
xCmd' :: a -> IdeGhcM (IdeResult b)
```

pattern

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
bubba: The lint command isn't accessible via LSP, i.e. `Server.hs` just manually calls `lint` to send off the diagnostic notifications, and it doesn't make any sense for this to be a command in LSP (I don't think commands can really return anything or than workspace edits?). `applyOne`/`applyAll` are still there since code actions can call them

src/Haskell/Ide/Engine/Version.hs
bubba: This has no change for LSP clients, it was only used for querying information about plugins from the JSON transport and isn't accessible via LSP

src/Haskell/Ide/Engine/Plugin/Package.hs
bubba: This is one of the benefits of removing the `CommandFunc` wrapper, as it allows us to coalesce all the `xCmd` `xCmd'` functions into just one. This should be an NFC

src/Haskell/Ide/Engine/Version.hs
```diff
@@ -1,93 +1,26 @@
-{-# LANGUAGE CPP                   #-}
-{-# LANGUAGE DataKinds             #-}
-{-# LANGUAGE GADTs                 #-}
-{-# LANGUAGE OverloadedStrings     #-}
-{-# LANGUAGE PartialTypeSignatures #-}
-{-# LANGUAGE TemplateHaskell       #-}
-module Haskell.Ide.Engine.Plugin.Base where
+{-# LANGUAGE CPP, TemplateHaskell, OverloadedStrings #-}
```
alanz: Why are you collapsing the pragmas?

bubba: I am not sure. I think this came in when shuffling parts about from the old PR. Will fix

PR #1492, opened 2019-12-22
Refactor plugins and commands now that JSON transport is gone

bubba:
- Remove the name and description fields, since they were only used by the JSON transport when querying and can't be seen by LSP
- Similarly remove the bios plugin as it can't be called by LSP, so there's no point in it being there
- Remove `CommandFunc` new type so we can coalesce the cmd/cmd' pattern into just one

```haskell
xCmd :: CmdFunc a b
```
```haskell
xCmd' :: a -> IdeGhcM (IdeResult b)
```


src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -41,12 +41,9 @@ type HintTitle = T.Text
 applyRefactDescriptor :: PluginId -> PluginDescriptor
 applyRefactDescriptor plId = PluginDescriptor
   { pluginId = plId
-  , pluginName = "ApplyRefact"
-  , pluginDesc = "apply-refact applies refactorings specified by the refact package. It is currently integrated into hlint to enable the automatic application of suggestions."
   , pluginCommands =
-      [ PluginCommand "applyOne" "Apply a single hint" applyOneCmd
-      , PluginCommand "applyAll" "Apply all hints to the file" applyAllCmd
-      , PluginCommand "lint" "Run hlint on the file to generate hints" lintCmd
```
fendor: Why is "lint" removed?

bubba: It can't be accessed by LSP. Previously old IDEs communicating via the JSON transport would call this to get back a list of diagnostics. Now we just manually call it to send the publish diagnostic notifications in Server.hs

alanz: Is it still possible to send a code action that maps back to this? or to any of the others that have been removed?

In particular, when `HaRe` is re-enabled, it sends commands which need to be activated when returned.

bubba: The applyOne/applyAll ones are still there which hlint code actions can call, but as far as I can tell the lint command doesn’t really work with LSP since the only thing commands can do are return document edits?

alanz: Ok, I guess we then have the concept visible, as actually used. So the currently unused ones can go.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -185,12 +186,12 @@ mkLspCommand plid cn title args' = do
...
-    go (plid, cmds) = mapM (mkLspCmdId plid . commandName) cmds
+    go (plid, cmds) = mapM (mkLspCmdId plid . commandId) cmds
 
-mkLspCmdId :: HasPidCache m => PluginId -> CommandName -> m T.Text
-mkLspCmdId plid cn = do
+mkLspCmdId :: HasPidCache m => PluginId -> CommandId -> m T.Text
+mkLspCmdId plid cid = do
   pid <- T.pack . show <$> getPidCache
-  return $ pid <> ":" <> plid <> ":" <> cn
+  return $ pid <> ":" <> coerce plid <> ":" <> coerce cid
```
alanz: Is `coerce` the best way to do this? Are we losing clarity? Could we not use a newtype?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -325,11 +328,6 @@ newtype IdePlugins = IdePlugins
   { ipMap :: Map.Map PluginId PluginDescriptor
   } deriving (Generic)
 
--- TODO:AZ this is a defective instance, do we actually need it?
--- Perhaps rather make a separate type explicitly for this purpose.
-instance ToJSON IdePlugins where
-  toJSON (IdePlugins m) = toJSON $ fmap (\x -> (commandName x, commandDesc x)) <$> fmap pluginCommands m
-
```
alanz: :+1:

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
bubba: Yes, and the actual functionality is still there in the lint command which is just called manually

src/Haskell/Ide/Engine/Plugin/HfaAlign.hs
```diff
@@ -28,10 +28,8 @@ import qualified Safe
 hfaAlignDescriptor :: PluginId -> PluginDescriptor
 hfaAlignDescriptor plId = PluginDescriptor
   { pluginId = plId
-  , pluginName = "Align Equals"
-  , pluginDesc = "An example of writing an HIE plugin\nbased on http://www.haskellforall.com/2018/10/detailed-walkthrough-for-beginner.html"
```
alanz: Can you convert the pluginDesc into a comment in the file? It gives context, for how to take a feature description and implement it in hie. Which is the point of the example.

src/Haskell/Ide/Engine/Server.hs
```diff
@@ -855,22 +854,23 @@ requestDiagnostics DiagnosticsRequest{trigger, file, trackingNumber, documentVer
       forM_ dss $ \(pid,ds) -> do
         debugm $ "requestDiagnostics: calling diagFunc for plugin:" ++ show pid
         let
+          pid' = coerce pid
```
alanz: Leaves me uncomfortable, as before.
If we decide to keep `coerce`, perhaps add a signature to this function, to document what the conversion is.

src/Haskell/Ide/Engine/Plugin/HfaAlign.hs
bubba: That’s a good point. I’ll insert it as a haddock doc for the module

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
bubba: Yes, the `coerce` is definitely a bit hairy. I'm a bit uncomfortable passing about raw `Texts`s since generally these ids shouldn't be mutated, so I'd like to keep the newtype if possible. I'll replace the `coerce` with a `let` binding instead?

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -261,10 +261,13 @@ type FormattingProvider = T.Text -- ^ Text to format
...
 
+newtype PluginId = PluginId T.Text
+  deriving (Show, Read, Eq, Ord)
+instance IsString PluginId where
+  fromString = PluginId . T.pack
+
 data PluginDescriptor =
   PluginDescriptor { pluginId                 :: PluginId
-                   , pluginName               :: T.Text
-                   , pluginDesc               :: T.Text
```
mpickering: I don't see why removing these fields is desirable at all. Surely it is useful to be able to inspect a generic plugin, even if for debugging purposes.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -309,16 +308,18 @@ toDynJSON = CD.toDyn
 runPluginCommand :: PluginId -> CommandId -> Value
                   -> IdeGhcM (IdeResult DynamicJSON)
 runPluginCommand p com arg = do
+  let PluginId p' = p
+      CommandId com' = com
```
alanz: Thanks, but a minor nit. For `mkLspCmdId` just above you match the `PluginId` and `CommandId` directly as parameters.  Why is this different? if you need the original, perhaps do it as

```haskell
runPluginCommand p@(PluginId p') com@(CommandId cid) arg = do
```

mpickering:
I am strongly against removing `pluginName` and `pluginDescription` fields, it seems like a step in the wrong direction.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
bubba: These are just vanity names/descriptions, which aren’t accessible to the user at all via LSP. For debugging you’ll still always be able to read the pluginId which is like a string “applyRefact” or “floskell”, which is basically the same as the plugin name

alanz:
> I am strongly against removing pluginName and pluginDescription fields, it seems like a step in the wrong direction.

I agree with @mpickering. They are not needed in the current use-case, but the intention has always been to have some kind of discoverability around plugins, and the ability to use alternative ones, e.g. choose the one that supports the formatter you prefer. Having names and descriptions makes this easier.

bubba:
Ok, I will amend this PR later this evening. I’m not sure what options LSP gives us for discovering different plugins, at the moment for choosing the formatter plugins it is just a matter of setting a string inside the LSP config

alanz:
@bubba Thanks. That capability does not exist at present, but I think keeping the info in at the moment is worth doing, we can strip it later if we choose to.

Somewhere along the line I would like to see a much more loosely coupled plugin architecture, which allows us to configure per installation, or per project what plugins should be used.

Having meta information available helps with this. And it is likely it may move to a different place, but lets keep the flame alive in the meantime.

Which directly opposes YAGNI, but anyway.

Issue #1490, opened 2019-12-22
HieWrapper discovers the wrong ghc version in a multi-stack cradle


fendor:
In a project, with a hie.yaml such as
````yaml
cradle:
  stack:
    - path: .
      component: "package:lib"
````
Hie will identify this as a "Multi-Cradle". 
A hie.yaml such as:
````yaml
cradle:
  cabal:
    - path: .
      component: "lib:package"
````
Will also be identified as a "Multi-Cradle".

Haskell IDE Engine can not tell the difference between a `"Cabal"-Multi-Cradle` and a `"Stack"-Multi-Cradle`, but it should.

Moreover, I suggest that the following should be handled gracefully:

```yaml
cradle:
  multi:
    - path: .
      config:
        cradle:
          cabal: 
            component: "lib:package"
    - path: examples/
      config:
        cradle:
          none:
```

I think, that this still should be identified as a `Cabal-Multi-Cradle` and analogously the `Stack-Multi-Cradle`.
Support for this feature needs to be implemented in `hie-bios`.

cc @mpickering 

Issue is encountered in #1482.

fendor:
Only closed if hie-bios gets updated.

Issue #1431, opened 2019-10-29
Make cabal a first class citizen for CI


fendor:
CI hardly uses cabal, although it is one of the two supported installation methods.
For all supported GHC Versions, we should add a CI option to build HIE with cabal.
Also, we should run the tests with cabal as well.

jneira:
I am adding cabal jobs with tests for ghc-8.6.5 and ghc-8.4.4: https://github.com/haskell/haskell-ide-engine/pull/1571
@fendor It would be enough to close this or you think all supported ghcs will be needed?
or maybe it should be added to circle ci...

fendor:
Good enough for me! Tests have still a run-time dependency on stack, right?

jneira:
Yeah, the `CabalHelper` unit tests fails if stack (or cabal) is not in path

PR #1496, opened 2019-12-23
Find the libdir directory of ghc at run-time

### LINKED ISSUE #1495: HIE crashes if there is no "ghc" on the path with explicit stack config
Use something like `showMessage` to report that GHC is not available, and suggest installing it, when starting hie.

This would help in cases such as https://github.com/alanz/vscode-hie-server/issues/181

fendor:
Fixes #1495 

Finds the libdir directory of the appropriate ghc version at run-time. 
Introduces the `isCabalCradle` predicate to complement the `isStackCradle`

* [x] Documentation is missing.
* [x] `getProjectGhcVersion` can be expressed with `execProjectGhc`

While it works in manual test-cases I have the following considerations:
* Cradle initialisation happens in too many places. At least the one in "NotInitialized" can be skipped, imo.
* We might select the wrong ghc version in certain circumstances.
* I dont know what happens if the system lib dir is `Nothing`

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -81,16 +81,26 @@ getProjectGhcPath :: Cradle -> IO (Maybe FilePath)
 getProjectGhcPath crdl = do
   isStackInstalled <- isJust <$> findExecutable "stack"
   isCabalInstalled <- isJust <$> findExecutable "cabal"
-  if isStackCradle crdl && isStackInstalled
+  ghcpath <- if isStackCradle crdl && isStackInstalled
     then
-      catch (Just <$> tryCommand "stack path --compiler-exe") $ \(_ :: IOException) ->
+      catch (Just <$> tryCommand "stack path --compiler-exe") $ \(_ :: IOException) -> do
+        errorm "Command `stack path --compiler-exe` failed."
         return Nothing
     else if isCabalCradle crdl && isCabalInstalled then do
-      Just ghcCabalVersion <- catch (Just <$> tryCommand "cabal v2-exec ghc -- --numeric-version") $ \(_ ::IOException) ->
+      ghcCabalVersion <- catch (Just <$> tryCommand "cabal v2-exec -v0 ghc -- --numeric-version") $ \(_ ::IOException) -> do
+        errorm "Command `cabal v2-exec -v0 ghc -- --numeric-version` failed."
         return Nothing
-      findExecutable ("ghc-" ++ ghcCabalVersion)
-    else
+      case ghcCabalVersion of
+        Just ghcNumericVersion -> do
+          let ghcVersion = "ghc-" ++ ghcNumericVersion
+          logm $ "Ghc Version to find: " ++ ghcVersion
+          findExecutable ghcVersion
+        Nothing -> return Nothing
+    else do
+      logm "Neither cabal nor stack project, look for ghc project."
       findExecutable "ghc"
+  logm $ "Found ghc path: " ++ show ghcpath
+  return ghcpath
 
 tryCommand :: String -> IO String
 tryCommand cmd =
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -26,11 +26,11 @@ import           Data.Ord (Down(..))
 import           Data.String (IsString(..))
 import qualified Data.Text as T
 import           Data.Foldable (toList)
-import           Control.Exception (IOException, try, catch)
+import           Control.Exception
 import           System.FilePath
 import           System.Directory (getCurrentDirectory, canonicalizePath, findExecutable)
 import           System.Exit
-import           System.Process (readCreateProcess, shell)
+import           System.Process (readCreateProcessWithExitCode, shell)
 
 -- | Find the cradle that the given File belongs to.
 --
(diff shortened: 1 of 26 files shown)


hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -57,6 +58,54 @@ isStackCradle = (`elem` ["stack", "Cabal-Helper-Stack", "Cabal-Helper-Stack-None
...
+tryCommand :: String -> IO String
+tryCommand cmd =
+  T.unpack . T.strip .T.pack <$> readCreateProcess (shell cmd) ""
+
+getProjectGhcLibDir :: Cradle -> IO (Maybe FilePath)
+getProjectGhcLibDir crdl = do
+  mGhcPath <- getProjectGhcPath crdl
+  case mGhcPath of
+    Nothing -> return Nothing
+    Just ghcPath -> catch (Just <$> tryCommand (ghcPath ++ " --print-libdir")) $ \(_ :: IOException) -> return Nothing
```
alanz: Catching and discarding the exception is probably a bad thing.

At the very least we should log something on failure.

Is the the exception that would be thrown if GHC is not installed?  Do we report it gracefully elsewhere if that is the case?

fendor: No, not yet. Not sure how to properly report it, since it is executed before the server is actually started, e.g. When we start the GHC thread

alanz: Perhaps we should re-look at this startup sequencing, and launch the various processes once the initialize message comes in, while processing it.

That way we can determine the cradle according to the passed-in project root, and we have a way to report errors.  The spec says we are not to process anything before initialization is complete anyway, so it *is* an appropriate place to do it.

cc @bubba @mpickering

bubba: Yes, I think the LSP server should be bootstrapped before the dispatcher. And while we're at it, we should provide an easier way to log/show these messages via LSP. Presumably some sort of transformer stack hierarchy:

```
IO a < LspM a < IdeM a < IdeGhcM a
```

fendor: Would that mean to execute `onStartup` https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/Server.hs#L125 in `NotInitialized`?

alanz: No, `onStartup` is the one that is called when the initialize request comes in. At this point haskell-lsp has comms established with the client, and has received a capability list from it, as well as the root directory.

So we can find the libdir and launch the dispatchers from here, and if there is a problem send back a message which will be displayed on the client.

alanz: See https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#initialize

alanz: And perhaps we should rename `onStartup` to `onInitialize`, and make it clear where in the server life cycle it operates.

fendor: Oh, so that is actuallly fine as-is? Just need to refactor when the cradle is loaded?

alanz: yes

alanz: And this change should also resolve https://github.com/alanz/vscode-hie-server/issues/164

fendor: Not entirely. For hie-bios multi-cradles, also #1490 needs to be resolved.

.circleci/config.yml
```diff
@@ -30,6 +30,10 @@ defaults: &defaults
         name: Stack upgrade
         command: stack upgrade
 
+    - run:
+        name: Cabal update
+        command: cabal v2-update
```
alanz: For repeatability, we might consider something like

```
cabal v2-update 'hackage.haskell.org,2019-12-26T09:49:12Z'
```

fendor: this is just an experiment if this fixes the errors.
The tests are mainly executed with stack in mind, e.g. generated `hie.yaml` all use stack. So, actually, any cabal execution in this CI build is wrong, imo.
To fix this, we have to tackle #1431.

fendor: Found the problems and removed that line again. Only one small test-failure.

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -57,6 +58,64 @@ isStackCradle = (`elem` ["stack", "Cabal-Helper-Stack", "Cabal-Helper-Stack-None
...
+        Nothing -> return Nothing
+    else do
+      logm "Neither cabal nor stack project, look for ghc project."
+      findExecutable "ghc"
+  logm $ "Found ghc path: " ++ show ghcpath
+  return ghcpath
+
+tryCommand :: String -> IO String
+tryCommand cmd =
+  T.unpack . T.strip .T.pack <$> readCreateProcess (shell cmd) ""
```
fendor: I will change this to something like `String -> IO (ExitCode, String, String)` for easier logs. Any objections?

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -57,6 +58,64 @@ isStackCradle = (`elem` ["stack", "Cabal-Helper-Stack", "Cabal-Helper-Stack-None
...
+    ["cabal"
+    , "Cabal-Helper-Cabal-V1"
+    , "Cabal-Helper-Cabal-V2"
+    , "Cabal-Helper-Cabal-V1-Dir"
+    , "Cabal-Helper-Cabal-V2-Dir"
+    , "Cabal-Helper-Cabal-None"
+    ]
+  )
+  . BIOS.actionName
+  . BIOS.cradleOptsProg
```
fendor: TODO: upstream to hie-bios

src/Haskell/Ide/Engine/Server.hs
```diff
@@ -152,10 +152,23 @@ run scheduler _origDir plugins captureFp = flip E.catches handlers $ do
...
+        currentDir <- liftIO getCurrentDirectory
+
+        -- Check for mismatching GHC versions
+        -- Ignore hie.yaml parse errors. They get reported in ModuleCache.hs
+        let parseErrorHandler (_ :: Yaml.ParseException) = do
+              logm "Caught a yaml parse exception"
+              return Nothing
+            dummyCradleFile = fromMaybe currentDir lspRootDir </> "File.hs"
+        logm $ "Dummy Cradle File: " ++ dummyCradleFile
+        mcradle <- liftIO $ E.catch (Just <$> findLocalCradle dummyCradleFile) parseErrorHandler
```
fendor: Should the check in `NotInitialized` be moved here or can we pass the found cradle somehow to `onStartup`? Or the other way around?

src/Haskell/Ide/Engine/Server.hs
```diff
@@ -424,7 +436,9 @@ reactor inp diagIn = do
                   reactorSend $ NotShowMessage $ fmServerShowMessageNotification J.MtWarning cabalMsg
                   reactorSend $ NotLogMessage $ fmServerLogMessageNotification J.MtWarning cabalMsg
 
-            Nothing -> return ()
+            Left (_ :: Yaml.ParseException) -> do
+              logm "Failed to parse it"
+              reactorSend $ NotShowMessage $ fmServerShowMessageNotification J.MtError "Couldn't parse hie.yaml"
```
fendor: @jneira I could not make your test happy in any other way. 
Would that be alright to already show this error in the startup? I would actually say, yes.

test/functional/HieBiosSpec.hs
```diff
@@ -9,6 +9,7 @@ import Language.Haskell.LSP.Messages
 import System.FilePath ((</>))
 import Test.Hspec
 import TestUtils
+import           Debug.Trace                    ( trace )
```
fendor: remove

test/functional/HieBiosSpec.hs
```diff
@@ -20,15 +21,14 @@ spec = beforeAll_ (writeFile (hieBiosErrorPath </> "hie.yaml") "") $ do
...
+
   where hieBiosErrorPath = "test/testdata/hieBiosError"
-  
+
         isMessage (NotShowMessage (NotificationMessage _ _ (ShowMessageParams MtError s))) =
           "Couldn't parse hie.yaml" `T.isInfixOf` s
-        isMessage _ = False
-        
\ No newline at end of file
+        isMessage m = trace ("Message: " ++ show m) $ False
```
fendor: remove

src/Haskell/Ide/Engine/Server.hs
bubba: We should keep the check in NotInitialized as it's really the only appropriate time to send the `showMessage` request to the user. We probably don't need to log anything here then, just swallow the exception. 
As a side note, should `findCradle` in hie-bios not do this exception handling for us, returning a `Maybe` instead?

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
alanz: Sounds ok to me

src/Haskell/Ide/Engine/Server.hs
alanz: The best time to send the messages is in `onStartup`. `NotInitialized` is just the final handshake from the client that it is done.  `onStartup` is our internal processor of the `initialize` request from the client.

fendor: I would prefer it in `onStartup` also for DRY reasons. And if it is like @alanz says, I think it is just as valid and reasonable.

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -57,6 +58,98 @@ isStackCradle = (`elem` ["stack", "Cabal-Helper-Stack", "Cabal-Helper-Stack-None
...
+-- failed, a @Nothing@ is returned.
+-- The exact error is written to logs.
+--
+-- E.g. for a stack cradle, we use `stack ghc` and for a cabal cradle
+-- we are taking the @ghc@ that is on the path.
+execProjectGhc :: Cradle -> [String] -> IO (Maybe String)
+execProjectGhc crdl args = do
+  isStackInstalled <- isJust <$> findExecutable "stack"
+  -- isCabalInstalled <- isJust <$> findExecutable "cabal"
+  ghcpath <- if isStackCradle crdl && isStackInstalled
```
fendor: old name

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -57,6 +58,98 @@ isStackCradle = (`elem` ["stack", "Cabal-Helper-Stack", "Cabal-Helper-Stack-None
...
+
+
+-- | Get the directory of the libdir based on the project ghc.
+getProjectGhcLibDir :: Cradle -> IO (Maybe FilePath)
+getProjectGhcLibDir crdl =
+  catch
+    (execProjectGhc crdl ["--print-libdir"])
+    $ \(_ :: IOException) -> do
+      logm "Could not obtain the libdir."
+      return Nothing
```
fendor: should no longer throw an exception.

src/Haskell/Ide/Engine/Scheduler.hs
```diff
@@ -143,8 +145,9 @@ runScheduler
      -- ^ A handler to run the requests' callback in your monad of choosing.
   -> Maybe (Core.LspFuncs Config)
      -- ^ The LspFuncs provided by haskell-lsp, if using LSP.
+  -> Maybe Bios.Cradle
```
fendor: Needs a comment

src/Haskell/Ide/Engine/Version.hs
```diff
@@ -39,55 +35,18 @@ hieVersion =
...
-        L.infoM "hie" "Couldn't find stack version, falling back to plain GHC"
-        getGhcVersion
-    else do
-      L.infoM "hie" "Using plain GHC version"
-      getGhcVersion
+getProjectGhcVersion :: Bios.Cradle -> IO String
+getProjectGhcVersion crdl =
+  execProjectGhc crdl ["--numeric-version"] >>= \case
+    Just version -> return version
+    Nothing -> return "No System GHC Found."
```
fendor: `fromMaybe "No System GHC Found." <$> execProjectGhc crdl ["--numeric-version"]`

fendor:
@bubba thanks for fixing the tests! 
PR is ready for merge, waiting for some reviews now.

src/Haskell/Ide/Engine/Server.hs
```diff
@@ -151,12 +152,64 @@ run scheduler _origDir plugins captureFp = flip E.catches handlers $ do
...
+              `E.catch` \(e :: E.SomeException) ->
+              (errorm $ "Reactor thread exited unexpectedly: " ++ show e)
+            ))
+        flip labelThread "diagnostics" =<<
+            (forkIO (
+              diagnosticsQueue tr
+              `E.catch` \(e :: E.SomeException) ->
+              (errorm $ "Diagnostic thread exited unexpectedly: " ++ show e)
+            ))
+
```
alanz: At some future date we should go back to launching these threads with `race` rather than `forkIO`, so if one does die, we know immediately, and can see something in the log.  With this setup it can still limp along.

PR #1521, opened 2019-12-29
Replaced cabal-helper submodule with hackage version

alanz:


bubba:
Then we can finally delete the submodules folder. And keep it that way (hopefully)

alanz:
@bubba , except the pending GHC 8.8 support branch brings in apply-refact as a submodule :(

alanz:
Now if someone feels brave we can consider hackage.

PR #1522, opened 2019-12-29
Improve quality and information density of error message

### LINKED ISSUE #1509: Give a reason why we could not obtain build flags
Users currenty see the message `Could not obtain flags for <...>` quite often, but it actually gives no indicator why it happens. We can improve that!

fendor:
Fixes #1509
Improves amount of information the error message gives.
Also aims to give: possible solutions, and teaching material.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -581,7 +581,7 @@ getComponent env unitCandidates fp = getComponent' [] [] unitCandidates >>=
                 ++ concatMap ppShowUnitInfo triedUnits
                 ++ [ ""
                    , ""
-                   , "If you dont know how to expose a module, take a look at:"
+                   , "To expose a module, refer to:"
                    , "https://www.haskell.org/cabal/users-guide/developing-packages.html"
                    , ""
                    ]
@@ -828,4 +828,3 @@ cradleDisplay cradle = fromString result
       | "multi" `isInfixOf` name = "Multi Component project"
       | otherwise = "project"
     name = map toLower $ BIOS.actionName (BIOS.cradleOptsProg cradle)
-
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -13,14 +13,14 @@ import           Distribution.Helper (Package, projectPackages, pUnits,
                                       unChModuleName, Ex(..), ProjLoc(..),
                                       QueryEnv, mkQueryEnv, runQuery,
                                       Unit, unitInfo, uiComponents,
-                                      ChEntrypoint(..), uComponentName)
+                                      ChEntrypoint(..), UnitInfo(..))
 import           Distribution.Helper.Discover (findProjects, getDefaultDistDir)
 import           Data.Char (toLower)
 import           Data.Function ((&))
-import           Data.List (isPrefixOf, isInfixOf, sortOn, find, intercalate)
+import           Data.List (isPrefixOf, isInfixOf, sortOn, find)
 import qualified Data.List.NonEmpty as NonEmpty
 import           Data.List.NonEmpty (NonEmpty)
-import qualified Data.Map as M
+import qualified Data.Map as Map
 import           Data.Maybe (listToMaybe, mapMaybe, isJust)
 import           Data.Ord (Down(..))
 import           Data.String (IsString(..))
@@ -146,7 +146,7 @@ getProjectGhcLibDir :: Cradle -> IO (Maybe FilePath)
 getProjectGhcLibDir crdl =
   execProjectGhc crdl ["--print-libdir"] >>= \case
     Nothing -> do
-      logm "Could not obtain the libdir."
+      errorm "Could not obtain the libdir."
       return Nothing
     mlibdir -> return mlibdir
 
@@ -548,7 +548,7 @@ getComponent env unitCandidates fp = getComponent' [] [] unitCandidates >>=
       (tried, failed, Nothing) -> return (Left $ buildErrorMsg tried failed)
       (_, _, Just comp) -> return (Right comp)
   where
-    getComponent' :: [Unit pt] -> [Unit pt] -> [Unit pt] -> IO ([Unit pt], [Unit pt], Maybe ChComponentInfo)
+    getComponent' :: [UnitInfo] -> [(Unit pt, IOException)] -> [Unit pt] -> IO ([UnitInfo], [(Unit pt, IOException)], Maybe ChComponentInfo)
     getComponent' triedUnits failedUnits [] = return (triedUnits, failedUnits, Nothing)
     getComponent' triedUnits failedUnits (unit : units) =
       try (runQuery (unitInfo unit) env) >>= \case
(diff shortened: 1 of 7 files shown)


hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -156,34 +163,49 @@ loadCradle iniDynFlags (NewCradle fp) def action = do
...
+            sev = Just DsError
+            range = Range (Position 0 0) (Position 1 0)
+            msgTxt =
+              [ "Fail on initialisation for \"" <> Text.pack fp <> "\"."
+              ] <> map Text.pack msg
+            source = Just "bios"
+            diag = Diagnostic range sev Nothing source (Text.unlines msgTxt) Nothing
+
+        liftIO $ publishDiagnostics maxBound normalizedUri Nothing
+            (Map.singleton source (SL.singleton diag))
```
fendor: This is now the publishing of diagnostics. @alanz I dont understand how we intend to change the `maxToSend` flag.

alanz: it should not be part of `PublishDiagnostics`.   Rather make the type

```haskell
type PublishDiagnostics = J.NormalizedUri -> J.TextDocumentVersion -> J.DiagnosticsBySource -> IO ()
```

Then the one from `Server.hs` can operate like a callback, where it looks up that value before sending.  Which of course potentially brings in a race condition, but at least it is a read only.

It does leave the way open for the passed in function to be something like `passHandler`

```haskell
passHandler :: TChan ReactorInput -> (a -> FromClientMessage) -> Core.Handler a
passHandler rin c notification = do
  atomically $ writeTChan rin (c notification)
```
which does keep things separate, at the cost of having a different type for the `ReactorInput`.

fendor:
Example error message:
```
Fail on initialisation for "/home/baldr/Documents/haskell/fprog/src/Angabe6.hs".
Could not obtain flags for: "src/Angabe6.hs".

This Module was not part of any component we are aware of.

Component: ChLibName ChMainLibName with source directory: ["src"]
Component: ChExeName "intcode" with source directory: ["app"]


If you dont know how to expose a module, take a look at:
https://www.haskell.org/cabal/users-guide/developing-packages.html

We could not build all components.
If one of these components exposes this Module, make sure they compile.
You can try to invoke the commands yourself.
The following commands failed:

callProcessStderr: cabal v2-build --with-ghc=/home/baldr/.nix-profile/bin/ghc --with-ghc-pkg=/home/baldr/.nix-profile/bin/ghc-pkg --with-haddock=/home/baldr/.nix-profile/bin/haddock --project-file=/home/baldr/Documents/haskell/fprog/cabal.project --builddir=/home/baldr/Documents/haskell/fprog/dist-newstyle --only-configure fprog:exe:fprog (exit 1): failed

callProcessStderr: cabal v2-build --with-ghc=/home/baldr/.nix-profile/bin/ghc --with-ghc-pkg=/home/baldr/.nix-profile/bin/ghc-pkg --with-haddock=/home/baldr/.nix-profile/bin/haddock --project-file=/home/baldr/Documents/haskell/fprog/cabal.project --builddir=/home/baldr/Documents/haskell/fprog/dist-newstyle --only-configure fprog:test:fprog-test (exit 1): failed
bios
Peek Problem
No quick fixes available
```

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -534,25 +538,78 @@ cabalHelperCradle file = do
...
+            ++ concat
+              [
+                [ "This Module was not part of any component we are aware of."
+                , ""
+                ]
+                ++ concatMap ppShowUnitInfo triedUnits
+                ++ [ ""
+                   , ""
+                   , "If you dont know how to expose a module, take a look at:"
+                   , "https://www.haskell.org/cabal/users-guide/developing-packages.html"
```
jneira: Afaiu this will be called if we are using stack as well so maybe we should replace this link with a build tool agnostic one (or add an stack one)

bubba: Both use the underlying cabal file format, so this link should be fine right? Stack doesn't provide any of its own documentation for it does it?

src/Haskell/Ide/Engine/Server.hs
```diff
@@ -247,9 +256,13 @@ run scheduler _origDir plugins captureFp = flip E.catches handlers $ do
...
-type ReactorInput
-  = FromClientMessage
-      -- ^ injected into the reactor input by each of the individual callback handlers
+data ReactorInput
+  = CM FromClientMessage
+      -- ^ injected into the reactor input by each of the individual
+      -- callback handlers
+    | PD J.NormalizedUri J.TextDocumentVersion DiagnosticsBySource
+      -- ^ injected into the reactor input by any scheduler needing to
+      -- publish additional diagnostics
```
bubba: This seems kind of hairy. My original thought that was the "reacted" to messages coming from the client, but this also now handles messages going in both directions?

I agree that there's not really any nice solution to this, but seeing as returning a PD is just calling `liftIO Core.publishDiagnosticsFunc` at the end of the day albeit queued on the reactor thread, would it just be simpler to not bother with `passPublishDiagnostics`? I don't think there is any particular benefit to keeping it on the one thread

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -534,25 +538,78 @@ cabalHelperCradle file = do
...
+            ]
+            ++ concat
+              [
+                [ "This Module was not part of any component we are aware of."
+                , ""
+                ]
+                ++ concatMap ppShowUnitInfo triedUnits
+                ++ [ ""
+                   , ""
+                   , "If you dont know how to expose a module, take a look at:"
```
bubba: Minor nitpick: could this be reworded to

```suggestion
                   , "To expose a module, refer to:"
```

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -534,25 +538,78 @@ cabalHelperCradle file = do
...
+              | not (null triedUnits)
+              ]
+            ++ concat
+              [
+                [ "We could not build all components."
+                , "If one of these components exposes this Module, make sure they compile."
+                , "You can try to invoke the commands yourself."
+                , "The following commands failed:"
+                ]
+                ++ concatMap (ppShowIOException . snd) failedUnits
```
bubba: This is so much more user-friendly 👍

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
jneira: Well, it has its own documentation, covering `package.yaml` and `stack.yaml`. Thanks to some of the lastest issues i am discovering that there are quite users using that combo and not sure if all of them even cares about the `.cabal` file. Moreover all cli examples are with cabal and they are not applicable with stack.

I think the more similar stack doc link could be: https://docs.haskellstack.org/en/stable/GUIDE/

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
fendor: It is still horrible, but we discussed on irc how to improve cabal-helper to have good error messages!

src/Haskell/Ide/Engine/Server.hs
alanz: @bubba the problem is we need to honour the config of max diagnostics to send, which can be changed at any time.

And I would rather do it this way, than have another (global var) IORef in the mix.

bubba: I don’t think we would need an IORef, just pass a function over to it which is what it seems to be doing already. 
The max diagnostics can also be fetched via the LspFuncs, so I think it should be possible if we just passed that instead

alanz: Hmm, just took a look at `LspFuncs` again. It has grown quite a bit. And I agree, it has a way to access the config.

And as an aside, given the maxToSend is a config options, maybe we should just automatically use it in the `publishDiagnostics` LspFuncs.

fendor: Are the LspFuncs updates though in every thread? I could not tell what would update them over the different threads.

alanz: I think the architecture has mutated to a model where the LspFuncs are basically a lock on the key data structures.  So it should be fine.

I also have a feeling the architecture may change as a result of the ghcide/hie-core collaboration.

bubba:
This LGTM! Looking forward to having this in

PR #1523, opened 2019-12-29
Change Maybe LspFuncs to LspFuncs inside IdeEnv

bubba:
Now that its not possible to *not* have LspFuncs since the only transport is LSP


src/Haskell/Ide/Engine/Scheduler.hs
```diff
@@ -143,13 +143,13 @@ runScheduler
      -- ^ A handler for any errors that the dispatcher may encounter.
   -> CallbackHandler m
      -- ^ A handler to run the requests' callback in your monad of choosing.
-  -> Maybe (Core.LspFuncs Config)
+  -> Core.LspFuncs Config
      -- ^ The LspFuncs provided by haskell-lsp, if using LSP.
```
bubba:
```suggestion
     -- ^ The LspFuncs provided by haskell-lsp.
```

PR #1526, opened 2019-12-30
[WIP] Append modules to cradle cache if not already in module graph

bubba:
This can happen whenever a cradle for a module resolves to something outside of the module graph, i.e. across components. We should still cache the cradle for the module since they will use it.
This fixes the cradle repeatedly being initialized.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
@@ -130,11 +130,13 @@ loadCradle :: forall a m . (MonadIde m, HasGhcModuleCache m, GHC.GhcMonad m
 loadCradle _ _ ReuseCradle _def action = do
   -- Since we expect this message to show up often, only show in debug mode
   debugm "Reusing cradle"
+  sendTelemetry $ Aeson.String "loadCradle:ReuseCradle"
   IdeResultOk <$> action
 
 loadCradle _ _iniDynFlags (LoadCradle (CachedCradle crd env co)) _def action = do
   -- Reloading a cradle happens on component switch
   logm $ "Switch to cradle: " ++ show crd
+  sendTelemetry $ Aeson.String "loadCradle:LoadCradle"
   -- Cache the existing cradle
   maybe (return ()) cacheCradle =<< (currentCradle <$> getModuleCache)
   GHC.setSession env
@@ -144,6 +146,7 @@ loadCradle _ _iniDynFlags (LoadCradle (CachedCradle crd env co)) _def action = d
 loadCradle publishDiagnostics iniDynFlags (NewCradle fp) def action = do
   -- If this message shows up a lot in the logs, it is an indicator for a bug
   logm $ "New cradle: " ++ fp
+  sendTelemetry $ Aeson.String $ Text.pack $ "loadCradle:NewCradle:" <> fp
   -- Cache the existing cradle
   -- And append this module (fp) to the cradle if it isn't already in it
   -- This can happen if fp is outside of the module graph of the cradle, but should
--- a/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
@@ -148,22 +148,7 @@ loadCradle publishDiagnostics iniDynFlags (NewCradle fp) def action = do
   logm $ "New cradle: " ++ fp
   sendTelemetry $ Aeson.String $ Text.pack $ "loadCradle:NewCradle:" <> fp
   -- Cache the existing cradle
-  -- And append this module (fp) to the cradle if it isn't already in it
-  -- This can happen if fp is outside of the module graph of the cradle, but should
-  -- still use that cradle. e.g.
-  --
-  -- Components                         | Modules in component
-  -- -----------------------------------+----------------------
-  -- pkg:lib <-- cradle is set to this  | Lib.hs
-  -- pkg:exe                            | Exe.hs
-  --
-  -- Exe.hs might import across a component Lib.hs. So it *won't* appeari n the module graph
-  -- And so won't appear in the currentCradle. We need to manually append it otherwise the
-  -- cradle will never get cached for it.
-  let appendThisModule (fps, c)
-        | fp `notElem` fps = (fp:fps, c)
-        | otherwise = (fps, c)
-  maybe (return ()) cacheCradle =<< (fmap appendThisModule . currentCradle <$> getModuleCache)
+  maybe (return ()) cacheCradle =<< (currentCradle <$> getModuleCache)
 
   -- Now load the new cradle, accounting for hie.yaml parse errors
   let parseErrorHandler = return . Left . Yaml.prettyPrintParseException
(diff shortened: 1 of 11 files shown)


bubba:
Here is a test case for this problem. Beforehand editing `Main.hs` would always initialise the cradle

[comperrortest.zip](https://github.com/haskell/haskell-ide-engine/files/4008901/comperrortest.zip)

alanz:
@bubba do you intend actually adding the test?

bubba:
@alanz I'm trying to think of how to test for caching. Presumably this can't be a functional test, we might need to expose some of the stuff inside `ModuleCache.hs`?

alanz:
@bubba Perhaps we should consider emitting ['telemetry/event'](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#telemetry_event) notifications about it from hie, and add functionality to `lsp-test` to be able to make assertions about those?

bubba:
@alanz that’s a good idea. And we can add a flag/initialisation option to enable them in testing  so they don’t pollute regular users

alanz:
LGTM

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -451,6 +454,17 @@ withIndefiniteProgress t c f = do
   lf <- ideEnvLspFuncs <$> getIdeEnv
   control $ \run -> Core.withIndefiniteProgress lf t c (run f)
 
+-- | Sends a telemetry/event notification to the client, if the environment variable
+-- @HIE_TELEMETRY@ is set.
```
alanz: Do we document all these various environment variables we use to configure the tests anywhere?

It might be really good to dump them via a CLI flag, but perhaps we should consider some mechanism that gathers them all in one place, so we can be sure that we have a single list. Some kind of structure with a list of used environment variables, and an API to request it, which fails if the variable is not registered.

alanz:
Should there be a bump of the `lsp-test` dependency, or does it already handle telemetry messages?

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -136,8 +138,24 @@ loadCradle _iniDynFlags (LoadCradle (CachedCradle crd env)) _def action = do
...
+  -- And append this module (fp) to the cradle if it isn't already in it
+  -- This can happen if fp is outside of the module graph of the cradle, but should
+  -- still use that cradle. e.g.
+  --
+  -- Components                         | Modules in component
+  -- -----------------------------------+----------------------
+  -- pkg:lib <-- cradle is set to this  | Lib.hs
+  -- pkg:exe                            | Exe.hs
+  --
+  -- Exe.hs might import across a component Lib.hs. So it *won't* appeari n the module graph
```
fendor: small typo

fendor:
I see the following in the logs:
```haskell
modGraph:
    [ ModLocation
        { ml_hs_file = Just "Foo.hs"
        , ml_hi_file = "/home/baldr/.cache/hie-bios/f8a5d6e6f9184b34cc37124c0dfcd509f2e62335/Foo.hi"
        , ml_obj_file = "/home/baldr/Downloads/comperrortest/comperrortest/dist-newstyle/build/x86_64-linux/ghc-8.6.4/hieBiosComponentErrors-0.1.0.0/build/Foo.o"
        }
    , ModLocation
        { ml_hs_file = Just "/tmp/haskell-lsp10150/Main.hs-00002--5323200879228195963.hs"
        , ml_hi_file = "/home/baldr/.cache/hie-bios/f8a5d6e6f9184b34cc37124c0dfcd509f2e62335/Main.hi"
        , ml_obj_file = "/home/baldr/Downloads/comperrortest/comperrortest/dist-newstyle/build/x86_64-linux/ghc-8.6.4/hieBiosComponentErrors-0.1.0.0/build/Main.o"
        }
    , ModLocation
        { ml_hs_file = Just "/home/baldr/Downloads/comperrortest/comperrortest/./OtherMain.hs"
        , ml_hi_file = "/home/baldr/.cache/hie-bios/f8a5d6e6f9184b34cc37124c0dfcd509f2e62335/OtherMain.hi"
        , ml_obj_file = "/home/baldr/Downloads/comperrortest/comperrortest/dist-newstyle/build/x86_64-linux/ghc-8.6.4/hieBiosComponentErrors-0.1.0.0/build/OtherMain.o"
        }
    ]
```
This should actually give us all these files for the module graph.

bubba:
@fendor yeah I agree, this is definitely a plaster across the symptom. Thanks for finding that in the logs, I’ll try to use that

bubba:
@alanz there’s actually some really nefarious issue with how lsp-test parses messages that meant the `message :: Session TelemtetryNotification` actually parsed showmessage notifications too, hence the weird satisfy predicate. I’m kinda hoping @wz1000’s work on Haskell-lsp might fix this at some point? But for now the current lsp-test is working fine

fendor:
@bubba I will help, but we won't fix that properly before the release and I feel hesitant to just put a plaster on the issue. Does this happen with all executable components?

bubba:
@fendor Sure, I wasn't planning on getting this into the release anyway. it's a pretty niche problem.
After some digging anyway, here's what I found.

For the test case there are three modules:
```
Foo <- Main -> OtherMain
```

The issue was occurring when opening up Main.hs. I was wondering why in `setTargetFilesWithMessage` in Check.hs got the full module graph during typechecking, but `initSessionWithMessage` in Api.hs did *not* when called from `initialiseCradle` in ModuleCache.hs. Its module graph only contained the single module that was passed in.

I then realised that the main difference was that `setTargetFilesWithMessage` was setting the target to Main.hs, but `initSessionWithMessage` was only setting the target to Foo.hs. And thus the graph didn't include the whole story. I presume this is happening because the cradle is only going to be specifying the library as its component.

bubba:
As an aside -- setting the target in `initSessionWithMessage` to whatever module ModuleCache passed in works, overriding whatever the cradleAction had set in its ComponentOptions. But this probably isn't the right solution. Should the cradle try to pull in dependencies from multiple components that depend on the specified component?

bubba:
One possibility that I think might work quite well -- we could just let cabal decide what to load when getting the flags:

```patch
diff --git a/src/HIE/Bios/Cradle.hs b/src/HIE/Bios/Cradle.hs
index c2e2948..088b718 100644
--- a/src/HIE/Bios/Cradle.hs
+++ b/src/HIE/Bios/Cradle.hs
@@ -330,8 +330,8 @@ getCabalWrapperTool (ghcPath, ghcArgs) wdir = do
 cabalAction :: FilePath -> Maybe String -> LoggingFunction -> FilePath -> IO (CradleLoadResult ComponentOptions)
 cabalAction work_dir mc l _fp = do
   wrapper_fp <- getCabalWrapperTool ("ghc", []) work_dir
-  let cab_args = ["v2-repl", "--with-compiler", wrapper_fp]
-                  ++ [component_name | Just component_name <- [mc]]
+  let cab_args = ["v2-repl", "--with-compiler", wrapper_fp, _fp]
+                  -- ++ [component_name | Just component_name <- [mc]]
   (ex, output, stde, args) <-
     readProcessWithOutputFile l Nothing work_dir "cabal" cab_args
   deps <- cabalCradleDependencies work_dir
```

This causes the correct module/target to be loaded by just passing in the file path instead of whatever component hie.yaml specified, and letting cabal figure the rest out. It would make the component field in cabal redundant though.

mpickering:
@bubba Yes, but is this documented anywhere?

bubba:
@mpickering From the [docs](https://cabal.readthedocs.io/en/latest/nix-local-build.html#cabal-v2-build):
> (There is also syntax for specifying module and file targets, but it doesn’t currently do anything.)

 The [logic is definitely there though](https://github.com/haskell/cabal/blob/65522af9b1a3e7f3eb2c712f85bec38a01af3cc2/cabal-install/Distribution/Client/TargetSelector.hs#L364-L379), and seems to work for targets everywhere inc. `cabal build`
Maybe they added it and forgot to update the docs?

PR #1531, opened 2019-12-31
Look for stack.yaml before selecting .cabal location as project root

### LINKED ISSUE #1530: Implicit Project discovery discovers cabal-project although there is a stack.yaml
Reproduce, have cabal and stack installed:
```bash
git clone https://github.com/haskell/haskell-ide-engine
rm cabal.project # currently needed to force stack usage
code . # lsp client, and open Cradle.hs
```
We can see in the logs that `hie-plugin-api` is selected as the root of the project, but it should be `haskell-ide-engine`.

fendor:
Fixes #1530 
Behaviour changed because of https://github.com/DanielG/cabal-helper/pull/84
cc  @bubba

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -235,12 +235,12 @@ findCabalHelperEntryPoint fp = do
       isStackProject (Ex ProjLocStackYaml {}) = True
       isStackProject _ = False
 
-      isCabalNewProject (Ex ProjLocV2Dir {}) = True
       isCabalNewProject (Ex ProjLocV2File {}) = True
       isCabalNewProject _ = False
 
       isCabalOldProject (Ex ProjLocV1Dir {}) = True
       isCabalOldProject (Ex ProjLocV1CabalFile {}) = True
+      isCabalOldProject (Ex ProjLocV2Dir {}) = True
       isCabalOldProject _ = False
 
 {- | Given a FilePath, find the cradle the FilePath belongs to.
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -231,8 +231,8 @@ findCabalHelperEntryPoint fp = do
       supported :: Ex ProjLoc -> Bool -> Bool -> Bool
       supported proj stackInstalled cabalInstalled
         | isCabalProject proj = cabalInstalled
-        -- isStackProject proj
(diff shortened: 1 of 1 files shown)


bubba:
Just to clarify, a directory with just a .cabal file is a v2 project as of Cabal 3, since it defaults to v2-build, without the need for a cabal.project file

fendor:
I think I named them like that back then, at least git blame, blames me.

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -201,7 +201,10 @@ Just (Ex (ProjLocV2File { plProjectDirV2 = "/Foo/"}))
 
 In the given example, it is not guaranteed which project type is found,
 it is only guaranteed that it will not identify the project
-as a cabal v1-project.
+as a cabal v1-project. Note that with latest cabal-helper, by default a *.cabal file
```
alanz: Should we maybe give a specific version of cabal-helper here? Either by commit hash, or tag, whatever. Or even/also just put the date in brackets. These comments tend to hang around forever.

fendor: Fair, good idea.

hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
```diff
@@ -219,29 +222,29 @@ findCabalHelperEntryPoint fp = do
...
-      supported (Ex ProjLocStackYaml {}) stackInstalled _ = stackInstalled
-      supported (Ex ProjLocV2Dir {}) _ cabalInstalled = cabalInstalled
-      supported (Ex ProjLocV2File {}) _ cabalInstalled = cabalInstalled
-      supported (Ex ProjLocV1Dir {}) _ cabalInstalled = cabalInstalled
-      supported (Ex ProjLocV1CabalFile {}) _ cabalInstalled = cabalInstalled
+      supported :: Ex ProjLoc -> Bool -> Bool -> Bool
+      supported proj stackInstalled cabalInstalled
+        | isCabalProject proj = cabalInstalled
+        | isStackProject proj = stackInstalled
+        | otherwise = False
```
fendor: I thought being explicit is better than implicit. It may happen that c-h starts to support more projects then just cabal and stack and we dont want to misidentify them as stack projects.
Actually, is the old code better since this would turn such an event into a compile-time error?

alanz: There is something to be said for exhaustive matching, because the compiler can then warn about missing matches.

PR #1533, opened 2019-12-31
Add sample hie.yaml files for stack and cabal

alanz:
The cabal one is tested per component.

The stack one has issues because of the nested library.

Note: the suffix on the cabal one cannot be '.cabal', as that
extension is already used.

### DIFF
--- a/hie.yaml.cbl
+++ b/hie.yaml.cbl
@@ -1,0 +1,28 @@
+cradle:
+  cabal:
+    - path: "./test/dispatcher/"
+      component: "haskell-ide-engine:dispatcher-test"
+
+    - path: "./test/functional/"
+      component: "haskell-ide-engine:func-test"
+
+    - path: "./test/unit/"
+      component: "haskell-ide-engine:unit-test"
+
+    - path: "./test/plugin-dispatcher/"
+      component: "haskell-ide-engine:plugin-dispatcher-test"
+
+    - path: "./test/wrapper/"
+      component: "haskell-ide-engine:wrapper-test"
+
+    - path: "./hie-plugin-api/"
+      component: "lib:hie-plugin-api"
+
+    - path: "./app/MainHie.hs"
+      component: "haskell-ide-engine:hie:exe"
+
+    - path: "./app/HieWrapper.hs"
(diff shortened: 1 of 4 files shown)


hie.yaml.cbl
```diff
@@ -0,0 +1,28 @@
...
+      component: "lib:hie-plugin-api"
+
+    - path: "./app/MainHie.hs"
+      component: "haskell-ide-engine:hie:exe"
+
+    - path: "./app/HieWrapper.hs"
+      component: "haskell-ide-engine:hie-wrapper:exe"
+
+    - path: "./"
+      component: "lib:haskell-ide-engine"
```
fendor: utils is missing.

hie.yaml.stack
```diff
@@ -0,0 +1,31 @@
...
+      component: "haskell-ide-engine:test:unit-test"
+
+    - path: "./test/plugin-dispatcher/"
+      component: "haskell-ide-engine:test:plugin-dispatcher-test"
+
+    - path: "./test/wrapper/"
+      component: "haskell-ide-engine:test:wrapper-test"
+
+    - path: "./test/utils/"
+      component: "haskell-ide-engine:lib:hie-test-utils"
```
fendor: Does this work? Did you try this?

fendor: Because when I do `stack repl haskell-ide-engine:lib:hie-test-utils` I get an error.

hie.yaml.cbl
alanz: oops, need to copy it over.

hie.yaml.stack
alanz: No, that does not work, because of current lack of sublibrary support in stack. Which is why I switched to the cabal cradle.  But I put it in for completeness, in the hope that it will someday work.

fendor: I made it work by using `--no-package-hiding`, assuming it does what we want, we may want to patch hie-bios.

alanz: @fendor can you capture this `--no-package-hiding` possibility in an issue?

PR #1540, opened 2020-01-02
[WIP] Download hlint data-files if they cant be found.

### LINKED ISSUE #1143: Hlint data-files can not be easily distributed

fendor:
Closes #1143
Changes the behaviour in the following way:
* Introduces a `initializeHlint` function
* Introduce `HIE_HLINT_DATADIR` env variable for the location of the data-files.
* Download missing data-files with `stack` or `cabal`, whatever is available.
* Copy data-files into `XDG_DATA_HOME/hie/hlint`
* Write the data-dir location into the extensible state.

Missing stuff:
* [ ] After downloading stuff, set the data-dir location in the extensible state.
* [ ] Use a https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#window_showMessageRequest to ask if we should download stuff.
* [ ] Add a test to prove the code does what we want.
* [ ] Update `install.hs` to do this stuff for us, cc @jneira. 
* [ ] Check if at least one file is in the data-dir.

To check it out, open a project wait for typecheck, then reopen the project again. Now no data-files should be downloaded and hlint should produce diagnostics. On the first try, it definitely does not produce diagnostics, since I didnt implement `After downloading stuff, set the data-dir location in the extensible state.` yet.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
+++ b/src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
@@ -1,6 +1,9 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveAnyClass        #-}
 {-# LANGUAGE DeriveGeneric         #-}
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE OverloadedStrings     #-}
 -- | apply-refact applies refactorings specified by the refact package. It is
 -- currently integrated into hlint to enable the automatic application of
@@ -17,6 +20,7 @@ import           Control.Exception              ( IOException
 import           Control.Lens            hiding ( List )
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Except
+import           Control.Monad                  ( filterM, forM_ )
 import           Data.Aeson                        hiding (Error)
 import           Data.Maybe
 import           Data.Monoid                       ((<>))
@@ -32,6 +36,18 @@ import           Language.Haskell.HLint4           as Hlint
 import qualified Language.Haskell.LSP.Types        as LSP
 import qualified Language.Haskell.LSP.Types.Lens   as LSP
 import           Refact.Apply
+import           System.Environment             ( lookupEnv )
+import           System.Directory               ( doesFileExist
+                                                , getXdgDirectory
+                                                , XdgDirectory(..)
+                                                , findExecutable
+                                                , listDirectory
+                                                , copyFile
+                                                )
+import           System.FilePath                ( (</>) )
+import           System.Process
+import           System.IO.Temp                 ( withSystemTempDirectory )
+import           System.Exit                    ( ExitCode(..) )
 
 -- ---------------------------------------------------------------------
 {-# ANN module ("HLint: ignore Eta reduce"         :: String) #-}
@@ -105,6 +121,103 @@ applyAllCmd uri = pluginGetFile "applyAll: " uri $ \fp -> do
       Right fs -> return (IdeResultOk fs)
 
 -- ---------------------------------------------------------------------
+
+newtype HlintDataDir = HlintDataDir
+  { getHlintDataDir :: Maybe FilePath
+  }
+  deriving (Show, Eq, Read, Ord)
+
+
+instance ExtensionClass HlintDataDir where
+  initialValue = HlintDataDir Nothing
+
+data InitialisationError
+  = NoDataFiles FilePath
+  | InvalidDataDir FilePath
+  | Other
+
+ppInitialisationError :: InitialisationError -> String
+ppInitialisationError (NoDataFiles fp) =
+  "We could not find the data-files at: \"" <> fp <> "\""
+ppInitialisationError (InvalidDataDir fp) =
+  "The location \""
+    <> fp
+    <> "\" does not exist. We obtained that location from the environment variable: "
+    <> hlintDataDirEnvironmentVariable
+ppInitialisationError Other = "We could not determine why it failed."
+
+data Tool = Stack | Cabal
+
+showTool :: Tool -> String
+showTool Stack = "stack"
+showTool Cabal = "cabal"
+
+hlintName :: String
+hlintName = "hlint-" ++ VERSION_hlint
+
+hlintDataDirDefaultLocation :: MonadIO m => m FilePath
+hlintDataDirDefaultLocation = liftIO $ getXdgDirectory XdgData ("hie" </> "hlint")
+
+hlintDataDirEnvironmentVariable :: String
+hlintDataDirEnvironmentVariable = "HIE_HLINT_DATADIR"
+
+initializeHlint :: IdeGhcM (Either InitialisationError FilePath)
+initializeHlint = do
+  explicitDataDirLocation <- liftIO $ lookupEnv hlintDataDirEnvironmentVariable
+  hieHlintDataDir <- hlintDataDirDefaultLocation
+  dataDirExists <- liftIO (doesFileExist hieHlintDataDir)
+  result <- case explicitDataDirLocation of
+    Just explicitDir -> do
+      liftIO (doesFileExist explicitDir)
+        >>= \case
+          True ->
+            return $ Right explicitDir
+          False ->
+            return $ Left $ InvalidDataDir explicitDir
+    Nothing ->
+      if dataDirExists
+        then return $ Right hieHlintDataDir
+        else return $ Left $ NoDataFiles hieHlintDataDir
+
+  case result of
+    Right dataDir -> put $ HlintDataDir $ Just dataDir
+    _ -> return ()
+
+  return result
+
+downloadHlintDatafiles :: IO ()
+downloadHlintDatafiles = do
+  defaultLocation <- hlintDataDirDefaultLocation
+  availTools      <- liftIO
+    $ filterM (fmap isJust . findExecutable . showTool) [Stack, Cabal]
+  case availTools of
+    []       -> return ()
+    tool : _ -> withSystemTempDirectory "hie-hlint-datafiles" $ \tmpDir -> do
+      let toolArgs t = case t of
+            Cabal -> ["get", hlintName]
+            Stack -> ["unpack", hlintName]
+
+      let process =
+            (proc (showTool tool) (toolArgs tool)) { cwd = Just tmpDir }
+
+      (exitCode, _sout, _serr) <- liftIO
+        $ readCreateProcessWithExitCode process ""
+
+      case exitCode of
+        ExitFailure _ ->
+          error
+            $  "Command `"
+            ++ unwords (showTool tool : toolArgs tool)
+            ++ "` failed."
+        ExitSuccess -> do
+          let dirPath = tmpDir </> hlintName </> "data"
+          dataFiles <- liftIO $ listDirectory dirPath
+          forM_ dataFiles
+            $ \dataFile -> do
+              let absoluteFile = dirPath </> dataFile
+              logm $ "Copy File \"" ++ absoluteFile ++ "\" to \"" ++ defaultLocation ++ "\""
+              liftIO $ copyFile absoluteFile defaultLocation
+
 
 -- AZ:TODO: Why is this in IdeGhcM?
 lint :: Uri -> IdeGhcM (IdeResult PublishDiagnosticsParams)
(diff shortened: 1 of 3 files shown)


src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -106,6 +122,103 @@ applyAllCmd uri = pluginGetFile "applyAll: " uri $ \fp -> do
...
+downloadHlintDatafiles = do
+  defaultLocation <- hlintDataDirDefaultLocation
+  availTools      <- liftIO
+    $ filterM (fmap isJust . findExecutable . showTool) [Stack, Cabal]
+  case availTools of
+    []       -> return ()
+    tool : _ -> withSystemTempDirectory "hie-hlint-datafiles" $ \tmpDir -> do
+      let toolArgs t = case t of
+            Cabal -> ["get", hlintName]
+            Stack -> ["unpack", hlintName]
```
fendor: What if the cabal is not updates to know about the hlint version?
What if the hlint version is not on stackage? (very unlikely, just feels like not good software)

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -270,9 +384,9 @@ applyHint fp mhint fileMap = do
         throwE (show err)
 
 -- | Gets HLint ideas for
-getIdeas :: MonadIO m => FilePath -> Maybe OneHint -> ExceptT String m [Idea]
-getIdeas lintFile mhint = do
-  let hOpts = hlintOpts lintFile (oneHintPos <$> mhint)
+getIdeas :: MonadIO m => HlintDataDir -> FilePath -> Maybe OneHint -> ExceptT String m [Idea]
+getIdeas dataDir lintFile mhint = do
+  let hOpts = hlintOpts dataDir lintFile (oneHintPos <$> mhint)
```
fendor: Do more locations need to be updates?

src/Haskell/Ide/Engine/Server.hs
```diff
@@ -457,7 +457,29 @@ reactor inp diagIn = do
...
+
+              hlintCallback :: Either ApplyRefact.InitialisationError FilePath -> R ()
+              hlintCallback (Left err) = flip runReaderT renv $ do
+                reactorSend
+                  $ NotShowMessage
+                  $ fmServerShowMessageNotification J.MtWarning
+                  $ "Required HLint files are missing: " <> T.pack (ApplyRefact.ppInitialisationError err)
+                logm "Downloading hlint data-files now."
+                liftIO ApplyRefact.downloadHlintDatafiles
+                logm "Finished downloading."
```
fendor: Should probably be messages to the user.

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -106,6 +122,103 @@ applyAllCmd uri = pluginGetFile "applyAll: " uri $ \fp -> do
...
+
+showTool :: Tool -> String
+showTool Stack = "stack"
+showTool Cabal = "cabal"
+
+hlintName :: String
+hlintName = "hlint-" ++ VERSION_hlint
+
+hlintDataDirDefaultLocation :: MonadIO m => m FilePath
+hlintDataDirDefaultLocation = liftIO $ getXdgDirectory XdgData ("hie" </> "hlint")
```
fendor: I feel like we have to duplicate this code in `install.hs` and especially to `VERSION_hlint` we have no access. Any ideas @jneira?

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -106,6 +122,103 @@ applyAllCmd uri = pluginGetFile "applyAll: " uri $ \fp -> do
...
+initializeHlint = do
+  explicitDataDirLocation <- liftIO $ lookupEnv hlintDataDirEnvironmentVariable
+  hieHlintDataDir <- hlintDataDirDefaultLocation
+  dataDirExists <- liftIO (doesFileExist hieHlintDataDir)
+  result <- case explicitDataDirLocation of
+    Just explicitDir -> do
+      liftIO (doesFileExist explicitDir)
+        >>= \case
+          True ->
+            return $ Right explicitDir
```
fendor: Add a check.

alanz:
I am not convinced that this is the right way to do this.

Can we not copy the config file at compile time?  Even if we write some sort of helper exe.

And if we do end up doing this, I do not believe the downloading code belongs in the ApplyRefact plugin.

Perhaps we should reach out to @ndmitchell for some ideas around this.  I am sure he has a way of installing hlint for CI via a simple download. Rather tap into existing infrastructure.

fendor:
We discussed this issue in https://github.com/ndmitchell/hlint/issues/620. I took away from it, that it is probably the best to use `cabal get` or `stack unpack` to get the data-files.
We can do it only at compile-time and handle it like we do with `hoogle`. 
Downloading them at run-time has the advantage that it is viable for users to just download the executable and it works right away. But if you think this is not worth the additional overhead, we can do it only at compile-time.

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
jneira: Mmm and what about download the data file directly from https://raw.githubusercontent.com/ndmitchell/hlint/master/data/hlint.yaml?
Are the rest of files inside data needed for the hie usage?
Hardcoding the url is not ideal, but it uses fewer moving parts

alanz:
The [hlint install CI script](https://github.com/ndmitchell/hlint#running-with-continuous-integration) does

```bash
$ curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/run.sh
#!/bin/sh
# This script is invoked from my Travis commands
# It bootstraps to grab the a binary release and run it
set -e # exit on errors

PACKAGE=$1
if [ -z "$PACKAGE" ]; then
    echo No arguments provided, please pass the project name as the first argument
    exit 1
fi
shift

case "$(uname)" in
    "Darwin")
        OS=osx;;
    MINGW64_NT-*|MSYS_NT-*)
        OS=windows;;
    *)
        OS=linux
esac

if [ "$OS" = "windows" ]; then
    EXT=.zip
    ESCEXT=\.zip
else
    EXT=.tar.gz
    ESCEXT=\.tar\.gz
fi

echo Downloading and running $PACKAGE...
# Don't go for the API since it hits the Appveyor GitHub API limit and fails
RELEASES=$(curl --silent --show-error https://github.com/ndmitchell/$PACKAGE/releases)
URL=https://github.com/$(echo $RELEASES | grep -o '\"[^\"]*-x86_64-'$OS$ESCEXT'\"' | sed s/\"//g | head -n1)
VERSION=$(echo $URL | sed -n 's@.*-\(.*\)-x86_64-'$OS$ESCEXT'@\1@p')
TEMP=$(mktemp -d .$PACKAGE-XXXXXX)

cleanup(){
    rm -r $TEMP
}
trap cleanup EXIT

retry(){
    ($@) && return
    sleep 15
    ($@) && return
    sleep 15
    $@
}

retry curl --progress-bar --location -o$TEMP/$PACKAGE$EXT $URL
if [ "$OS" = "windows" ]; then
    7z x -y $TEMP/$PACKAGE$EXT -o$TEMP -r > /dev/null
else
    tar -xzf $TEMP/$PACKAGE$EXT -C$TEMP
fi
$TEMP/$PACKAGE-$VERSION/$PACKAGE $*
```

Downloading https://github.com/ndmitchell/hlint/releases/download/v2.2.5/hlint-2.2.5-x86_64-linux.tar.gz (from the hlint releases page on github) shows it has the exe and data dir.

Can't we do this at compile time?

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
alanz: Yes, that is also an option. Just making sure that we get the one as per the version of hlint we are using in the build.  Probably the best option.

fendor:
We can do all the stuff at compile-time (not sure what the script does, but we can solve the problem at compile-time). @jneira and I just thought that downloading the files at run-time would be more user-friendly.

alanz:
@fendor the fewer moving parts the better. I would rather we install it something like the hlint CI images, where you get a platform-specific, ready-to-roll executable, with all its data files.

Otherwise we end up re-implementing a download manager, dealing with security issues, proxies, corporate firewalls, failed downloads, version mismatches, partial downloads and corrupt files, etc.

src/Haskell/Ide/Engine/Server.hs
jneira: Yeah, progress status messages in the status bar would be great for that

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
jneira: we are using the same version of hlint for all ghc versions, right? It is hardcoded in all `stack-*.yaml`, so maybe it is not a big deal hardcoding it in the script? Not very elegant, though.

Mmm maybe overriding the environment variable `hlint_datadir` defined by cabal in `Paths_hlint.hs` before installing hie, so hlint data files will go in that location (not sure it works that way)

jneira:
> @fendor the fewer moving parts the better. I would rather we install it something like the hlint CI images, where you get a platform-specific, ready-to-roll executable, with all its data files.

A zip/tar self-contained with all the data files needed woud be great too.

> Otherwise we end up re-implementing a download manager, dealing with security issues, proxies, corporate firewalls, failed downloads, version mismatches, partial downloads and corrupt files, etc.

Agree it is better to avoid it if possible. To be fair the pr tried to download the data files indirectly using stack or cabal (that have more chances to work if you are already using them)

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
```diff
@@ -106,6 +122,103 @@ applyAllCmd uri = pluginGetFile "applyAll: " uri $ \fp -> do
...
+showTool Cabal = "cabal"
+
+hlintName :: String
+hlintName = "hlint-" ++ VERSION_hlint
+
+hlintDataDirDefaultLocation :: MonadIO m => m FilePath
+hlintDataDirDefaultLocation = liftIO $ getXdgDirectory XdgData ("hie" </> "hlint")
+
+hlintDataDirEnvironmentVariable :: String
+hlintDataDirEnvironmentVariable = "HIE_HLINT_DATADIR"
```
bubba: [Cabal already provides an environment variable for package data directories](https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=paths#accessing-data-files-from-package-code), and from whenever I last tested this it seemed to work with dependencies. Is there any possibility we could take advantage of `HASKELL_IDE_ENGINE_DATADIR` or `HLINT_DATADIR`?

jneira:
Reading the hlint code i've found where it is using the executable path to looking for the data file:

https://github.com/ndmitchell/hlint/blob/master/src/CmdLine.hs#L53-L60

It is in the path called by the hlint executable and no the lib (we are using it as a lib, right?) so we would need set `hlint_datadir` with the value of `getExecutablePath ++ "/hlint/data"`. I dont fully like use env vars but it would be the quick one.
I think we still should show an error with a suggesting about how to fix it manually in case the data files would not be there.

src/Haskell/Ide/Engine/Plugin/ApplyRefact.hs
fendor: Ok, I thought that these were not honoured in dependencies.

fendor:
Closing in favour of a different solution.

ndmitchell:
I received a PR to embed the files via Template Haskell, which I'm planning to accept after changes. With that done, you won't have to worry again: https://github.com/ndmitchell/hlint/issues/620

PR #1542, opened 2020-01-04
Add cmd script to install with cabal

jneira:
Sorry for restore an os specific script but i've got tired of typing or copypasting it

### DIFF
--- a/cabal-hie-install
+++ b/cabal-hie-install
@@ -1,1 +1,2 @@
+#!/bin/sh
 cabal v2-run ./install.hs --project-file install/shake.project $@
--- a/cabal-hie-install
+++ b/cabal-hie-install
@@ -1,0 +1,1 @@
+cabal v2-run ./install.hs --project-file install/shake.project $@
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ cabal v2-run ./install.hs --project-file install/shake.project <target>
 or using the existing alias script
 
 ```bash
-cabal-hie-install <target>
+./cabal-hie-install <target>
 ```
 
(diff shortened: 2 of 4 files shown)


fendor:
Well, I am fine with a single line script.

alanz:
I am happy with this too. Should we just put it there, or should there be something about it in the README?

When you invoke it, do you type `cabal-install`, or `cabal-install.cmd`? I am worried about the name, with respect to the package `cabal-install` too.

And lastly,

jneira:
A comment in the readme could be useful, i'll do
You can invoke it without ext: `cabal-install hie-8.6.5` f.e.

fendor:
That is a bit ambiguous, imo. Maybe a name like `cabal-hie-install` or something, making clear this is a hie specific installation script?

cabal-hie-install
```diff
@@ -0,0 +1 @@
+cabal v2-run ./install.hs --project-file install/shake.project $@
```
alanz: This should have `#!/bin/sh` on top, surely? and must be `chmod +x`'d too, if not already done.

jneira: I already set the executable bit. I could execute as is in my msys bash but i'll add the sehbang just in case.

power-fungus:
I guess we should poke the cabal team to allow setting the cabal project file inside the script. Hvr has hinted that this should be no problem to support but nobody has done that yet

jneira:
> I guess we should poke the cabal team to allow setting the cabal project file inside the script. Hvr has hinted that this should be no problem to support but nobody has done that yet

[The issue i opened](https://github.com/haskell/cabal/issues/6353) has not much feedback 🤔

PR #1545, opened 2020-01-05
Enable azure releases and some fixes

jneira:
* It makes release for all jobs (total 24), example: https://dev.azure.com/jneira/haskell-ide-engine/_build/results?buildId=371&view=artifacts&type=publishedArtifacts
* It fixes the z3 installation for macos
* It adds the hoogle databse generation, needed for tests
* It fixes the cache for windows, make it like the unix and macos ones
  * I've open an issue upstream: https://github.com/microsoft/azure-pipelines-tasks/issues/12070
* I've make some advancesin the task that uses cabal to run the install.hs script but it still fails trying to install hie
* It adds specific triggers

The history is a liitle bit hairy, maybe it should be squashed.

### DIFF
--- a/.azure/linux-stack.yml
+++ b/.azure/linux-stack.yml
@@ -34,9 +34,9 @@ jobs:
     displayName: "Download cache"
   - bash: |
       mkdir -p $STACK_ROOT
-      tar -xzf .azure-cache/stack-root.tar.gz -C /
+      tar -vxzf .azure-cache/stack-root.tar.gz -C /
       mkdir -p .stack-work
-      tar -xzf .azure-cache/stack-work.tar.gz
+      tar -vxzf .azure-cache/stack-work.tar.gz
     displayName: "Unpack cache"
     condition: eq(variables.CACHE_RESTORED, 'true')
   - bash: |
@@ -45,7 +45,7 @@ jobs:
   - bash: |
       mkdir -p ~/.local/bin
       curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | \
-        tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
+        tar vxz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
     displayName: Install stack
   - bash: |
       source .azure/linux.bashrc
@@ -74,7 +74,7 @@ jobs:
         GHC_MINOR_VERSION=nightly
       fi
       ARTIFACT_NAME=hie-$(hie --numeric-version)-ghc-$GHC_MINOR_VERSION-linux-x86_64
-      tar -czf $(Build.ArtifactStagingDirectory)/$ARTIFACT_NAME.tar.xz *
+      tar -vczf $(Build.ArtifactStagingDirectory)/$ARTIFACT_NAME.tar.xz *
     displayName: Install `hie` 
   - bash: |
       source .azure/linux.bashrc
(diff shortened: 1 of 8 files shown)


alanz:
I am thinking of starting a release process, should I wait for this first?

fendor:
I dont think so, the artifacts dont work at the moment, iirc, since we havent solved the problem with hlint yet.
Moreover, azure is still not enaled for our repository, right?

.azure/linux-stack.yml
```diff
@@ -73,10 +89,20 @@ jobs:
       source .azure/linux.bashrc
       stack install --resolver=lts-11.18 liquid-fixpoint-0.7.0.7 dotgen-0.4.2 fgl-visualize-0.1.0.1 located-base-0.1.1.1 liquidhaskell-0.8.2.4
     displayName: "Install Runtime Test-Dependencies: liquidhaskell"
+  - bash: |
+      source .azure/linux.bashrc
+      stack install hoogle --stack-yaml=$(YAML_FILE)
```
fendor: `stack build hoogle` should be enough, iirc.

.azure/windows-installhs-cabal.yml
```diff
@@ -17,21 +18,21 @@ jobs:
     displayName: Install GHC
   - bash: |
       source .azure/windows.bashrc
-      stack install cabal-install --stack-yaml $(YAML_FILE)
+      stack install Cabal-3.0.0.0 cabal-install-3.0.0.0 --stack-yaml $(YAML_FILE)
```
fendor: Maybe document this why we are using stack to install cabal? Why not choco?

.azure/linux-stack.yml
jneira: but then `stack run` should be used in the next line, no?
I used install+exec taking the circleci config as a model

.azure/windows-installhs-cabal.yml
jneira: I checked it and it seems choco is available for azure windows agents so it would be doable.
But i would left it to another pr, if you dont mind (adding a TODO)

.azure/linux-stack.yml
fendor: I thought `stack run` doesnt exist aynmore? 
However, in install.hs we are using `stack build` and `stack exec`.

.azure/windows-installhs-cabal.yml
fendor: adding a todo is fine for me!

.azure/linux-stack.yml
jneira: sorry i wanted to mean poorly that i just have tested `stack build + stack exec` and it worked, so i will change

jneira:
[After testing we already can use the hlint data file](https://github.com/haskell/haskell-ide-engine/issues/1143#issuecomment-571521806) under the same directory of hie executable i've change the scripts to include it:
* with `stack unpack hlint --stack-yaml=../$(YAML_FILE)`we'll get the appropiate hlint version

I've added `-v`to all tar commands as suggested by @fendor

jneira:
> I am thinking of starting a release process, should I wait for this first?

@alanz @fendor i think we could start the release process labeling it as beta after merging this:

* There is a unofficial azure account named `haskell-ide-engine`: https://dev.azure.com/haskell-ide-engine/ created by @power-fungus 
* We could upload manually all the artifacts to github (until we get to automatically publish the artifacts to github) or/and point to azure releases
* Testing previously the precompiled binaries in clean environments
* Waiting for the issues opened by users that want to give a try to them

alanz:
I just sent a mail to the haskell committee, who manage the haskell organisation on github, I think.

Hopefully they can do something for us.

jneira:
I think it could be merged so if nobody disagree i'll do a squash&merge soon

hvr:
Where do I need to go to enable Azure releases for this repo?

jneira:
@hvr thanks a lot!

* https://azure.microsoft.com/ is the main page
* sign in with the github account
* i think then you can continue from github settings > Integration Services > Azure Pipelines

let me know if you want more details

alanz:
@jneira Given this is a CI PR, which generally has a pretty boring and not useful history, I am ok with a squash merge, if that is what you want to do.

Issue #1453, opened 2019-11-14
Support for stack scripts?


marcusbuffett:
Just C+P-ing [my question from stack overflow](https://stackoverflow.com/questions/58849093/how-to-get-hie-to-detect-packages-installed-via-stack-for-scripts):

I'm just getting started with learning how to use haskell for scripting, and I've got something like this:
```
#!/usr/bin/env stack
-- stack --resolver lts-14.13 script
{-# LANGUAGE OverloadedStrings #-}
import Turtle

main :: IO ()
main = echo "hello"
```

When I run this with `stack script.hs`, it auto-magically installs turtle and prints "hello". However, my editor gives me this error on the import line:
```
[ghcmod] [E] Could not find module ‘Turtle’
```

This makes some sense to me because the hie-wrapper I installed (from source, if it matters), isn't aware of where stack has installed the turtle package.

The way I have my HIE hooked up is like so, using coc.nvim:
```
  "languageserver": {
  		"haskell": {
			"command": "hie-wrapper",
			"rootPatterns": [
				".stack.yaml",
				"cabal.config",
				"package.yaml"
			],
			"filetypes": [
				"hs",
				"lhs",
				"haskell"
			],
			"initializationOptions": {
				"languageServerHaskell": {
				}
			},
		}
  },
```

Are there some flags or something I can pass to hie-wrapper to get it to play nicely with Haskell scripts, ie. detect packages? Would greatly appreciate any answer that makes it a bit more clear how this stuff fits together, ghc/hie/hie-wrapper/stack/cabal are a big black box to me currently, making this hard to debug.

fendor:
Unfortunately, stack scripts are not supported at the moment.

fendor:
With the updates from #1126, I think it will actually be possible to support stack scripts.

fendor:
Actually, preliminary tests  it seems like stack scripts are working right now.
What you can do right now is creating a `hie.yaml`
```yaml
cradle:
  stack:
    component: "filename.hs"
```
cc @mpickering , this works because you can pass a filepath to stack repl.

fendor:
So, this feature is implemented and improvements to it should be in a different issue, closed by #1126

PR #1548, opened 2020-01-05
Updating Changelog for 2019-12 monthly release

alanz:


fendor:
Should we mention features that we added or something like that? 
I am thinking about the accidental support for stack scripts #1453 
And now even in default mode you can open test and executable components which did not work before, and Cabal 3.0 support?

alanz:
fendor, do you have a more comprehensive list?  If you put it in a comment here, I will add them, else make a PR on my ChangeLog.

If it is only those things, I will just add them.

fendor:
I can not think about much more:
* Cabal 3.0 support.
* Support for stack scripts.
* Work on test, executable, benchmark and library components at the same time.
* Set the build-tool to use explicitly.
* Various memory leaks have been fixed. (empirical, can be omitted)
* Various performance improvements. (empirical, can be omitted)

Other important changes:
* Disabled HaRe (rename, caseSplitCmd, etc...).

PR #1552, opened 2020-01-06
Add unit-tests for Cabal-Helper cradles

fendor:


fendor:
First take on #1552 
I dont know how much time I have in the next weeks and I think it is a good start to have some unit-tests.

Issue #1556, opened 2020-01-06
Fails to build


hasufell:
Command that was run:

```sh
cabal v2-run ./install.hs --project-file install/shake.project hie-8.6.5
```

Version:

```
1.0.0.0-3-g903bd869
```

Last part of the output:

```
Linking /home/jule/git/haskell-ide-engine/dist-newstyle/build/x86_64-linux/ghc-8.6.5/haskell-ide-engine-1.0.0.0/x/hie/build/hie/hie ...
# stack (for cabal-hie-8.6.5)
Downloading lts-14.11 build plan ...
Downloaded lts-14.11 build plan.
Preparing to install GHC (tinfo6) to an isolated location.
This will not interfere with any system-level installation.
Preparing to download ghc-tinfo6-8.6.5 ...
ghc-tinfo6-8.6.5: download has begun
ghc-tinfo6-8.6.5:  355.53 KiB / 180.58 MiB (  0.19%) downloaded...
ghc-tinfo6-8.6.5:    1.29 MiB / 180.58 MiB (  0.72%) downloaded...
ghc-tinfo6-8.6.5:    1.68 MiB / 180.58 MiB (  0.93%) downloaded...
ghc-tinfo6-8.6.5:    2.01 MiB / 180.58 MiB (  1.11%) downloaded...
ghc-tinfo6-8.6.5:    2.07 MiB / 180.58 MiB (  1.15%) downloaded...
ghc-tinfo6-8.6.5:    2.19 MiB / 180.58 MiB (  1.21%) downloaded...
ghc-tinfo6-8.6.5:    2.29 MiB / 180.58 MiB (  1.27%) downloaded...
ghc-tinfo6-8.6.5:    2.44 MiB / 180.58 MiB (  1.35%) downloaded...
ghc-tinfo6-8.6.5:    2.62 MiB / 180.58 MiB (  1.45%) downloaded...
ghc-tinfo6-8.6.5:    2.85 MiB / 180.58 MiB (  1.58%) downloaded...
ghc-tinfo6-8.6.5:    3.14 MiB / 180.58 MiB (  1.74%) downloaded...
ghc-tinfo6-8.6.5:    3.57 MiB / 180.58 MiB (  1.98%) downloaded...
ghc-tinfo6-8.6.5:    4.28 MiB / 180.58 MiB (  2.37%) downloaded...
ghc-tinfo6-8.6.5:    5.39 MiB / 180.58 MiB (  2.99%) downloaded...
[...]

PR #1557, opened 2020-01-06
Don't mix stack with cabal

### LINKED ISSUE #1380: shake install with cabal shouldn't depend on having stack installed
* Currently running the shake install with cabal [needs stack to get the `symlink-bindir`/`installdir`](https://github.com/haskell/haskell-ide-engine/blob/master/install/src/Cabal.hs#L44)
* We can't simply fallback to the value in global cabal config cause install needs it to copy hie executables [prefixed with versions](https://github.com/haskell/haskell-ide-engine/blob/master/install/src/Cabal.hs#L53-L56) and afaik there is no a simple way to get the actual value of `symlink-bindir`/`installdir` from global cabal config

* A possible solution would be fallback to a relative to cwd `./bin` path but it will make users to copy manually the executables to the final location 
* Another solution would be add options to shake install to let users do somethink like `cabal v2-run install.hs -- hie-8.6.5 --cabal-options="--installdir=path"` 
  * It would be a improvement itself but it would add complexity to build

//cc @power-fungus

hasufell:
Fixes #1380 and #1556

### DIFF
--- a/install/src/Cabal.hs
+++ b/install/src/Cabal.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
+
 module Cabal where
 
 import           Development.Shake
@@ -15,16 +17,26 @@ import           System.Directory                         ( findExecutable
 import           Version
 import           Print
 import           Env
-import           Stack
+import           Data.Functor.Identity
+#if RUN_FROM_STACK
+import           Control.Exception                        ( throwIO )
+#else
+import           Cabal.Config
+#endif
+
+getInstallDir :: IO FilePath
+#if RUN_FROM_STACK
+-- we should never hit this codepath
+getInstallDir = throwIO $ userError "Stack and cabal should never be mixed"
+#else
+getInstallDir = runIdentity . cfgInstallDir <$> readConfig
+#endif
 
 execCabal :: CmdResult r => [String] -> Action r
-execCabal = execCabalWithOriginalPath
+execCabal = command [] "cabal"
 
 execCabal_ :: [String] -> Action ()
-execCabal_ = execCabalWithOriginalPath
-
-execCabalWithOriginalPath :: CmdResult r => [String] -> Action r
-execCabalWithOriginalPath = withoutStackCachedBinaries . (command [] "cabal")
+execCabal_ = command [] "cabal"
 
 cabalBuildData :: Action ()
 cabalBuildData = do
@@ -51,7 +63,7 @@ cabalBuildHie versionNumber = do
 
 cabalInstallHie :: VersionNumber -> Action ()
 cabalInstallHie versionNumber = do
-  localBin <- getLocalBin
+  localBin <- liftIO $ getInstallDir
   cabalVersion <- getCabalVersion
   ghcPath <- getGhcPathOfOrThrowError versionNumber
 
@@ -83,18 +95,6 @@ cabalInstallHie versionNumber = do
              ++ majorVerExe ++ " and "
              ++ minorVerExe
              ++ " to " ++ localBin
-
-installCabalWithStack :: Action ()
-installCabalWithStack = do
-  -- try to find existing `cabal` executable with appropriate version
-  mbc <- withoutStackCachedBinaries (liftIO (findExecutable "cabal"))
-
-  case mbc of
-    Just c  -> do
-      cabalVersion <- checkCabal
-      printLine $ "There is already a cabal executable in $PATH with the required minimum version: " ++ cabalVersion
-     -- install `cabal-install` if not already installed
-    Nothing ->  execStackShake_ ["install", "cabal-install"]
 
 checkCabal_ :: Action ()
 checkCabal_ = checkCabal >> return ()
--- a/install/src/Cabal.hs
+++ b/install/src/Cabal.hs
@@ -51,16 +51,6 @@ getGhcPathOfOrThrowError versionNumber =
       error (ghcVersionNotFoundFailMsg versionNumber)
     Just p -> return p
 
-cabalBuildHie :: VersionNumber -> Action ()
-cabalBuildHie versionNumber = do
-  ghcPath <- getGhcPathOfOrThrowError versionNumber
-  execCabal_
-    [ "v2-build"
-    , "-w", ghcPath
-    , "--write-ghc-environment-files=never"
-    , "--max-backjumps=5000"
-    , "--disable-tests"]
-
 cabalInstallHie :: VersionNumber -> Action ()
 cabalInstallHie versionNumber = do
   localBin <- liftIO $ getInstallDir
@@ -77,6 +67,7 @@ cabalInstallHie versionNumber = do
     , "-w", ghcPath
     , "--write-ghc-environment-files=never"
     , installDirOpt, localBin
+    , "--max-backjumps=5000"
     , "exe:hie"
     , "--overwrite-policy=always"
     ]
--- a/install/src/HieInstall.hs
+++ b/install/src/HieInstall.hs
@@ -48,85 +48,64 @@ defaultMain = do
 
   -- used for cabal-based targets
   ghcPaths <- findInstalledGhcs
-  let ghcVersions = map fst ghcPaths
+  let cabalVersions = map fst ghcPaths
 
   -- used for stack-based targets
-  hieVersions <- getHieVersions
+  stackVersions <- getHieVersions
 
-  let versions = BuildableVersions { stackVersions = hieVersions
-                                   , cabalVersions = ghcVersions
-                                   }
+  let versions = if isRunFromStack then stackVersions else cabalVersions 
 
-  let latestVersion = last hieVersions
+  let toolsVersions = BuildableVersions stackVersions cabalVersions
 
-  putStrLn $ "run from: " ++ buildSystem
+  let latestVersion = last versions
 
   shakeArgs shakeOptions { shakeFiles = "_build" } $ do
     want ["short-help"]
     -- general purpose targets
     phony "submodules"  updateSubmodules
     phony "short-help"  shortHelpMessage
-    phony "all"         shortHelpMessage
-    phony "help"        (helpMessage versions)
-    phony "check-stack" checkStack
-    phony "check-cabal" checkCabal_
+    phony "help"        (helpMessage toolsVersions)
+    
+    phony "check" (if isRunFromStack then checkStack else checkCabal_)
 
-    phony "cabal-ghcs" $ do
-      let
-        msg =
-          "Found the following GHC paths: \n"
-            ++ unlines
-                 (map (\(version, path) -> "ghc-" ++ version ++ ": " ++ path)
-                      ghcPaths
-                 )
-      printInStars msg
+    phony "data" $ do
+      need ["submodules"]
+      need ["check"]
+      if isRunFromStack then stackBuildData else cabalBuildData
 
-    -- default-targets
-    phony "build" $ need [buildSystem ++ "-build"]
-    phony "build-latest" $ need [buildSystem ++ "-build-latest"]
-    phony "build-data" $ need [buildSystem ++ "-build-data"]
     forM_
-      (getDefaultBuildSystemVersions versions)
-      (\version ->
-        phony ("hie-" ++ version) $ need [buildSystem ++ "-hie-" ++ version]
-      )
-
-    -- stack specific targets
-    when isRunFromStack $ do
-      phony "stack-build-latest" (need ["stack-hie-" ++ last hieVersions])
-      phony "stack-build"  (need ["build-data", "stack-build-latest"])
-      phony "stack-build-data" $ do
+      versions
+      (\version -> phony ("hie-" ++ version) $ do
         need ["submodules"]
-        need ["check-stack"]
-        stackBuildData
-      forM_
-        hieVersions
-        (\version -> phony ("stack-hie-" ++ version) $ do
-          need ["submodules"]
-          need ["check-stack"]
+        need ["check"]
+        if isRunFromStack then do 
           stackBuildHie version
           stackInstallHie version
-        )
+        else
+          cabalInstallHie version
+      )
+
+    phony "latest" (need ["hie-" ++ latestVersion])
+    phony "build"  (need ["data", "latest"])
 
     -- cabal specific targets
     when isRunFromCabal $ do
-      phony "cabal-build-latest" (need ["cabal-hie-" ++ last ghcVersions])
-      phony "cabal-build"  (need ["build-data", "cabal-build-latest"])
-      phony "cabal-build-data" $ do
-        need ["submodules"]
-        cabalBuildData
-      forM_
-        ghcVersions
-        (\version -> phony ("cabal-hie-" ++ version) $ do
-          need ["submodules"]
-          cabalInstallHie version
-        )
+
+      phony "ghcs" $ do
+        let
+          msg =
+            "Found the following GHC paths: \n"
+              ++ unlines
+                  (map (\(version, path) -> "ghc-" ++ version ++ ": " ++ path)
+                        ghcPaths
+                  )
+        printInStars msg
 
     -- macos specific targets
     phony "icu-macos-fix"
           (need ["icu-macos-fix-install"] >> need ["icu-macos-fix-build"])
     phony "icu-macos-fix-install" (command_ [] "brew" ["install", "icu4c"])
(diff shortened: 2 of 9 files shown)


jneira:
Thanks for the pr, but i m afraid it would break some use cases:
* now `stack` users can start to use the right version of cabal to work with hie symply running `stack install.hs stack-install-cabal`
  * I think we should keep that use case so i would not remove `installCabalWithStack`
* The script lets users run the script with one tool but install hie with the other: `stack install.hs cabal-hie-8.6.5` or `cabal v2-run install.hs stack-hie-8.6.5`
  * Run the script with stack and then use cabal to install hie needs `execCabalWithOriginalPath` to use cabal from the user `$PATH` and no the cached one locally by `stack`
  * I am not sure what was the reason for use both tools at once and if we could remove it, maybe @power-fungus could help with that
  * If we remove that double tool uses we should remove `cabal-*` targets for stack and `stack-*` targets for cabal

fendor:
> I am not sure what was the reason for use both tools at once and if we could remove it, maybe @power-fungus could help with that

As far as I know, hie is completely build-tool agnostic. Installation of `cabal-install` is some legacy stuff that we picked up back from the Makefile, iirc.

> now stack users can start to use the right version of cabal to work with hie simply running stack install.hs stack-install-cabal

Same as above, I think, that we can use hie without any cabal on the path.

> The script lets users run the script with one tool but install hie with the other: stack install.hs cabal-hie-8.6.5 or cabal v2-run install.hs stack-hie-8.6.5

If users want to use stack scripts but to install hie with cabal, I think they can be bothered to install cabal themselves. Moreover, we are inconsistent, because you can not do the same the other way around. You can not use cabal to install stack, right?

> If we remove that double tool uses we should remove cabal-* targets for stack and stack-* targets for cabal

I think that could make sense.

jneira:
Well, i think that use both tools is a pretty common scenario although now hie doesnt require to have cabal installed for all projects (you still need it for projects without `stack.yaml`), so install the right version of cabal could be handy. But i am not strongly against remove it if you think it does not worth the added complexity.

>You can not use cabal to install stack, right?

He, fair enough. I guess there are more cases of users using stack that wants to use or needs cabal that the other way around. But if we keep one i would accept a pr to add the other 😉 

But the above is only about the possibility of install cabal using stack. I would remove the `cabal-*` and `stack-*` targets if there is no good reasons to keep them.

fendor:
I think it is a good idea to decouple cabal and stack builds, e.g. a cabal build *should not* depend on stack in any way. 
However, retaining the possibility to install cabal via stack and vice-versa is fine for me.

> But the above is only about the possibility of install cabal using stack. I would remove the cabal-* and stack-* targets if there is no good reasons to keep them.

I think that makes sense.

hasufell:
IMO, packages should **never** install toolchain and build tools unquestioned. I don't want to use stack and it breaks HIE installation as you can see in https://github.com/haskell/haskell-ide-engine/issues/1556

People using HIE are **developers** and they should be bothered to install their toolchain and should expect to have proper control over it without packages arbitrarily invoking stack, downloading GHCs that I have already installed and messing with my global state.

> I think that makes sense.

I cannot follow. Is there anything actionable for this PR?

fendor:
> I cannot follow. Is there anything actionable for this PR?

I dont think so, that is something for a follow-up PR.

> Edit: ultimately, when looking at the shakefile, I think it is not useful at all. Almost everything is phony rules, which defeats the entire purpose of shake. It seems this can be easily rewritten with a Makefile or a simple shell script. But this is outside of this PRs scope, which just unbreaks HIE.

I agree partially. A Makefile is not platform independent, so no, that would not be a solution. Neither is a shell script. The initial reason was to simplify the installation, which, judging by the amount of issues we receive now and before using this installation script, succeeded.
The shake script was written when installation was more complex. However, using shake is overkill and it could be reduced to a simpler haskell script.

hasufell:
> A Makefile is not platform independent, so no, that would not be a solution. Neither is a shell script.

Right, I forgot windows users. Shame on me.

install/src/Cabal.hs
```diff
@@ -84,18 +85,6 @@ cabalInstallHie versionNumber = do
              ++ minorVerExe
              ++ " to " ++ localBin
 
-installCabalWithStack :: Action ()
-installCabalWithStack = do
-  -- try to find existing `cabal` executable with appropriate version
-  mbc <- withoutStackCachedBinaries (liftIO (findExecutable "cabal"))
```
jneira: I think we should keep this one for the reasons i gave in my comment.
In any case, if it is removed we should remove `withoutStackCachedBinaries` cause no one will be using it.

install/src/HieInstall.hs
```diff
@@ -115,13 +114,11 @@ defaultMain = do
     phony "cabal-build"  (need ["build-data", "cabal-build-latest"])
     phony "cabal-build-data" $ do
       need ["submodules"]
-      need ["cabal"]
       cabalBuildData
     forM_
       ghcVersions
       (\version -> phony ("cabal-hie-" ++ version) $ do
         need ["submodules"]
-        need ["cabal"]
```
jneira: If we remove the cabal target we should remove all `cabal-*` targets from stack executions and all `stack-*` targets from cabal executions

fendor: I think that is an unrelated change. The `cabal` target is for checking an existing installation of cabal. The `cabal-*` and `stack-*` targets can be removed in another PR.

jneira:
Trying to execute the script with stack throws an build error cause `cabal-install-parsers` is not in stackage. Adding it to `install/shake.yaml` make stack ask for other packages:
```
In the dependencies for cabal-install-parsers-0.2:
    Cabal-2.4.1.0 from stack configuration does not match ^>=3.0.0.0  (latest matching version is 3.0.0.0)
    aeson-1.4.5.0 from stack configuration does not match ^>=1.4.6.0  (latest matching version is 1.4.6.0)
    binary-instances must match ^>=1, but the stack configuration has no specified version  (latest matching version is 1)
    lukko must match ^>=0.1.1, but the stack configuration has no specified version  (latest matching version is 0.1.1.1)
needed due to hie-install-0.8.0.0 -> cabal-install-parsers-0.2
```

Not sure if we can get a package set to build this version with stack. Maybe some cpp would be needed (as anticipated in https://github.com/haskell/haskell-ide-engine/issues/1380#issuecomment-569356999)

Ideally the script should be buildable with all resolver of stack-*.yaml of hie, to let users use the same ghc building the script and building hie.

hasufell:
> Trying to execute the script with stack throws an build error

Hm, why did the CIs succeed then? :confused:

fendor:
@hasufell Currently enabled CI is actually not even building the shake script... There is some CI for azure, but azure is not yet enabled for this project :(

install/src/HieInstall.hs
jneira: Fine with make it in other pr, but we have to take in account that  `stack install.hs cabal-hie-8.6.5` will not check the cabal version and will not install them if it is not present.

hasufell:
I CPPed the relevant bits and also changed the README. I didn't run a full stack build, but at least it starts building.

install/src/Cabal.hs
```diff
@@ -15,16 +17,26 @@ import           System.Directory                         ( findExecutable
...
-import           Stack
+import           Data.Functor.Identity
+#if RUN_FROM_STACK
+import           Control.Exception                        ( throwIO )
+#else
+import           Cabal.Config
+#endif
+
+getInstallDir :: IO FilePath
+-- we should never hit this codepath, this i
```
jneira: It seems there is a missing phrase here

install/src/Cabal.hs
```diff
@@ -15,16 +17,26 @@ import           System.Directory                         ( findExecutable
...
+#if RUN_FROM_STACK
+getInstallDir = throwIO $ userError "Shake and cabal should never be mixed"
```
jneira:
```suggestion
getInstallDir = throwIO $ userError "Stack and cabal executions are isolated"
```

jneira:
@hasufell have you tested it with `cabal-2.4.1.0`? If it doesnt work with that version we should bump up the cabal required version to `3.0.0.0`

install/src/HieInstall.hs
```diff
@@ -93,15 +92,14 @@ defaultMain = do
...
-    phony "stack-build-data" $ do
+    when isRunFromStack $ phony "stack-build-latest" (need ["stack-hie-" ++ last hieVersions])
+    when isRunFromStack $ phony "stack-build"  (need ["build-data", "stack-build-latest"])
+
+    when isRunFromStack $ phony "stack-build-data" $ do
       need ["submodules"]
       need ["check-stack"]
       stackBuildData
-    forM_
+    when isRunFromStack $ forM_
```
jneira: maybe we could group all stack based targets under the same condition:
```haskell
when isRunFromStack $ do
  phony "stack-build-latest" (need ["stack-hie-" ++ last hieVersions])
  phony "stack-build"  (need ["build-data", "stack-build-latest"])
  -- .....
```

install/src/HieInstall.hs
```diff
@@ -111,17 +109,15 @@ defaultMain = do
       )
 
     -- cabal specific targets
-    phony "cabal-build-latest" (need ["cabal-hie-" ++ last ghcVersions])
-    phony "cabal-build"  (need ["build-data", "cabal-build-latest"])
-    phony "cabal-build-data" $ do
+    when isRunFromCabal $ phony "cabal-build-latest" (need ["cabal-hie-" ++ last ghcVersions])
```
jneira: Same for cabal based targets:
```haskell
when isRunFromCabal $ do
  phony "cabal-build-latest" (need ["cabal-hie-" ++ last ghcVersions])
  phony "cabal-build"  (need ["build-data", "cabal-build-latest"])
  -- .....
```

or even better:
```haskell
phony (buildSystem ++ "-build-latest") (need [buildsystem ++ "-hie-" ++ last ghcVersions])
phony (buildSystem ++ "-build")  (need ["build-data",buildsystem ++ "-build-latest"])
-- etc
```

install/src/HieInstall.hs
```diff
@@ -111,17 +109,15 @@ defaultMain = do
       )
 
     -- cabal specific targets
-    phony "cabal-build-latest" (need ["cabal-hie-" ++ last ghcVersions])
-    phony "cabal-build"  (need ["build-data", "cabal-build-latest"])
```
jneira: The help target continues showing the previous targets for both tools:
```
D:\ws\haskell\haskell-ide-engine>cabal v2-run install.hs --project-file install\shake.project help
Resolving dependencies...
......
Linking D:\ws\haskell\haskell-ide-engine\dist-newstyle\build\x86_64-windows\ghc-8.6.5\fake-package-0\x\script\build\script\script.exe ...
run from: cabal

Usage:
    stack install.hs <target>
    or
    cabal v2-run install.hs --project-file install/shake.project <target>

Targets:
    help                   Show help message including all targets

    build                  Build hie with the latest available GHC and the data files
    build-latest           Build hie with the latest available GHC
    build-data             Get the required data-files for `hie` (Hoogle DB)
    hie-8.4.4              Builds hie for GHC version 8.4.4
    hie-8.6.4              Builds hie for GHC version 8.6.4
    hie-8.6.5              Builds hie for GHC version 8.6.5

    stack-build            Build hie with the latest available GHC and the data files; with stack
    stack-build-latest     Build hie with the latest available GHC; with stack
    stack-build-data       Get the required data-files for `hie` (Hoogle DB); with stack
    stack-hie-8.4.2        Builds hie for GHC version 8.4.2; with stack
    stack-hie-8.4.3        Builds hie for GHC version 8.4.3; with stack
    stack-hie-8.4.4        Builds hie for GHC version 8.4.4; with stack
    stack-hie-8.6.1        Builds hie for GHC version 8.6.1; with stack
    stack-hie-8.6.2        Builds hie for GHC version 8.6.2; with stack
    stack-hie-8.6.4        Builds hie for GHC version 8.6.4; with stack
    stack-hie-8.6.5        Builds hie for GHC version 8.6.5; with stack

    cabal-ghcs             Show all GHC versions that can be installed via `cabal-build`.
    cabal-build            Build hie with the latest available GHC and the data files; with cabal
    cabal-build-latest     Build hie with the latest available GHC; with cabal
    cabal-build-data       Get the required data-files for `hie` (Hoogle DB); with cabal
    cabal-hie-8.4.4        Builds hie for GHC version 8.4.4; with cabal
    cabal-hie-8.6.4        Builds hie for GHC version 8.6.4; with cabal
    cabal-hie-8.6.5        Builds hie for GHC version 8.6.5; with cabal

    icu-macos-fix          Fixes icu related problems in MacOS

Build completed in 0.04s
```

It should be now something like:

```
Targets:
    help                   Show help message including all targets

    build                  Build hie with the latest available GHC and the data files
    build-latest           Build hie with the latest available GHC
    build-data             Get the required data-files for `hie` (Hoogle DB)
    hie-8.4.4              Builds hie for GHC version 8.4.4
    hie-8.6.4              Builds hie for GHC version 8.6.4
    hie-8.6.5              Builds hie for GHC version 8.6.5

    icu-macos-fix          Fixes icu related problems in MacOS
```

jneira:
@hasufell thanks for include the remove of obsolete targets, it looks much better now. 

As for `stack-install-cabal`, i would accept remove it only to being able to remove the `withoutStackCachedBinaries ` monstrosity. I hope users will not miss it but we alway can restore it.

Let me know if you need any help with the last requested changes, i promise they are the last ones :smile:

install/src/Cabal.hs
hasufell: It was late :laughing:

install/src/HieInstall.hs
jneira: Or take ride of prefixed targets entirily and simple executing the targets with one tool or another

hasufell:
> As for `stack-install-cabal`, i would accept remove it only to being able to remove the `withoutStackCachedBinaries ` monstrosity.

That was already done.

hasufell:
> have you tested it with `cabal-2.4.1.0`?

Seems to work fine

hasufell:
If you have further remarks, I suggest you push your changes to this PR.

bubba:
I am personally in favour of not having the install script install cabal via stack, which seems kind of sneaky. Can we double check that hie still relies on it?

fendor:
@bubba Hie does not rely on it and afaik never really has. I *think* it was needed to install Cabal to speed up cabal-helper, but no one is even sure that this was ever necessary.

bubba:
@fendor probably just best then to let cabal-helper install Cabal? Also seeing as with a hie.yaml file it might not be needed

fendor:
I am not sure that *ever* worked, since cabal-helper installs Cabal into `~/.cache/cabal-helper/Cabal` or something like that. And we used `cabal v1-install`. Currently just cabal-helper takes care of it and I think it is good the way it is.

bubba:
> I am not sure that _ever_ worked, since cabal-helper installs Cabal into `~/.cache/cabal-helper/Cabal` or something like that. And we used `cabal v1-install`. Currently just cabal-helper takes care of it and I think it is good the way it is.

cabal-helper searches for it installed globally first before installing its own copy, but agreed anyway

fendor:
@bubba Ah, did not know that! however, we are not installing Cabal for some time now so, it seems like nobody is complaining right now.

jneira:
> If you have further remarks, I suggest you push your changes to this PR.

I'll pushed the required changes upt to +85-202 :grin:
@hasufell thanks a lot for the pr, the script is cleaner now

The help looks now:
```
Usage:
    stack install.hs <target>
    or
    cabal v2-run install.hs --project-file install/shake.project <target>

Targets:
    help                Show help message including all targets

    hie                 Build hie with the latest available GHC and the data files
    latest              Build hie with the latest available GHC
    data                Get the required data-files for `hie` (Hoogle DB)
    hie-8.4.4           Builds hie for GHC version 8.4.4
    hie-8.6.4           Builds hie for GHC version 8.6.4
    hie-8.6.5           Builds hie for GHC version 8.6.5

    ghcs                Show all GHC versions that can be installed via `cabal-build`.

    icu-macos-fix       Fixes icu related problems in MacOS
```
I have taken the opportunity to change some target names and make them shorter:
* build -> hie
* build-latest -> latest
* build-data -> data
* cabal-ghcs -> ghcs

Do you agree with new names @fendor @bubba?
We'll have to change the docs in the vscode extension to match them

bubba:
@jneita LGTM

PR #1559, opened 2020-01-06
Readme lsp flag

### LINKED ISSUE #1558: Add --lsp flag to integration instructions in README

Anrock:
Adds --lsp flag to integration instructions for vim+coc and vim+languageclient
Part of #1558

### DIFF
--- a/README.md
+++ b/README.md
@@ -589,7 +589,7 @@ into `~/.vim/pack/XXX/start/`, where `XXX` is just a name for your "plugin suite
 
 ```vim
 set rtp+=~/.vim/pack/XXX/start/LanguageClient-neovim
-let g:LanguageClient_serverCommands = { 'haskell': ['hie-wrapper'] }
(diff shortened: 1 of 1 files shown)


andys8:
I updated https://github.com/neoclide/coc.nvim/wiki/Language-servers#haskell with `--lsp`. Is this flag mandatory?

alanz:
Yes, it is, to allow dumping human readable debug info it run without it.

andys8:
Okay, good to know.

Maybe too late, but I'd prefer the end-user having no flags meaning less ways to mis-configure the setup. On the other hand, there doesn't seem to be a consistent way for language server configuration anyway ([different language server implementations ](https://github.com/neoclide/coc.nvim/wiki/Language-servers))

alanz:
Well, when the end-user is debugging running without flags is a good option. It also mirrors the behaviour in `ghcide`, so gives some consistency.

It also was the standard for a very long time, only changed recently.

PR #1562, opened 2020-01-06
Avoid building HIE twice

hasufell:
cabal install works different than cabal build
and does not re-use its artifacts. Even some
dependencies may be rebuilt.

Future cabal versions may fix that, we don't know.


jneira:
Wow i thought it reused always the dependencies 🤔

fendor:
I think it is reusing the dependencies, or at least most, but you still re-building the main-lib, like hie-plugin-api and hie itself which takes quite some time, too.

PR #1571, opened 2020-01-12
Enable all working test suites and add linux-cabal job in Azure

jneira:
* Enable test-suites:
  * In linux, with stack or cabal, all test suites are enabled 😄 
  * In macos:
      * ghc-8.6.*: all suites enabled
      * ghc-8.4.*: func-test and dispatcher-test disabled
  * In windows
      * Only unit-test enabled 😟
             * Except CabalHelper ones for ghc-8.6.4: see https://github.com/DanielG/cabal-helper/issues/91
* cabal jobs:
   * I've added one for unix, the next one will be add it for macos and windows


jneira:
I am afraid it takes many time to finish,, what do you think about support only the last two minor versions of ghc?

.azure/linux-cabal.yml
```diff
@@ -0,0 +1,75 @@
...
+    displayName: Sync submodules
+  - bash: |
+      source .azure/linux-cabal.bashrc
+      cabal v2-update
+      cabal v2-build --only-dependencies
+    displayName: Build dependencies
+  - bash: |
+      source .azure/linux-cabal.bashrc
+      cabal v2-build
+    displayName: Build `hie`
```
fendor: As we have seen in #1562, this does not seem to be necessary. However, it is a minor reduction in time-consumption, thus, nothing to worry about or even necessary to change, imo.

.azure/linux-cabal.yml
```diff
@@ -0,0 +1,75 @@
...
+      source .azure/linux-cabal.bashrc
+      cabal v2-build --enable-tests --enable-benchmarks --only-dependencies
+    displayName: Build Test-dependencies
+  - bash: |
+      sudo apt update
+      sudo apt install z3
+    displayName: "Install Runtime Test-Dependencies: z3"
+  - bash: |
+      source .azure/linux-cabal.bashrc
+      # to not resintall hie
```
fendor: s/resintall/reinstall/

.azure/linux-cabal.yml
```diff
@@ -0,0 +1,75 @@
...
+      cabal v2-install liquidhaskell-0.8.6.2 -w /opt/ghc/8.6.5/bin/ghc
+    displayName: "Install Runtime Test-Dependencies: liquidhaskell"
+  - bash: |
+      source .azure/linux-cabal.bashrc
+      cabal v2-build hoogle
+      cabal v2-exec hoogle generate
+    displayName: "Install Runtime Test-Dependencies: hoogle database"
+  - bash: |
+      source .azure/linux-cabal.bashrc
+      cabal v2-test -j1
```
fendor: why only one job?

.gitignore
```diff
@@ -36,6 +36,7 @@
 /docs/source/__pycache__/
 /docs/source/plugins/
 /elisp/.cask/
+/test/testdata/*.exe
```
fendor: What exe files are generated? Out of interest, not really a review comment.

.azure/linux-cabal.yml
jneira: Sadly it fails if testing use parallelism, like in circleci:
https://github.com/haskell/haskell-ide-engine/blob/92ac896029a5ba5f740ebca168533c83848548c6/.circleci/config.yml#L85

Maybe we could investigate why...

.gitignore
jneira: ./test/testdata/FileWithWarning.exe
./test/testdata/HareRename.exe

.azure/linux-cabal.yml
jneira: Yeah, only to make it explicit as a separate step that will fail if the code has some compile time error

jneira:
> It looks like it is taking longer than stack builds because the cache isnt used, yet. Maybe it is not warmed-up yet?

I am afraid that run the jobs with test enabled and full caching takes 40 min on average (+-) so 
40 jobs / 10 parallel jobs * 40 min =160 min, with all jobs running the func-test

And we want to add more cabal jobs so time will be increased.

We could only run the func-test suite for last minor versions and keep the builds and the rest of suites?

.azure/linux-cabal.yml
fendor: A TODO maybe for the future?

fendor:
I think these times are fine.

jneira:
> I think these times are fine.

Mmm well the only caveat is they make pr's heavier and somewhat less agile but if you are fine, me too :grinning:

jneira:
I am trying to enable the maximum functional tests possible on windows

PR #1482, opened 2019-12-20
GHC 8.8 support

Avi-D-coder:
I'm not particularly happy with per file pragmas, demoting unused imports to a globally warning is cleaner.

Once `apply-refact` is updated we can remove the submodule. Preferably before merging.

PR #1579, opened 2020-01-20
Ormolu 8.8

Avi-D-coder:
This is a possible stop gap, needed until https://github.com/tweag/ormolu/pull/501 lands in hackage.
The alternative is setting `allow-newer`.

Follow up to #1482

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -18,6 +18,8 @@ write-ghc-environment-files: never
 
 index-state: 2020-01-07T22:09:46Z
 
+allow-older: ormolu:base
+
 source-repository-package
   type: git
   location: https://github.com/Avi-D-coder/ormolu.git
--- a/cabal.project
+++ b/cabal.project
@@ -17,3 +17,8 @@ constraints:
 write-ghc-environment-files: never
 
 index-state: 2020-01-07T22:09:46Z
+
+source-repository-package
+  type: git
+  location: https://github.com/Avi-D-coder/ormolu.git
+  tag: dbd93d753cf5370620e235fea91cc029f2e988fb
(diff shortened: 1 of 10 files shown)


Avi-D-coder:
Forgot about cabal

jneira:
Hi! thanks for the quick fix.
I think `stack-8.8.1.yaml` would need a similar git dependency to work.
Other times we've added the forked project (`cabal-helper`, `hare`, etc) to submodules and as local package in build config. Not sure if it can be applied to `ormolu`.
Personally i am fine with git dependencies

Avi-D-coder:
My reason for not using a submodule is that only 8.8 needs it. If I'm not mistaken git dependency support in cabal is relatively new, so was not an option before.

Avi-D-coder:
Azure Windows_Stack stack-8.4.2 appears to be another instance of https://github.com/haskell/haskell-ide-engine/pull/1482#issuecomment-576055839

jneira:
@Avi-D-coder yeah, we could ignore for now the Session time outs

Mmm, `Linux_Stack stack-8.8.1` is failing due a slighty diff output:
```
haskell-ide-engine>   test/functional/FormatSpec.hs:106:23: 
haskell-ide-engine>   1) Format.ormolu formats correctly
haskell-ide-engine>        expected: "module    Format where\nfoo   :: Int ->  Int\nfoo  3 = 2\nfoo    x  = x\nbar   :: String ->   IO String\nbar s =  do\n      x <- return \"hello\"\n      return \"asdf\"\n      \n"
haskell-ide-engine>         but got: "module Format where\n\nfoo :: Int -> Int\nfoo 3 = 2\nfoo x = x\n\nbar :: String -> IO String\nbar s = do\n  x <- return \"hello\"\n  return \"asdf\"\n"
haskell-ide-engine> 
haskell-ide-engine>   To rerun use: --match "/Format/ormolu/formats correctly/"
```
i guess it is not reproduced in local?

alanz:
I suggest we wait a little to give the ormolu guys a chance to respond to their PR, and if not merge this.

PR #1582, opened 2020-01-21
Pass Ormolu cradle flags & default-extensions

Avi-D-coder:
Other formatters, may be able to use this, but since other formatters have config files they don't need it.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
@@ -134,6 +134,8 @@ import           Haskell.Ide.Engine.Config
 import           Haskell.Ide.Engine.GhcModuleCache
 import           Haskell.Ide.Engine.MultiThreadState
 
+import           HIE.Bios
+
 import qualified Language.Haskell.LSP.Core     as Core
 import           Language.Haskell.LSP.Types.Capabilities
 import           Language.Haskell.LSP.Types     ( Command(..)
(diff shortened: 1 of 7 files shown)


Avi-D-coder:
All the test failures so far are unrelated

fendor:
I think we need to change the caching mechanism to give us also raw access to the compile flags for each file.

fendor:
Another suggestion, does ormolu create a ghc session in order to correctly parse the source? If so, we can plug this interface into our own ghc session where all DynFlags are already accessible.
That is actually the suggestion for `brittany`, dont open a new ghc session but reuse the existing one, as well as re-using already parsed sources, avoiding a lot of work in the process.

Avi-D-coder:
@fendor if I recall a ghc session somewhere in there, but it's not exposed. Caching `getCompilerOptions` is the way to go.

hie-plugin-api/Haskell/Ide/Engine/PluginsIdeMonads.hs
```diff
@@ -261,6 +263,7 @@ data FormattingType = FormatText
 -- and file uri, does not suffice.
 type FormattingProvider = T.Text -- ^ Text to format
         -> Uri -- ^ Uri of the file being formatted
+        -> Maybe Cradle -- ^ HIE.Bios Cradle containing the file being formatted
```
bubba: I would rather the cradle be accessed via a method of `IdeM` rather than passing it, since accessing the cradle is also probably useful in other contexts, and means we don't have to extract it out and pass it along inside the reactor

Avi-D-coder:
I'll finish up the rewrite in a few hour and reopen.

PR #1584, opened 2020-01-21
Azure builds improvements

jneira:
* Tests
  * Fix a intermittent failing test in `FunctionalLiquidSpec` due to interleaved diagnostics responses from the server
  * Add to `CabalHelperSpec` a test to check explicitly that `stack`and `cabal` are in $PATH
* Azure
  * Change `stack-root`to `D:` cause it seems the `C:` drive of azure agents is already almost full


jneira:
The last push only changed commit messages

PR #1589, opened 2020-01-22
Pass Ormolu cradle flags & default-extensions

Avi-D-coder:
Take 2

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
@@ -4,6 +4,8 @@
 
 module Haskell.Ide.Engine.GhcModuleCache where
 
+import           Control.Monad.IO.Class ( liftIO , MonadIO(..) )
+
 import qualified Data.Map as Map
 import           Data.Dynamic (Dynamic)
 import           Data.Typeable (TypeRep)
@@ -100,21 +102,59 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
 lookupCradle :: FilePath -> GhcModuleCache -> LookupCradleResult
 lookupCradle fp gmc =
   case currentCradle gmc of
-    Just (dirs, _c) | (any (\d -> d `isPrefixOf` fp) dirs) -> ReuseCradle
+    Just (dirs, _c, _) | any (`isPrefixOf` fp) dirs -> ReuseCradle
     _ -> case T.match (cradleCache gmc) (B.pack fp) of
            Just (_k, c, _suf) -> LoadCradle c
            Nothing  -> NewCradle fp
 
-data CachedCradle = CachedCradle BIOS.Cradle HscEnv
+getComponentOptions
+  :: (MonadIO m, HasGhcModuleCache m)
+  => FilePath
+  -> m (Maybe BIOS.ComponentOptions)
+getComponentOptions fp = do
+  mc <- getModuleCache
+  case currentCradle mc of
+    Just (dirs, cradle, co) | any (`isPrefixOf` fp) dirs -> case co of
+      Just _ -> return co
+      _      -> setCo fp cradle setCur
+    _ -> case T.match (cradleCache mc) (B.pack fp) of
+      Just (_, CachedCradle _ _ (Just co), _) -> return $ Just co
+      Just (p, CachedCradle cradle _ Nothing, _) ->
+        setCo (B.unpack p) cradle setCac
+      _ -> return Nothing
+ where
+  setCo fp' cradle mod' = do
+    res <- liftIO $ BIOS.getCompilerOptions fp' cradle
+    case res of
+      BIOS.CradleSuccess opts -> do
+        modifyModuleCache $ mod' opts fp'
+        return $ Just opts
+      _ -> return Nothing
+  setCur opts _ mc'@(GhcModuleCache _ _ (Just (fps, c', _))) =
+    mc' { currentCradle = Just (fps, c', Just opts) }
+  setCur _ _ mc' = mc'
+  setCac opts fp' mc' = mc'
+    { cradleCache = T.adjust (\cc -> cc { mCompOpts = Just opts }) (B.pack fp')
+                      $ cradleCache mc'
+    }
+
+
+
+data CachedCradle = CachedCradle
+  { ccradle :: BIOS.Cradle
+  , hscEnv :: HscEnv
+  , mCompOpts :: Maybe BIOS.ComponentOptions
+  }
 
 instance Show CachedCradle where
-  show (CachedCradle x _) = show x
+  show (CachedCradle x _ _) = show x
 
 data GhcModuleCache = GhcModuleCache
   { cradleCache :: !(T.Trie CachedCradle)
               -- ^ map from FilePath to cradles
+              -- May not include currentCradle
   , uriCaches  :: !UriCaches
-  , currentCradle :: Maybe ([FilePath], BIOS.Cradle)
+  , currentCradle :: Maybe ([FilePath], BIOS.Cradle, Maybe BIOS.ComponentOptions)
               -- ^ The current cradle and which FilePath's it is
               -- responsible for
   } deriving (Show)
--- a/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
@@ -101,6 +101,7 @@ lookupCradle :: FilePath -> GhcModuleCache -> LookupCradleResult
 lookupCradle fp gmc =
   lookupInCache fp gmc (const $ const ReuseCradle) LoadCradle $ NewCradle fp
 
+-- | Find the cradle wide 'ComponentOptions' that apply to a 'FilePath'
 lookupComponentOptions
   :: HasGhcModuleCache m => FilePath -> m (Maybe BIOS.ComponentOptions)
 lookupComponentOptions fp = do
@@ -110,8 +111,11 @@ lookupComponentOptions fp = do
 lookupInCache
   :: FilePath
   -> GhcModuleCache
+  -- | Called when file is in the current cradle
   -> (BIOS.Cradle -> BIOS.ComponentOptions -> a)
+  -- | Called when file is a member of a cached cradle
   -> (CachedCradle -> a)
+  -- | Default value to return is a cradle is not found
   -> a
   -> a
 lookupInCache fp gmc cur cached def = case currentCradle gmc of
@@ -120,6 +124,7 @@ lookupInCache fp gmc cur cached def = case currentCradle gmc of
     Just (_k, c, _suf) -> cached c
     Nothing            -> def
 
+-- | A 'Cradle', it's 'HscEnv' and 'ComponentOptions'
 data CachedCradle = CachedCradle
   { ccradle :: BIOS.Cradle
   , hscEnv :: HscEnv
@@ -131,12 +136,12 @@ instance Show CachedCradle where
 
 data GhcModuleCache = GhcModuleCache
   { cradleCache :: !(T.Trie CachedCradle)
-              -- ^ map from FilePath to cradles
+              -- ^ map from FilePath to cradle and it's config.
               -- May not include currentCradle
   , uriCaches  :: !UriCaches
   , currentCradle :: Maybe ([FilePath], BIOS.Cradle, BIOS.ComponentOptions)
-              -- ^ The current cradle and which FilePath's it is
-              -- responsible for
+              -- ^ The current cradle, it's config,
+              -- and which FilePath's it is responsible for.
   } deriving (Show)
 
 -- ---------------------------------------------------------------------
--- a/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
@@ -4,8 +4,6 @@
 
 module Haskell.Ide.Engine.GhcModuleCache where
 
-import           Control.Monad.IO.Class ( liftIO , MonadIO(..) )
-
 import qualified Data.Map as Map
 import           Data.Dynamic (Dynamic)
 import           Data.Typeable (TypeRep)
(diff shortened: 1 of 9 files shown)


hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -100,21 +102,59 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
 
-data CachedCradle = CachedCradle BIOS.Cradle HscEnv
+getComponentOptions
+  :: (MonadIO m, HasGhcModuleCache m)
+  => FilePath
+  -> m (Maybe BIOS.ComponentOptions)
+getComponentOptions fp = do
+  mc <- getModuleCache
+  case currentCradle mc of
+    Just (dirs, cradle, co) | any (`isPrefixOf` fp) dirs -> case co of
```
bubba: The `dirs` here are always absolute paths right?

fendor: Yes they do!

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -100,21 +102,59 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
+      Just _ -> return co
+      _      -> setCo fp cradle setCur
+    _ -> case T.match (cradleCache mc) (B.pack fp) of
+      Just (_, CachedCradle _ _ (Just co), _) -> return $ Just co
+      Just (p, CachedCradle cradle _ Nothing, _) ->
+        setCo (B.unpack p) cradle setCac
+      _ -> return Nothing
+ where
+  setCo fp' cradle mod' = do
+    res <- liftIO $ BIOS.getCompilerOptions fp' cradle
```
fendor: I dont really like this. You potentially block a thread for an undetermined amount of time (potentially for minutes) and the user gets zero feedback, just wondering why nothing works. Also, on which thread gets this evaluated? I think it is the `Scheduler` thread, but I am not sure. If it is the `IDE` thread, then it is truly not acceptable as we require that nothing on the `IDE` thread has a blocking call.
 I would prefer if that function acts more like `lookupCradle` that it only looks up the component options if they are already cached, e.g. `lookupComponentOptions`.

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -245,17 +245,17 @@ setCurrentCradle cradle = do
     let ps = mapMaybe (GHC.ml_hs_file . GHC.ms_location) (mgModSummaries mg)
     debugm $ "Modules in the cradle: " ++ show ps
     ps' <- liftIO $ mapM canonicalizePath ps
-    modifyCache (\s -> s { currentCradle = Just (ps', cradle) })
+    modifyCache (\s -> s { currentCradle = Just (ps', cradle, Nothing) })
```
fendor: This should definitely have an option to cache the `ComponentOptions` here. You can obtain the ComponentOptions by splitting the call to `BIOS.initializeFlagsWithCradleWithMessage` into its implementation https://github.com/mpickering/hie-bios/blob/master/src/HIE/Bios/Ghc/Api.hs#L79. Ideally, we would expose that somehow in the API, maybe make an issue upstream?

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
fendor: However, this might have other impliciations, e.g. what happens when you try to format a file before it has loaded? I suppose that should not be allowed?

Avi-D-coder: Blocking is limited since, the cradle needs to be initialized before we attempt `BIOS.getCompilerOptions`. 

If the cradle is not loaded we may format the document incorrectly. While we could fail to format before the cradle is loaded the current behavior allows you to format even when hie-bios is unable to load the cradle. It's up to the user to decided if they can format without the cradle options. I still get strange `unable to satisfy z-package`  errors when I load a project from a particular broken states, so I don't want my formatter to be depend on having a loaded cradle.

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
Avi-D-coder: That does seem like better approach

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
fendor: > Blocking is limited since, the cradle needs to be initialized before we attempt BIOS.getCompilerOptions.

When you use `hie.yaml`, the function `BIOS.getCompilerOptions` will compile all dependencies of the project if they havent been built before. So, it might block for multiple minutes.

> It's up to the user to decided if they can format without the cradle options. I still get strange unable to satisfy z-package errors when I load a project from a particular broken states, so I don't want my formatter to be depend on having a loaded cradle.

When this happens, the compiler options may be just as broken. In general the problem is that formatting depends very much on compiler options that brittany and ormolu and floskell can not obtain by themselves. I think, failing to format if we cant load the project is the more expected behaviour, since we can only format the module correctly if it compiles.

> I still get strange unable to satisfy z-package errors 

This happens because `stack repl` fails when you have internal libraries. When you use cabal, this is not a problem, iirc.

Avi-D-coder: Can a cradle end up in `GhcModuleCache` prior to it's initialization? My thinking was since `getCompilerOptions` was already called in initialization, the second call will not have to build dependencies, except in pathological cases.

"splitting the call to `BIOS.initializeFlagsWithCradleWithMessage` into its implementation" will replace this any way.

fendor: You are right that should not happen. But it may if someone changes the configuration in between calls, e.g. modifies `package.yaml` or `*.cabal`.

Avi-D-coder: > When this happens, the compiler options may be just as broken. In general the problem is that formatting depends very much on compiler options that brittany and ormolu and floskell can not obtain by themselves. I think, failing to format if we cant load the project is the more expected behaviour, since we can only format the module correctly if it compiles.

Each formatter can handle this differently. Ormolu always enables all but a few extensions, so most of the time it's behavior is correct, the user can always just undo it, if it's incorrect. Working without a valid cradle especially after I enable ranged formats for Ormolu, is a behavior I want to preserve.

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -100,21 +100,36 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
-data CachedCradle = CachedCradle BIOS.Cradle HscEnv
+getComponentOptions
+  :: HasGhcModuleCache m => FilePath -> m (Maybe BIOS.ComponentOptions)
+getComponentOptions fp = do
+  mc <- getModuleCache
+  case currentCradle mc of
+    Just (dirs, _, co) | any (`isPrefixOf` fp) dirs -> return $ Just co
+    _ -> case T.match (cradleCache mc) (B.pack fp) of
+      Just (_, CachedCradle _ _ co, _) -> return $ Just co
+      _ -> return Nothing
```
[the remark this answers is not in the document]
Avi-D-coder: That does look attractive, now that it's no longer handling caching.

fendor:
@Avi-D-coder Shall we wait for an hie-bios release before merging this?

Avi-D-coder:
@fendor When is the next hie-bios release planed? I would like to get this pr in today, so I can send in ormolu 0.0.3.1 and range formats on top of it and only have to rebase once. If hie-bios needs more time I would prefer to temporarily add the `initializeFlagsWithCradleWithMessage` and remove it when hie-bios is updated.

fendor:
I would plan to update hie-bios before the HIE release, if possible. However, adressing the remaining issue, then merging and creating an issue is fine for me

alanz:
> When is the next hie-bios release planed? 

Probably the weekend of 1 Feb

hie-plugin-api/Haskell/Ide/Engine/ModuleCache.hs
```diff
@@ -166,13 +167,14 @@ loadCradle publishDiagnostics iniDynFlags (NewCradle fp) def action = do
   initialiseCradle :: (MonadIde m, HasGhcModuleCache m, GHC.GhcMonad m)
                   => Bios.Cradle -> (Progress -> IO ()) -> m (IdeResult a)
   initialiseCradle cradle f = do
-    res <- Bios.initializeFlagsWithCradleWithMessage (Just (toMessager f)) fp cradle
-    case res of
-      Bios.CradleNone ->
+    co <- liftIO $ Bios.getCompilerOptions fp cradle
```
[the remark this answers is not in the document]
Avi-D-coder: I went with `copts`

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -99,22 +99,42 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
 
 data GhcModuleCache = GhcModuleCache
   { cradleCache :: !(T.Trie CachedCradle)
               -- ^ map from FilePath to cradles
+              -- May not include currentCradle
   , uriCaches  :: !UriCaches
-  , currentCradle :: Maybe ([FilePath], BIOS.Cradle)
+  , currentCradle :: Maybe ([FilePath], BIOS.Cradle, BIOS.ComponentOptions)
               -- ^ The current cradle and which FilePath's it is
               -- responsible for
```
fendor: Extend comments, please

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -99,22 +99,42 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
+  }
 
 instance Show CachedCradle where
-  show (CachedCradle x _) = show x
+  show (CachedCradle x _ _) = show x
 
 data GhcModuleCache = GhcModuleCache
   { cradleCache :: !(T.Trie CachedCradle)
               -- ^ map from FilePath to cradles
+              -- May not include currentCradle
```
fendor: Why not? The cradle cache should also contain current cradle, no?

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -99,22 +99,42 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
+  return $ lookupInCache fp gmc (const Just) (Just . compOpts) Nothing
+
+lookupInCache
+  :: FilePath
+  -> GhcModuleCache
+  -> (BIOS.Cradle -> BIOS.ComponentOptions -> a)
+  -> (CachedCradle -> a)
+  -> a
+  -> a
+lookupInCache fp gmc cur cached def = case currentCradle gmc of
```
fendor: Please document with a sentence or two

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -99,22 +99,42 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
-  case currentCradle gmc of
-    Just (dirs, _c) | (any (\d -> d `isPrefixOf` fp) dirs) -> ReuseCradle
-    _ -> case T.match (cradleCache gmc) (B.pack fp) of
-           Just (_k, c, _suf) -> LoadCradle c
-           Nothing  -> NewCradle fp
-
-data CachedCradle = CachedCradle BIOS.Cradle HscEnv
+  lookupInCache fp gmc (const $ const ReuseCradle) LoadCradle $ NewCradle fp
+
+lookupComponentOptions
```
fendor: Also add a comment.

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -99,22 +99,42 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
+  Just (dirs, c, co) | any (`isPrefixOf` fp) dirs -> cur c co
+  _ -> case T.match (cradleCache gmc) (B.pack fp) of
+    Just (_k, c, _suf) -> cached c
+    Nothing            -> def
+
+data CachedCradle = CachedCradle
+  { ccradle :: BIOS.Cradle
+  , hscEnv :: HscEnv
+  , compOpts :: BIOS.ComponentOptions
+  }
```
fendor: If you are already at it, could you extend documentation for this, too?

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
Avi-D-coder: No idea, I added this as a warning, because I had expected it to be in the cache.

Avi-D-coder:
Can someone who uses cabal tell me why its not trying `ormolu-0.0.3.1`?
```
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: haskell-ide-engine-1.0.0.0 (user goal)
[__1] next goal: ormolu (dependency of haskell-ide-engine)
[__1] rejecting: ormolu-0.0.3.0, ormolu-0.0.2.0, ormolu-0.0.1.0 (conflict:
haskell-ide-engine => ormolu>=0.0.3.1)
[__1] fail (backjumping, conflict set: haskell-ide-engine, ormolu)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: haskell-ide-engine, ormolu
```

fendor:
@Avi-D-coder maybe you need to update your hackage index, e.g. `cabal v2-update`?

Avi-D-coder:
@fendor no luck both on circleci and locally

```
cabal v2-clean
cabal v2-update
cabal v2-configure 
```

hie-plugin-api/Haskell/Ide/Engine/GhcModuleCache.hs
```diff
@@ -99,24 +99,49 @@ data LookupCradleResult = ReuseCradle | LoadCradle CachedCradle | NewCradle File
...
+  return $ lookupInCache fp gmc (const Just) (Just . compOpts) Nothing
+
+lookupInCache
+  :: FilePath
+  -> GhcModuleCache
+  -- | Called when file is in the current cradle
+  -> (BIOS.Cradle -> BIOS.ComponentOptions -> a)
+  -- | Called when file is a member of a cached cradle
+  -> (CachedCradle -> a)
+  -- | Default value to return is a cradle is not found
```
fendor:
```suggestion
  -- | Default value to return if a cradle is not found
```

typo

fendor:
Last remark :)

Avi-D-coder:
Anyone else have a objection before CI finishes and I merge?

Avi-D-coder:
All the azure errors are unrelated.

Issue #1585, opened 2020-01-21
Overhaul test-suites


fendor:
A single test-run takes a huge amount of time and, by now, we have a huge array of VM tests.
Currently, we have the resulting matrix of test-runs:

Tool\GHC|8.4.2|8.4.3|8.4.4|8.6.1|8.6.2|8.6.3|8.6.4|8.6.5|8.8.1
-------------|-------|-------|------|------|-------|-------|------|-------|------
Cabal| :x: | :x: | :heavy_check_mark: | :x: | :x: | :x: | :x: | :heavy_check_mark: | :heavy_check_mark: 
Stack| :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: 

Resulting in *12* VMs to just build in every configuration, making it *36* VMs for every supported operating system: `Linux` | `MacOs` | `Windows`.

Tests take roughly [20 minutes](https://dev.azure.com/hvriedel/hvriedel/_build/results?buildId=53&view=logs&jobId=64f90618-c831-5d6e-048c-9ed6a9978957&j=0b3c19db-1883-5045-2e73-dfcdc072cf72&t=30056a90-ec38-5fcc-e0a9-b417dc28a579) or even more (on my machine 26 min) to run. 

To make the matter worse, we should have tests where *only* either `cabal` or `stack` is installed and both.

What I want to discuss in this post is an overhaul of the test system with the following points:
* Functional tests `import suggestions` are incredibly slow
  * Demote them to `unit-tests`
* Prefer `direct` cradle over stack cradle
  * On my system a call to `stack repl` takes roughly `1.5s`, 
    * I estimate we have 60 calls to `stack repl`, replacing this with a `direct` cradle will reduce run-time by a minute.
* Add tests for loading project context with `stack` and `cabal`, similar to cabal-helper.
  * Since `hie-bios` abstracts the build-tools, we dont have to execute all tests for each build-tool as long as we test that the build-tools initialize the context correctly and provide compilation arguments.

PR #1591, opened 2020-01-22
Demote HsImport func-tests to unit-test

fendor:
Previously, we had long running func-tests for HsImport.
However, after the first code-action tests, we actually are
testing the same functionality over and over again.
Moreover, we dont have to test that HsImport works as intended
since they have a huge set of tests, anyways.

Now, we only test that our implementation of the plugin
truly maps correctly to HsImport, accelerating execution time
of func-tests as a side-effect. Hooray!

Part of #1585

### DIFF
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -200,6 +200,7 @@ test-suite unit-test
                        GhcModPluginSpec
                        -- HaRePluginSpec
                        HooglePluginSpec
+                       HsImportSpec
                        JsonSpec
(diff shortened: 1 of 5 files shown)


fendor:
First results regarding test execution time:
Old: 19.10m (948e792)
New: 4.50m (2c33e49)
=> 4x speed up (or 25% of the previous time)

fendor:
The speedup is very relative, depending on how much needs to be rebuild, but there is a consistent speed up, at least it takes only 75% of the previous test-runs.

PR #1594, opened 2020-01-23
Drop GHC version 8.6.1, 8.6.2 and 8.6.3 (#1592)

### LINKED ISSUE #1592: Drop support for GHC 8.6.3
Afaict, it is broken for windows #1008, nobody should be using it.
Removing support for GHC 8.6.3 reduces pressure on CI

fendor:
Version 8.6.3 is broken on windows, so should not be used.
All other minor versions do not have a stackage LTS, so
stack users should not be using them, either, since they
can upgrade within an LTS with ease.

Closes #1592


alanz:
Given 6 month GHC major release cycle,  I think it makes sense to drop the earlier versions for a particular release.  Especially as the later ones in a series tend to be important bug-fixes.

So I am happy to merge this if everyone else is happy.

.azure/macos-stack.yml
```diff
@@ -13,8 +13,6 @@ jobs:
         YAML_FILE: stack-8.6.5.yaml
       stack-8.6.4:
         YAML_FILE: stack-8.6.4.yaml
```
jneira: Do we keep 8.6.4 in macos for some reason?

fendor:
wupps, sorry, didnt wait for @bubba, if you disagree we can revert this!

PR #1597, opened 2020-01-24
Fix error message parsing to import types

fendor:
Fix error message parsing to enable HsImport plugin to import Types.
Moreover, fix generation of Code Actions to avoid cases
where the same Code Action was displayed multiple times.

Overall tests for HsImport actions are lacking. This problem flew under the radar since I initially wrote the feature of import-lists.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/HsImport.hs
+++ b/src/Haskell/Ide/Engine/Plugin/HsImport.hs
@@ -339,7 +339,7 @@ codeActionProvider plId docId _ context = do
   importModuleAction
     :: SearchStyle -> ImportDiagnostic -> ModuleName -> SymbolName -> IdeM [J.CodeAction]
   importModuleAction searchStyle impDiagnostic moduleName symbolTerm =
-    catMaybes <$> sequenceA codeActions
+    catMaybes <$> sequenceA importListActions
     where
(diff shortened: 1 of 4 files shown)


src/Haskell/Ide/Engine/Plugin/HsImport.hs
```diff
@@ -391,6 +399,40 @@ codeActionProvider plId docId _ context = do
...
+      -- e.g. querying for 'Sum', possible results include:
+      --
+      -- * "data Sum f g a"
+      -- * "Sum :: a -> Sum a"
+      --
+      -- Depending on whether we are looking for a Type or Value,
+      -- we want to generate different Code Actions.
+      -- Important to generate the correct Code Actions.
+      --
+      -- >>> typeDeclaration "data Sum f g a"
```
bubba: Is this a doctest? Might be nice to turn it on in CI in some form

bubba: Also we should probably stick in a test case somewhere

fendor: It is intended as an usage example, but I think it is a doctest, too!
HsImport is vastly untested, unfortunately and quite a huge plugin. We need a way to extend these.

PR #1601, opened 2020-01-25
Update HIE to use latest hie-bios

fendor:
Closes https://github.com/haskell/haskell-ide-engine/issues/1490


fendor:
Weird single test-failure, but other than that it looks good.

fendor:
Weird test-failure that I can not explain, but it builds fine.

alanz:
Documenting the IRC discussion.  @fendor and I agreed to mark the failing test `pendingWith`, create an issue for it, then land this. so it can go into the monthly release. 

Nothing like living dangerously.

PR #1602, opened 2020-01-26
Ormolu range format support

Avi-D-coder:
This is a bit of a hack but it works pretty great.
We format lines of equal or increasing indention, not precise ranges, and there are a few documented edge cases.

### DIFF
--- a/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
+++ b/src/Haskell/Ide/Engine/Plugin/Ormolu.hs
@@ -20,9 +20,13 @@ import           Data.Char
 import           Data.List
 import           Data.Maybe
 import qualified Data.Text                     as T
+import           GHC
 import           Ormolu
 import           Haskell.Ide.Engine.PluginUtils
+import           Haskell.Ide.Engine.Support.HieExtras
 import           HIE.Bios.Types
+import qualified DynFlags                      as D
+import qualified EnumSet                       as S
 #endif
 
 ormoluDescriptor :: PluginId -> PluginDescriptor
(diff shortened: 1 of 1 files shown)


src/Haskell/Ide/Engine/Plugin/Ormolu.hs
```diff
@@ -34,24 +40,73 @@ ormoluDescriptor plId = PluginDescriptor
...
+  case typ of
+    FormatText -> ret (fullRange contents) <$> fmt contents
+    FormatRange r ->
+      let
+        txt = T.lines $ extractRange r contents
+        lineRange (Range (Position sl _) (Position el _)) =
+          Range (Position sl 0) $ Position el $ T.length $ last txt
+        -- Pragmas will not be picked up in a non standard location,
+        -- or when range starts on a Pragma
+        extPragmas = takeWhile ("{-#" `T.isPrefixOf`)
```
alanz: Given we are working in the context of a module that is likely to already be loaded by GHC, in principle we should be able to query these directly.

alanz: And if we put this behaviour into a utility, we can either get the pragmas from the module `DynFlags`, or do this parsing as a fallback. And the GHC API already has a function to do that.

Avi-D-coder: I didn't see that I could show extensions from the `EnumSet` in `DynFlags`, I mistakenly thought I was going to have to test if each extension was set, to construct the list.

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
```diff
@@ -34,24 +44,69 @@ ormoluDescriptor plId = PluginDescriptor
...
-
-        case result of
-          Left  err -> return $ IdeResultFail (IdeError PluginError (T.pack $ "ormoluCmd: " ++ show err) Null)
-          Right new -> return $ IdeResultOk [TextEdit (fullRange _contents) new]
-  where
-    exop s =
-      "-X" `isPrefixOf` s
-      || "-fplugin=" `isPrefixOf` s
-      || "-pgmF=" `isPrefixOf` s
+{-# LANGUAGE BlockArguments #-}
```
alanz: Do we really need this extension? It just adds cognitive overload, people reading the code have to go off and learn what it does before continuing.

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
```diff
@@ -34,24 +44,69 @@ ormoluDescriptor plId = PluginDescriptor
...
+        mStrip = case txt of
+          (l : _) ->
+            let ws = fst $ T.span isSpace l
+            in  (,) ws . T.unlines <$> traverse (T.stripPrefix ws) txt
+          _ -> Nothing
+        err = return $ IdeResultFail
+          (IdeError
+            PluginError
+            (T.pack
+              "You must format a whole block of code. Ormolu does not support arbitrary ranges."
```
alanz: What user experience does this result in? A pop-up window?  This is likely to be a fairly common experience, so should be non-threatening, just a notice.

Another possible approach is to find the enclosing block that *can* be formatted, and process that.  What does the brittany plugin do in this case?  I wonder if that logic should be captured in a library, for use in all RangeFormatting requests, so we have a standard behaviour across hie.

Avi-D-coder: I use nvim and coc so I don't know how it looks to most users. Brittany will often produce an error, in similar situations. The point of this line is just to provide a more helpful error than what we would usually get `unable to parse`.

alanz: I am happy with the message, and its contents. I just want to make sure the end user experience makes sense. But it can be tweaked easily enough, so its not a blocker.

src/Haskell/Ide/Engine/Plugin/Ormolu.hs
Avi-D-coder: Oops we don't even need it anymore

Avi-D-coder:
If there are no more objections I'm going to merge this.

Avi-D-coder:
There is one more edge, reintroduced in the switch to `DynFlags` case I found today.

Avi-D-coder:
I've been using this version today, no issues so far, so I'm going to merge.

PR #1606, opened 2020-01-26
Add a cabal.project workaround for https://github.com/haskell/haddock/issues/1123

bubba:


bubba:
This doesn't work with 8.4 or 8.6. Best just to put this in cabal.project.local if you're building with ghc-8.8.2 for now

PR #1609, opened 2020-01-27
Azure update macos and try to fix windows+stack builds

jneira:


jneira:
He, i want to let them as reliable and stable as possible before taking a rest.
A source of errors in my repo is not enough disk space in C: for windows but fortunately they are about to increase its capacity: https://github.com/actions/virtual-environments/pull/309

jneira:
We are using 1 Gb of cache per job on average so roughly  a total of 29 Gb.
I hope they will not limit the cache size :neutral_face:

jneira:
Finally all jobs green.. merging! :smile:

PR #1610, opened 2020-01-27
Deduplicate main for hie/hie-wrapper

Gurkenglas:
Not sure how to merge the CPP'd imports.
Note that hie-wrapper now partakes of the hSetBuffer stderr LineBuffering
  and the "hie-bios".

### DIFF
--- a/src/Haskell/Ide/Engine/Options.hs
+++ b/src/Haskell/Ide/Engine/Options.hs
@@ -1,9 +1,6 @@
 {-# LANGUAGE CPP #-}
 module Haskell.Ide.Engine.Options where
 
-#if __GLASGOW_HASKELL__ < 804
-import           Data.Semigroup
-#endif
 #if __GLASGOW_HASKELL__ < 808
 import           Data.Monoid                           ((<>))
 #endif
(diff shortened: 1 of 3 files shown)


src/Haskell/Ide/Engine/Options.hs
```diff
@@ -15,6 +29,27 @@ data GlobalOpts = GlobalOpts
   , optFiles         :: [FilePath]
   } deriving (Show)
 
+initApp :: String -> IO GlobalOpts
```
fendor: Since this is now happening out of context, it would be nice if there was some documentation what this function does, e.g. initialises logger, searches for some project root, etc...

src/Haskell/Ide/Engine/Options.hs
```diff
@@ -1,7 +1,21 @@
 {-# LANGUAGE CPP #-}
 module Haskell.Ide.Engine.Options where
 
+#if __GLASGOW_HASKELL__ < 804
+import           Data.Semigroup
+#endif
+#if __GLASGOW_HASKELL__ < 808
+import           Data.Monoid                           ((<>))
+#endif
```
fendor: We could move these to `Compat.hs` but then we would need to import that everywhere else. Not sure if this is better.
Btw, we are not supporting Versions lower than 8.4.2, so I think the first CPP block can be omitted.

src/Haskell/Ide/Engine/Options.hs
Gurkenglas: Maybe instead reestablish context, such as by saying in a comment that this is the common prefix of hie and hie-wrapper?

fendor: Yeah, that makes sense, too. Still, what it roughly does to set-up the application would be nice as well.

Issue #1416, opened 2019-10-18
Should ghcide be integrated?


Avi-D-coder:
ghcide's errors reporting, hover, goto definition is far more fault tolerant. It still works when started in a file with errors. It is also very snappy. I would argue that hie should fully integrate ghcide.

What do other people think? Are there any fundamental problems with relying on it?

mpickering:
The plan is to merge in the hie-bios branch then look into integrating ghcide.

alanz:
See also #1525, as a background info issue.

The current plan is to work on this at [Bristol Haskell Hackathon 2020](https://mpickering.github.io/bristol2020.html), kindly arranged my @mpickering .

alanz:
Our preliminary discussion at the Bristol Hackathon is that we will create a new umbrella repo, aiming to be called `github.com/haskell/ide`, which will make use of libraries from ghcide, haskell-ide-engine, and whatever else makes sense.  It should be a quite small application, forming an integration point.

The main interface to ghc will be via ghcide, which has a demand-driven shake based build system, also using hie-bios at the bottom.

The initial work will be about making sure ghcide supports multi-modules, and that hie-bios has good implicit-project support, probably making use of cabal-helper.

The plugin format will be harmonised, the goal being to be able to create a main haskell file which imports dependencies from hackage, puts them in a structure and launches the LSP server with them, that just works.  This means  for example that formatters can be pluggable, and you only need to build the one you intend using. It should also allow configuring in private plugins.

hasufell:
> which will make use of libraries from ghcide, haskell-ide-engine, and whatever else makes sense. It should be a quite small application, forming an integration point.

What is the proposed workflow of the joint project? What are the project goals? Will legacy code be dropped? Will there be feature regression tests?

IMO, a big part of tooling problems arise from the fact that codebases quickly become unmaintainable and inaccessible to contributors, because they are inherently difficult problems. That's even more a reason to be particularly worried about maintenance and project policies.

> probably making use of cabal-helper

Isn't cabal-helper one of the more problematic parts of the toolchain? It compiles things at runtime in my home directory somewhere and regularly fails.

alanz:
https://github.com/haskell/ide

Anyone needing/wanting to be a collaborator, shout

serras:
I am willing to help!

alanz:
> I am willing to help!

And it seems I need permissions tweaked on the repo to be able to invite collaborators. Will get it sorted tomorrow.

[...]

PR #1611, opened 2020-01-27
Put IdeResult in terms of Either, add ideError

Gurkenglas:
Let's try this again.


Avi-D-coder:
@Gurkenglas What is the motive for this change?

Gurkenglas:
You have answered your question :3

jneira:
Thanks for the pr, i think the intent is good and i always like the prs that removes locs (the actual new type has its own advantages though). 
However, as you surely already knows, we just have start a process of integration with ghcide so it is the actual main focus. 
We will continue to fix bugs but we will prefer that the new code or internal refactorings go in that direction (see https://github.com/haskell/haskell-ide-engine/issues/1416#issuecomment-579010923).

PR #1615, opened 2020-01-29
Add dev target to stack install.hs

### LINKED ISSUE #1565: Add a new target `nightly`/`dev` to install script
* To build with stack and the default `stack-yaml`

jneira:
To build hie with the default `stack.yaml` file
I've not added it to documentation cause its user will be developers
Closes #1565


jneira:
Merging, just before that was copied to hie 😞

PR #1618, opened 2020-01-30
[WIP] stylish-haskell plugin

lukasz-golebiewski:


stack.yaml
```diff
@@ -26,9 +26,12 @@ extra-deps:
 - hie-bios-0.3.2
 - hlint-2.2.8
 - hsimport-0.11.0
+- HsYAML-0.2.1.0@sha256:e4677daeba57f7a1e9a709a1f3022fe937336c91513e893166bd1f023f530d68,5311
```
Avi-D-coder:
```suggestion
- HsYAML-0.2.1.0
```

stack.yaml
```diff
@@ -26,9 +26,12 @@ extra-deps:
...
+- HsYAML-aeson-0.2.0.0@sha256:04796abfc01cffded83f37a10e6edba4f0c0a15d45bef44fc5bb4313d9c87757,1791
```
Avi-D-coder:
```suggestion
- HsYAML-aeson-0.2.0.0
```

lukasz-golebiewski:
I see that jobs for older stack/cabal stacks are failing because of absence of `stylish-haskell-0.10.0.0`. Is it possible to add a plugin only starting from the currently latest stack/cabal version?

Avi-D-coder:
@lukasz-golebiewski Add `- stylish-haskell-0.10.0.0` to each `stack-*.yaml` and a version bound in `haskell-ide-engine.cabal`.

stack.yaml
```diff
@@ -26,9 +26,12 @@ extra-deps:
 - hie-bios-0.3.2
 - hlint-2.2.8
 - hsimport-0.11.0
+- HsYAML-0.2.1.0
+- HsYAML-aeson-0.2.0.0
```
Avi-D-coder: You'll likely also need to add these to each `stack.yaml` and to `haskell-ide-engine.cabal`

lukasz-golebiewski: This doesn't seem to be working :-(

jneira: I have to note that HsYAML is using GPL and the project licensing would be affected (not sure if it already is using libs with gpl)

jneira: Personally i am fine with linking a gpl lib but the project will have two yaml libs and maybe we should use only one (not in this pr)

Avi-D-coder:
@lukasz-golebiewski There are a few more version conflicts that need to be explicitly specified

stack.yaml
Avi-D-coder: I'm also okay with GPL, however stylish-haskell just made the switch, we could ask them to switch back https://github.com/jaspervdj/stylish-haskell/commit/498d676c1e06af966c814e01967fc3ece1fd9225

stack-8.4.4.yaml
```diff
@@ -12,7 +12,7 @@ extra-deps:
 - bifunctors-5.5.6
 - brittany-0.12.1.1
 - bytestring-trie-0.2.5.0
-- Cabal-3.0.0.0@sha256:1ba37b8d80e89213b17db7b8b9ea0108da55ca65f8c0cbb7433881a284c5cf67,26027
+- Cabal-2.4.1.0
```
Avi-D-coder: This does not seem correct?

lukasz-golebiewski: Stack recommended to set cabal version explicitly in https://circleci.com/gh/haskell/haskell-ide-engine/13311
If this is incorrect then I don't think I know how to proceed with this

fendor: Does this compile? Because, iirc, stack snapshots are bundled with `cabal-install` and `Cabal` versions, iirc, setting a custom Cabal version for a snapshot may not succeed. If it compiles, everything is fine.

lukasz-golebiewski: Not yet, I'm slowly bisecting towards the cabal version satisfying every dependency

stack.yaml
fendor: I am not sure "fine" is the right wording, since we are not GPL, we do not really have the rights to using it, right? On the other hand, I doubt, too, that someone will sue us.

jaspervdj: Hey, I wasn't aware that HsYAML was GPL-licensed.  I don't think it's a problem -- it's fine for a BSD3 application to depend on a GPL-licensed library.  It does not affect the code of `haskell-ide-engine` or `stylish-haskell`, those remain BSD3-licensed.  However, the compiled binary must adhere to the GPL license (which seems fine in this case).  This is similar to e.g. the fact that GHC links against `libgmp` by default.

But, if this worries you, I am happy to add a flag to `stylish-haskell` that makes it pick the C-based `yaml` as a dependency rather than `HsYAML`?

jneira: Afaiu we can use it without changing our licenses files but the license for hie artifacts linking HsYaml becomes auto gpl. So anybody could issue to someone who doesnt respect the gpl resdistributing hie.

fendor: IANAL, if you argue that this is indeed just fine, I have nothing to say against it :)

fendor: Ah. Assuming this is correct, this sounds totally reasonable.

stack-8.4.4.yaml
lukasz-golebiewski: And back to stack suggesting `Cabal-3.0.0.0`. Looks like stylish `0.10.0.0` can't be used here and below because of conflicting dependencies :-(

Ailrun:
Is there any news?

PR #1619, opened 2020-01-30
Azure fix win cabal 8.4.4

jneira:
Moving cabal root dir from `%APPDATA%\cabal` to `D:\cabal` and free space in C:
They are about to increase its capacity but this could be kept afterwards


jneira:
windows stack-8.4.2 does not even start to build dependencies. I did not touch it so it should be caused for something else

Issue #1261, opened 2019-05-19
Failed to build HIE 0.9.0.0 with cabal 2.4.1.0


unclechu:
```bash
stack ./install.hs cabal-build-all
```
```
Progress 0/13ghc-pkg: cannot find package Cabal-2.4.1.0
ghc-pkg: cannot find package clock-0.7.2
Progress 1/13: cabal-doctest-1.0.6, clock-0.7.2, hashable-1.2.7.0, js-flot-0.8.3ghc-pkg: cannot find package js-flot-0.8.3
Progress 2/13: cabal-doctest-1.0.6, extra-1.6.14, hashable-1.2.7.0, js-flot-0.8.3ghc-pkg: cannot find package hashable-1.2.7.0
Progress 3/13: cabal-doctest-1.0.6, extra-1.6.14, hashable-1.2.7.0, js-jquery-3.3.1ghc-pkg: cannot find package cabal-doctest-1.0.6
Progress 4/13: cabal-doctest-1.0.6, extra-1.6.14, js-jquery-3.3.1, primitive-0.6.4.0ghc-pkg: cannot find package extra-1.6.14
Progress 5/13: extra-1.6.14, heaps-0.3.6, js-jquery-3.3.1, primitive-0.6.4.0ghc-pkg: cannot find package js-jquery-3.3.1
Progress 6/13: heaps-0.3.6, js-jquery-3.3.1, primitive-0.6.4.0, random-1.1ghc-pkg: cannot find package primitive-0.6.4.0
Progress 7/13: heaps-0.3.6, primitive-0.6.4.0, random-1.1, unordered-containers-0.2.9.0ghc-pkg: cannot find package random-1.1
Progress 8/13: heaps-0.3.6, random-1.1, unordered-containers-0.2.9.0, utf8-string-1.0.1.1ghc-pkg: cannot find package unordered-containers-0.2.9.0
Progress 9/13: heaps-0.3.6, unordered-containers-0.2.9.0, utf8-string-1.0.1.1ghc-pkg: cannot find package utf8-string-1.0.1.1
# git (for submodules)
Synchronizing submodule url for 'submodules/HaRe'
Synchronizing submodule url for 'submodules/HaRe/submodules/cabal-helper'
Synchronizing submodule url for 'submodules/HaRe/submodules/ghc-mod'
Synchronizing submodule url for 'submodules/brittany'
Synchronizing submodule url for 'submodules/cabal-helper'
[...]

PR #1622, opened 2020-02-01
Bump resolvers, hlint, brittany

alanz:
cabal now index state 2020-01-31T21:11:24Z
GHC 8.8.2 is nightly-2020-01-31
GHC 8.6.5 is lts-14.22

hlint is 2.2.9
brittany is 0.12.1.1

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: nightly-2019-09-21 # Last GHC 8.6.5
+resolver: nightly-2020-01-31
 packages:
 - .
 - hie-plugin-api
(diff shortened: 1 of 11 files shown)


alanz:
@jneira I had to flush the cache on CircleCI for this to pass. Could the azure failures be something like that?

jneira:
Mmm some linux failed with
```
Failures:

  test/dispatcher/Main.hs:297:7: 
  1) functional dispatch instantly responds to failed modules with no cache with the default
       expected: ("req7",Just [])
        but got: ("req8",Nothing)

  To rerun use: --match "/functional dispatch/instantly responds to failed modules with no cache with the default/"

Randomized with seed 429502550
```
i think i saw that error before, i've just rerun the failing jobs to see if it is transient

jneira:
Linux errors are gone, only left:
* build error with stack+stack.yaml: i don't see the concrete error 🤔 
* error running `cabal-hie-install latest` in windows: the second one

i would try both locally

jneira:
`cabal-hie-install latest` was my fault fix in #1627

alanz:
Ok. Heads up, I am rebasing this PR on the hie-bios update one, and some things are changing. We will see how it ends up.

Aside, I had to blow away my cabal store and rebuild it, to get rid of weird errors about diamond deps.

stack.yaml
```diff
@@ -1,39 +1,34 @@
-resolver: nightly-2019-09-21 # Last GHC 8.6.5
+resolver: nightly-2020-01-31
```
jneira: We cant build hie with `ghc-8.2.2`, if we use it in `stack.yaml` we should remove it from azure ci.
I think the plan was keeping ghc-8.6.5 in `stack.yaml` until `ghc-8.8.3` lands.

alanz: Can we get azure to use `stack-8.6.5.yaml` instead? Which makes it clear

jneira: mmm, there already is a job that uses `stack-8.6.5.yaml`. We can remove the one that uses `stack.yaml`

jneira:
Now, builds with ghc-8.4.3 are failing with:
```
warp                             > [21 of 39] Compiling Network.Wai.Handler.Warp.Recv
warp                             > 
warp                             > /tmp/stack6092/warp-3.2.25/Network/Wai/Handler/Warp/Recv.hs:96:34: error:
warp                             >     • Couldn't match expected type ‘CInt’ with actual type ‘IO CInt’
warp                             >     • In the first argument of ‘receiveloop’, namely ‘sock'’
warp                             >       In the second argument of ‘(<$>)’, namely
warp                             >         ‘receiveloop sock' ptr size'’
warp                             >       In a stmt of a 'do' block:
warp                             >         fromIntegral <$> receiveloop sock' ptr size'
warp                             >    |
warp                             > 96 |     fromIntegral <$> receiveloop sock' ptr size'
warp                             >    |                                  ^^^^^
warp                             > 
warp                             > /tmp/stack6092/warp-3.2.25/Network/Wai/Handler/Warp/Recv.hs:103:43: error:
warp                             >     • Couldn't match expected type ‘CInt’ with actual type ‘IO CInt’
warp                             >     • In the first argument of ‘receiveloop’, namely ‘fd’
warp                             >       In the second argument of ‘(<$>)’, namely
warp                             >         ‘receiveloop fd buf (fromIntegral siz)’
warp                             >       In a stmt of a 'do' block:
warp                             >         n <- fromIntegral <$> receiveloop fd buf (fromIntegral siz)
warp                             >     |
warp                             > 103 |         n <- fromIntegral <$> receiveloop fd buf (fromIntegral siz)
warp                             >     |                              
```

jneira:
I am working to fix ghc-8.4.3 builds

jneira:
There is a failing unit test but ,as @fendor as noted , it was introduced by #1261 so i think we can merge this.

alanz:
@jneira thanks

PR #1627, opened 2020-02-02
Fix cabal-hie-install in windows azure ci

jneira:
I forgot to unpack the cache in the correct drive


alanz:
When you are happy with this branch, merge

jneira:
Finally i had to use the default `CABAL_DIR` location to make `cabal-hie-install` work, not sure the cause though

PR #1633, opened 2020-02-04
Azure: not trigger pr builds by path

jneira:
Prs modifying docs continued to trigger the azure builds


jneira:
Yeah i tried it but you have to set them in the pr section too

PR #1655, opened 2020-02-14
Fix failing tests in windows (reloaded)

jneira:
* Using a patched version of cabal-helper (until the next release)
  * It includes the remove of the private lib
* Azure windows agent have changed their drives space now `C:` has lot of more space than `D:`, so i've switched again cache to `C:`
* Some of unit failing tests had been fixed by #1654


jneira:
mmm, i've found submodules more suited to mid or long term forks due to the additional effort to setup and maintain them, i hope this is a  provisional workaround

Issue #1664, opened 2020-02-21
HIE build issues (macOS)


awave1:
Hi,

I'm getting some build errors (again) on macOS 10.15.3 (19D76). 

Here's my log output for `stack ./install.hs hie-8.6.5 --verbose`: 
* [https://gist.github.com/d8fd3fe5c58007baaebf74a29df3897a](https://gist.github.com/d8fd3fe5c58007baaebf74a29df3897a)

I tried re-building `hie` using `stack clean --full` and then `stack ./install.hs hie-8.6.5`, but still get the same error. 

It seems the build fails because of this:
```
hie-bios          > <command line>: can't load .so/.DLL for: libargp.dylib (dlopen(libargp.dylib, 5): image not found)
```

I installed `argp-stanalone` and added `-largp` to `LDFLAGS`, unfortunately that did not help.

Also, note that building a simple project with stack works fine, e.g.

```sh
stack new temp-project --resolver lts-14.27 && cd temp-project && stack build
```

Any ideas why it would be failing?

mouse07410:
MacOS Mojave 10.14.6. 
```
$ ghcup debug-info
Script version: 0.0.8

Script variables:
  GHC install location:     /Users/ur20980/.ghcup/ghc
  Binary install location:  /Users/ur20980/.ghcup/bin
  Tarball cache location:   /Users/ur20980/.ghcup/cache
  Downloader:               curl -L --fail -O <url>
  Script update url:        https://gitlab.haskell.org/haskell/ghcup/raw/master//ghcup
  GHC download baseurl:     https://downloads.haskell.org/~ghc
  Meta download url         https://gitlab.haskell.org/haskell/ghcup/raw/master//.download-urls
  Meta download format      1
  Meta version url          https://gitlab.haskell.org/haskell/ghcup/raw/master//.available-versions
  Meta version format       1

Detected system information:
  Architecture:   x86_64
  Distribution:   Darwin
  Distro alias:   darwin
  Distro version: 18.7.0
$ 
```

Attempt to build HIE-8.8.2 fails:
```sh
$ ./cabal-hie-install hie
. . . . .
# cabal (for hie-8.8.2)
# cabal (for hie-8.8.2)
Wrote tarball sdist to
/Users/ur20980/src/haskell-ide-engine/dist-newstyle/sdist/haskell-ide-engine-1.1.tar.gz
Wrote tarball sdist to
[...]

PR #1665, opened 2020-02-24
Improvements of install script and macos-installhs-cabal new azure job

jneira:
* Add job to test the build of install.hs with macos and cabal
* Improvements over install.hs:
  * Remove unused imports
  * Pass number of jobs and verbosity shake args to build tools
  * Use custom cabal.project-${ghcVersion} if exists
    * I added `cabal.project-ghc-8.8.2` that will be used by the script
  * Add comments with the possible resolvers to shake.yaml

I am not sure about the "pass number of jobs and verbosity shake args to build tools". It is a cheap way to let users control the number of jobs and verbosity but they are shared between the run of the script itself and the calls to build tools.

Fixes some warts described in #1664


jneira:
Finally i removed the sharing between the jobs arg of shake and the build tool: the output of shake with more than one job was interleaved and, in general, we will want different values of jobs for shake and stack/cabal frequently.

jneira:
My mistake was mix azure and install changes so the pr is bigger than necessary

jneira:
Well all ci is green (including install.hs jobs) and some changes are needed to make easier ghc-8.8.3 support, so i am gonna merge it

PR #1685, opened 2020-03-07
Fix typos in README. Add hint to speed up HIE compilation

mb720:
Hi!

I'd like to offer two improvements to the installation instructions for Arch Linux:

  * Typo fixes
  * Add hint to reduce the number of supported GHC versions thus speeding up HIE compilation


fendor:
Thank you for catching this!

PR #1692, opened 2020-03-11
Report progress as a percentage

expipiplus1:
Previously it was displayed as a value from 0% to 1%

As specified https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workDoneProgress

See also https://github.com/alanz/haskell-lsp/blob/1f85ef3c150dc8b2626a181c47d8635015e67ee5/src/Language/Haskell/LSP/Core.hs#L150-L154


expipiplus1:
Thanks!

fendor:
Thank you for this sensible pull request :)

PR #1694, opened 2020-03-12
Use sha instead revision for hsimport and other fixes

### LINKED ISSUE #1689: Unable to build on darwin 10.14.5: haskell-src-exts-1.22.0 from stack configuration does not match >=1.18.0 && <1.22
on commit 02815bb0:

```bash
$ stack ./install.hs hie-8.6.5
# stack (for check)
# git (for submodules)
# git (for submodules)
Options:
    Verbosity level: Info
# stack (for hie-8.6.5)

Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for hsimport-0.11.0:
    haskell-src-exts-1.22.0 from stack configuration does not match >=1.18.0 && <1.22  (latest matching version is 1.21.1)
needed due to haskell-ide-engine-1.2 -> hsimport-0.11.0

Some different approaches to resolving this:

  * Set 'allow-newer: true' in /Users/timpierson/.stack/config.yaml to ignore all version constraints and build anyway.

  * Recommended action: try adding the following to your extra-deps in /Users/timpierson/arity/haskell-ide-engine/stack-8.6.5.yaml:

- haskell-src-exts-1.21.1@sha256:11d18ec3f463185f81b7819376b532e3087f8192cffc629aac5c9eec88897b35,4541

Plan construction failed.

********************************************************************************
Building failed, Try running `stack clean` and restart the build
If this does not work, open an issue at
	https://github.com/haskell/haskell-ide-engine
********************************************************************************

  at want, called at src/Development/Shake/Internal/Args.hs:83:69 in shake-0.18.5-AT2wsKie7GmK5OPJZ5kRVD:Development.Shake.Internal.Args
* Depends on: hie-8.6.5
  at command, called at src/Stack.hs:80:3 in hie-install-0.8.1.0-7UOJzsmkF9x8NsRSgy2ktK:Stack
* Raised the exception:
Development.Shake.command, system command failed
Command line: stack --stack-yaml=stack-8.6.5.yaml install --verbosity=info
Exit code: 1
Stderr:
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for hsimport-0.11.0:
    haskell-src-exts-1.22.0 from stack configuration does not match >=1.18.0 && <1.22  (latest matching version is 1.21.1)
needed due to haskell-ide-engine-1.2 -> hsimport-0.11.0

Some different approaches to resolving this:

  * Set 'allow-newer: true' in /Users/timpierson/.stack/config.yaml to ignore all version constraints and build anyway.

  * Recommended action: try adding the following to your extra-deps in /Users/timpierson/arity/haskell-ide-engine/stack-8.6.5.yaml:

- haskell-src-exts-1.21.1@sha256:11d18ec3f463185f81b7819376b532e3087f8192cffc629aac5c9eec88897b35,4541

Plan construction failed.
```

```bash
$ stack --version
Version 2.1.3, Git revision 0fa51b9925decd937e4a993ad90cb686f88fa282 (7739 commits) x86_64 hpack-0.31.2
```

```bash
$ git status
HEAD detached at 02815bb0
```

jneira:
* `@revn` way to specify a revision in `stack.yaml` files seems to be not reliable: it is using the package version without the revision under some circumstances. It is possible that wipping out the entire `$STACK_ROOT` could fix it but is is not desirable
* we have [tested ](https://github.com/haskell/haskell-ide-engine/issues/1689#issuecomment-597922927) that using pantry sha seems to make stack use the patched version, although it is somewhat ugly
* closes #1689

### DIFF
--- a/.azure/macos-installhs-cabal.yml
+++ b/.azure/macos-installhs-cabal.yml
@@ -21,7 +21,9 @@ jobs:
     displayName: "Unpack cache"
     condition: eq(variables.CACHE_RESTORED, 'true')
   - bash: |
-      brew install cabal-install
+      # TODO Install ghc-8.8.3 when adding stack-8.8.3.yaml 
+      brew install ghc@8.6
+      brew install cabal-install --ignore-dependencies ghc
       which cabal
       which ghc
       cabal update
--- a/.azure/macos-installhs-cabal.yml
+++ b/.azure/macos-installhs-cabal.yml
@@ -29,10 +29,12 @@ jobs:
       cabal update
     displayName: Install cabal and ghc
   - bash: |
+      export PATH="/usr/local/opt/ghc@8.6/bin:$PATH"
       source .azure/windows-cabal.bashrc
       cabal v2-run ./install.hs --project-file $PROJECT_FILE -- help
     displayName: Run help of `install.hs`
   - bash: |
+      export PATH="/usr/local/opt/ghc@8.6/bin:$PATH"
       source .azure/windows-cabal.bashrc
       cabal v2-run ./install.hs --project-file $PROJECT_FILE -- latest
     displayName: Run latest target of `install.hs`
--- a/haskell-ide-engine.cabal
+++ b/haskell-ide-engine.cabal
@@ -205,6 +205,7 @@ test-suite unit-test
                        LiquidSpec
                        OptionsSpec
                        PackagePluginSpec
+                       Paths_haskell_ide_engine
                        Spec
   -- Technically cabal-helper should be a 'run-tool-depends', but that doesn't exist yet
   build-tool-depends:  cabal-helper:cabal-helper-main, hspec-discover:hspec-discover
--- a/stack-8.4.2.yaml
+++ b/stack-8.4.2.yaml
@@ -36,7 +36,7 @@ extra-deps:
 - hie-bios-0.4.0
 - hlint-2.2.11
 - hoogle-5.0.17.15
-- hsimport-0.11.0@rev1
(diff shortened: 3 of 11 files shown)


jneira:
There is a unit test failing due to the version bump to `1.2`, weird that it was not triggered before :thinking:

jneira:
Fixed unit test with f6ba5ed (cc @gdziadkiewicz )

gdziadkiewicz:
Sorry for the problem :( and thanks for the cc, I appreciate the opportunity to learn how to do things better :)

jneira:
@gdziadkiewicz dont worry, your contributions have been great and that is a minor detail, only want to show you just in case you have not used the `Path_*` magic module before.

jneira:
The `macos-installhs-cabal` azure job failed cause brew installs `ghc-8.8.3` by default but we dont support it yet and there is no `stack-8.8.3.yaml`

fendor:
Nice catch, @jneira!

Infinisil:
Oh! I'm pretty sure that the problem here was that `@rev1` is the wrong syntax, it should be `@rev:1`, as documented [here](https://docs.haskellstack.org/en/stable/pantry/#hackage).

I looked at the pantry source code (which is what parses these fields), and it seems to just use the latest revision for `@rev1` (aka it couldn't parse it): https://github.com/commercialhaskell/pantry/blob/603fea5f4cb812b08edc74cd9e1178fa76309f55/src/Pantry/Types.hs#L769-L785

There's another instance of this in master, introduced in 0bdf8ccd9cfffff74bb2d3eb1950856ee1abed02. I'll make a PR to fix this

PR #1703, opened 2020-03-14
HsImport: Return codeactions importing from Internal modules last

### LINKED ISSUE #1702: HIE should deprioritize suggesting importing 'Internal' modules
Sometimes HIE prompts me to add an import (with hsimport I believe) and this is great!

It's a tiny papercut that the first result is sometimes for some library's `Internal` module, this means that to import the name from the non-internal module I have to navigate the code action list to select the right one instead of easily following the first selection (for which I have an editor shortcut defined)

It would be nice if the `Internal` imports were returned last in the list.

expipiplus1:
Fixes #1702


fendor:
Thank you, nice addition!

expipiplus1:
Thanks :)

Issue #1712, opened 2020-03-27
Build with GHC-8.10.1


mouse07410:
Fails. With "normal" Cabal config - fails to construct dependency tree:
```
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.10.1
$ ./cabal-hie-install hie-8.10.1
Warning: Unknown/unsupported 'ghc' version detected (Cabal 3.0.0.0 supports
'ghc' version < 8.10): /Users/ur20980/.ghcup/bin/ghc is version 8.10.1
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: fake-package-0 (user goal)
[__1] trying: base-4.14.0.0/installed-4.1... (dependency of fake-package)
[__2] trying: hie-install-0.8.1.0 (user goal)
[__3] trying: hie-install:-run-from-stack
[__4] next goal: cabal-install-parsers (dependency of hie-install
-run-from-stack)
[__4] rejecting: cabal-install-parsers-0.2, cabal-install-parsers-0.1
(conflict: base==4.14.0.0/installed-4.1..., cabal-install-parsers =>
base>=4.10 && <4.14)
[__4] fail (backjumping, conflict set: base, cabal-install-parsers,
hie-install, hie-install:run-from-stack)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: base, cabal-install-parsers,
hie-install, hie-install:run-from-stack, fake-package
Try running with --minimize-conflict-set to improve the error message.

```

If `--allow-newer: True` is set:
```
$ ./cabal-hie-install hie-8.10.1
Resolving dependencies...
Build profile: -w ghc-8.10.1 -O1
In order, the following will be built (use -v for more details):
 - binary-orphans-1.0.1 (lib) (requires build)
 - base-compat-0.11.1 (lib) (requires build)
 - base-orphans-0.8.2 (lib) (requires build)
 - base16-bytestring-0.1.1.6 (lib:base16-bytestring) (requires build)
 - clock-0.8 (lib) (requires build)
 - cryptohash-sha256-0.11.101.0 (lib) (requires build)
 - dlist-0.8.0.7 (lib) (requires build)
 - heaps-0.3.6.1 (lib) (requires build)
 - hsc2hs-0.68.7 (exe:hsc2hs) (requires build)
 - hashable-1.3.0.0 (lib) (requires build)
 - js-dgtable-0.5.2 (lib) (requires build)
[...]

PR #1715, opened 2020-03-28
Make install script use unix-time-0.4.7

### LINKED ISSUE #1714: ghc.exe: unable to load package "unix-time-0.4.5" unknown symbol  "_create_locale". ExitFailure 1
Actually ide engine worked fine for me, but then I screwed up some files. Now I'm trying `stack ./install.hs hie` and there's an error. So I wonder, is there a way to completely delete stack and ide engine from windows, and any mentions. Because when I had installed it for the first time, there were not any problems.

jneira:
* To avoid a bug in windows of previous unix-time versions
* It is throwed with stack executions within the `install.hs` script using `./install/shake.yaml` config
* Fixes #1714

### DIFF
--- a/install/shake.yaml
+++ b/install/shake.yaml
@@ -13,6 +13,7 @@ extra-deps:
 - shake-0.18.5
 # for resolvers with ghc < 8.6
 # - shake-0.17
+- unix-time-0.4.7
 
 
 nix:


mouse07410:
@jneira is there a reason why you're using resolver 13.19, rather than 15.6 or at least 14.27?

FWIW, I seem to be able to build the master branch successfully using resolver 15.6 in `install/shake.yaml`.

mouse07410:
@jneira I'm confused. 

On one hand, HIE must be built by the same GHC that it would serve, or there could/would be problems. 

On the other hand, you're freezing `install/shake.yaml` at the older GHC-8.6.5 LTS resolver - and not even the current GHC-8.6.5 LTS resolver (which is 14.27).  

I can understand why not `lts-15.6`. But why not `lts-14.27`? 

And how does it (or why it doesn't) impact building HIE with/for GHC-8.8.3 (e.g., on Mac and Linux)?

And yeah, I understand about 8.10.1. :-(

jneira:
I think maybe it is a little bit clear in https://github.com/haskell/haskell-ide-engine/issues/1712#issuecomment-608238790

mouse07410:
@jneira I hear you, thanks. 

I'd still suggest bringing the resolver to `lts-14.27` if you intend this script to stay with GHC-8.6.5,  or to `lts-15.6` if you're to move to GHC-8.8.3.

PR #1724, opened 2020-04-06
Bump resolvers

alanz:


jneira:
Maybe we could drop `cabal.project-8.8.2` and build for `ghc-8.8.2` with the default one if we update the index state.

alanz:
@jneira good idea. I actually tried that, but forgot to bump the index state.

Issue #1731, opened 2020-04-10
Build does not complete - floskell, ormolu, hlint failures


chexxor:
I got errors when building just now, too.

In Windows 10, I cloned this repo, then did `stack install`.

```
...
... (snip) ...
...
brittany          > Installing executable brittany in C:\sr\snapshots\0b20af29\bin
brittany          > Registering library for brittany-0.12.1.1..

--  While building package floskell-0.10.2 using:
      C:\sr\setup-exe-cache\x86_64-windows\Cabal-simple_Z6RU0evB_3.0.1.0_ghc-8.8.2.exe --builddir=.stack-work\dist\29cc6475 build --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1


--  While building package ormolu-0.0.3.1 using:
      C:\sr\setup-exe-cache\x86_64-windows\Cabal-simple_Z6RU0evB_3.0.1.0_ghc-8.8.2.exe --builddir=.stack-work\dist\29cc6475 build --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1


--  While building package hlint-2.2.11 using:
      C:\sr\setup-exe-cache\x86_64-windows\Cabal-simple_Z6RU0evB_3.0.1.0_ghc-8.8.2.exe --builddir=.stack-work\dist\29cc6475 build --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
Progress 8/14
```

Any ideas about what's going wrong here? I'm not an experienced Haskell developer.

chexxor:
It looks like the GHC version it's using is 8.8.2.
```
$ stack ghc -- --version
Stack has not been tested with GHC versions above 8.6, and using 8.8.2, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.0.1.0 was found, this may fail
The Glorious Glasgow Haskell Compilation System, version 8.8.2
```

chexxor:
When I cloned, I did not add the submodule flag.

That is I did this: `git clone https://github.com/haskell/haskell-ide-engine`

Where the README says to do this: `git clone https://github.com/haskell/haskell-ide-engine --recurse-submodules`

Could this be part of the problem?

chexxor:
It looks like the current version of the `.gitmodules` file is empty, so maybe submodules are no longer used in the project?

[...]

PR #1732, opened 2020-04-10
Clarify "Building", remove "submodules" in README

chexxor:
I had some confusion when following the install/build instructions. I think these changes might improve them. Let me know what you think.

https://github.com/haskell/haskell-ide-engine/issues/1731

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ cd haskell-ide-engine
 
 #### Building
 
-Note, on first invocation of the build script, a GHC is installed for execution. If you'd like to use a specific version of GHC for the `install.hs`, you can adjust the resolver in the `install/shake.yaml` file, as this project's build uses the [shake](https://shakebuild.com/) build tool.
+Note, on first invocation of the build script, a GHC is installed for execution. If you'd like to use a specific version of GHC for the `install.hs`, you can adjust the resolver in the `install/shake.yaml` file.
 
 Available commands can be seen with:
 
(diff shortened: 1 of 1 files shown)


jneira:
Hi! The project has not submodules now but it had them for a long time. It is not probable (but possible) that it will need again and `--recurse-submodules` does not throw an error if there are not actual submodules.

Otoh as it is not probable and the instructions are shorter this way, i think kwe could merge it.

jneira (approved):

fendor:
Installation script executes `git submodule update --init` anyways, right? So, we are still future-proof in a way

README.md
```diff
@@ -226,16 +226,13 @@ In addition make sure `hie.exe` is not running by closing your editor, otherwise
...
 cd haskell-ide-engine
 ```
 
 #### Building
 
-Uses the [shake](https://shakebuild.com/) build system for predictable builds.
-
-Note, on first invocation of the build script, a GHC is being installed for execution.
-The GHC used for the `install.hs` can be adjusted in `shake.yaml` by using a different resolver.
+Note, on first invocation of the build script, a GHC is installed for execution. If you'd like to use a specific version of GHC for the `install.hs`, you can adjust the resolver in the `install/shake.yaml` file, as this project's build uses the [shake](https://shakebuild.com/) build tool.
```
fendor: > as this project's build uses the [shake](https://shakebuild.com/) build tool.

That is not really the reason why that works, it works because it is a regular haskell project that just happens to use shake.

PR #1736, opened 2020-04-13
Strip RTS flags, since we cant honor them

### LINKED ISSUE #1725: Error "target \8216+RTS\8217 is not a module name or a source file" in Atom
I've just upgraded to the latest HIE as part of a move to Stack lts-15.2. When I save a file in Atom (triggering HIE) I get the error message
`target \8216+RTS\8217 is not a module name or a source file`
in a red box at the top right.

I'm filing this as an HIE bug because it is HIE that has changed rather than the Atom plug-in.

fendor:
Closes #1725

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -543,7 +543,7 @@ cabalHelperAction proj env package root fp = do
         let fs' = getFlags comp
         let fs = map (fixImportDirs root) fs'
         let targets = getTargets comp relativeFp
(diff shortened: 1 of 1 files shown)


jneira:
The ghc user guide says:
```
No -RTS option is required if the runtime-system options extend to the end of the command line, as in this example:

% hls -ltr /usr/etc +RTS -A5m

If you absolutely positively want all the rest of the options in a command line to go to the program (and not the RTS), use a --RTS.
```
will the code have to handle those cases?

fendor:
Nope, this only for loading a specific component. With cabal-helper, we obtain the ghc-options flags and we need to filter them, since we cant use them in an existing GHC session.

PR #1737, opened 2020-04-13
Use installed ghc in stack linux builds

### LINKED ISSUE #1729: Azure stack linux jobs fail due to disk space
Well, linux stack jobs have ran out of disk space since c8ed3e5 :
```
Filesystem      Size  Used Avail Use% Mounted on
udev            3.4G     0  3.4G   0% /dev
tmpfs           695M  8.9M  686M   2% /run
/dev/sda1        84G   80G  3.6G  96% /
tmpfs           3.4G  8.0K  3.4G   1% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.4G     0  3.4G   0% /sys/fs/cgroup
/dev/loop0       40M   40M     0 100% /snap/hub/43
/dev/loop1       94M   94M     0 100% /snap/core/8935
/dev/sda15      105M  3.6M  101M   4% /boot/efi
/dev/sdb1        14G   35M   13G   1% /mnt
```

It make the download of ghc-8.8.1, necessary for wrapper-tests fail:

```
Installing GHC ...
/home/vsts/.stack/programs/x86_64-linux/ghc-8.8.1/lib/ghc-8.8.1/Cabal-3.0.0.0/.copyFile75307-1209.tmp: copyFile: resource exhausted (No space left on device)
make[1]: *** [install_packages] Error 1
make: *** [install] Error 2
Run from: /home/vsts/.stack/programs/x86_64-linux/ghc-8.8.1.temp/ghc-8.8.1/


Error: Error encountered while installing GHC with
         make install
         run in /home/vsts/.stack/programs/x86_64-linux/ghc-8.8.1.temp/ghc-8.8.1/
       
       The following directories may now contain files, but won't be used by stack:
         - /home/vsts/.stack/programs/x86_64-linux/ghc-8.8.1.temp/
         - /home/vsts/.stack/programs/x86_64-linux/ghc-8.8.1/
       
       For more information consider rerunning with --verbose flag
       

Command `stack ghc -- --numeric-version` failed.

version checking
  picks up a stack.yaml with 8.8.1 FAILED [1]
  picks up a stack.yaml with 8.6.5
  picks up whatever version of ghc is on this machine

Failures:

  test/wrapper/HieWrapper.hs:18:9: 
  1) version checking picks up a stack.yaml with 8.8.1
       expected: "8.8.1"
        but got: "8.8.3"
```

jneira:
* To avoid download it again and not waste disk space
* Fixes #1729 
* The builds are failing due to timeouts when executing `stack exec hoogle generate`: see https://github.com/ndmitchell/hoogle/issues/347

### DIFF
--- a/.azure/linux-stack.bashrc
+++ b/.azure/linux-stack.bashrc
@@ -1,1 +1,1 @@
-export PATH=$HOME/.local/bin:$PATH
+export PATH=$HOME/.local/bin:/opt/ghc/$GHC_VERSION/bin:$PATH
--- a/.azure/linux-stack.yml
+++ b/.azure/linux-stack.yml
@@ -90,7 +90,7 @@ jobs:
     displayName: "Install Runtime Test-Dependencies: z3"
   - bash: |
       source .azure/linux-stack.bashrc
-      PATH=/opt/ghc/8.6.5/bin:$PATH
+      export PATH=/opt/ghc/8.6.5/bin:$PATH
       stack install --resolver=lts-13.20 liquid-fixpoint-0.8.0.2 liquidhaskell-0.8.6.2
     displayName: "Install Runtime Test-Dependencies: liquidhaskell"
   - bash: |
@@ -106,6 +106,8 @@ jobs:
     displayName: "Run Test: unit-test"
   - bash: |
       source .azure/linux-stack.bashrc
+      # we need the ghc versions used by wrapper-test suite
+      export PATH=$PATH:/opt/ghc/8.8.1/bin:/opt/ghc/8.6.5/bin
       stack test :dispatcher-test :plugin-dispatcher-test :wrapper-test --stack-yaml $(YAML_FILE)
     displayName: "Run Test: dispatcher-test, plugin-dispatcher-test and wrapper-test"
   - bash: |
--- a/.azure/linux-stack.yml
+++ b/.azure/linux-stack.yml
@@ -5,29 +5,31 @@ jobs:
     vmImage: ubuntu-16.04
   strategy:
     matrix:
-      stack-def:
+      default:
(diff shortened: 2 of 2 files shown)


jneira:
I am afraid that the actual configuration is not valid for wrapper-test, cause it assumes that stack will have the correct ghc downloaded or in path. It needs up to two different ghcs (ghc-8.8.1 and ghc-8.6.5) so not sure if it will be enough space for both although we are not downloading the main one, the used to build everything else.

fendor:
How about we mock the wrapper tests? You think that is possible?

fendor:
Idea for test-mocking:
* Add a number of scripts named `ghc-*` where `*` is the GHC version
* Each of these scripts only know the parameter `--numeric-version`
* When running wrapper tests, we do something like `export PATH=/path/to/scripts:$PATH`
* For `stack`, set `system-ghc: True`.

jneira:
Mmm, interesting, but stack with `system-ghc` needs an executable `ghc` (without the version suffix) in path, no (i have to test it)? could we cheat stack and make it think that the script is ghc?
In fact, we have the real ghcs! so we could point to the appropiate one for each test. This can be done manually running each test separately and set the path with the appropiate ghc in the bash script.

jneira:
> but stack with system-ghc needs an executable ghc (without the version suffix) in path, no? (i have to test it) 
I was wrong, stack picks a ghc-${version} in path if it needs it so the solution is simpler than i thought: we can set in path the needed ghcs, putting the main one in first place

jneira:
Ok, the stack linux jobs that has not failed due to the response timeout are successful now, so i think the pr could be merged

PR #1742, opened 2020-04-18
Update Sublime Text HIE command

ssanj:
The `hie` command needs to include the `--lsp` parameter for it to work as an LSP server.


fendor:
Thank you for catching this!

PR #1750, opened 2020-04-22
Move finding the package for a filepath

### LINKED ISSUE #1640: hie-wrapper should not build all dependencies
Happens only with cabal-helper, when cabal-helper initialises the context, it builds all deps of the package and may take a long time before responding, rendering our progress notification useless. Solution might include, moving initialisation code higher up the chain.

fendor:
Closes #1640

Currently, hie-wrapper uses the cradle type to find the ghc version to
use on the project. Cabal-Helper does not only return the cradle type
but also initialises the packages of the project, which may take a long
time since it starts building dependencies. Moreover, it causes
hie-wrapper to take way longer than necessary to find the project type,
and thus, the ghc version to use on the project.
This violates the isolation of the cradle, some work happens before
loading the options for a filepath, some during it.
This commit unifies the behaviour with hie-bios:

  * All the work happens during loading the options for the given
    filepath.

It speeds up the initial start-up of hie if users use the implicit
cradle discovery mechanism.
It makes the implementation a bit less hacky.

While there are behavioural changes, nothing should change for everyday
users.

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Cradle.hs
@@ -14,10 +14,11 @@ import           Distribution.Helper (Package, projectPackages, pUnits,
                                       unChModuleName, Ex(..), ProjLoc(..),
                                       QueryEnv, mkQueryEnv, runQuery,
                                       Unit, unitInfo, uiComponents,
-                                      ChEntrypoint(..), UnitInfo(..))
+                                      ChEntrypoint(..), UnitInfo(..),
+                                      pPackageName)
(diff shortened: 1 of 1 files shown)


fendor:
@jneira This change is also relevant for haskell/haskell-language-server

jneira:
Cabal unix jobs are failing due to disk space, so i think the pr is good

fendor:
Small revision of the explanation, it only configures the projects. However, it crashes in hie-wrapper, if configuration fails, e.g. dependencies have conflicting version constraints.
Now we can correctly catch these exceptions and display them to the user, similar to how we catch exceptions during building units.

PR #1770, opened 2020-05-12
Fix stack.yaml syntax for a specific revision

Infinisil:
See also https://github.com/haskell/haskell-ide-engine/pull/1694#issuecomment-627456134

Ping @jneira @fendor @gdziadkiewicz

### DIFF
--- a/stack-8.8.3.yaml
+++ b/stack-8.8.3.yaml
@@ -7,7 +7,7 @@ extra-deps:
 # - ./submodules/HaRe
 
 - apply-refact-0.7.0.0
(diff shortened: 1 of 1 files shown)


Infinisil:
Hm actually hold on, I looked into this because of https://github.com/input-output-hk/haskell.nix/issues/550, but I'm not sure it fixes it now

Infinisil:
Alright it's just a separate bug in nix-tools, this change here is good :)

Infinisil:
See https://github.com/input-output-hk/nix-tools/pull/90 for the fix of the other bug

PR #1771, opened 2020-05-12
Use cabal-helper 1.1

### LINKED ISSUE #1758: Update cabal-helper to 1.1.0.0
It has several bug fixes

jneira:
* Closes #1758 
* It has needed:
  * Use Cabal-3.0.1.0 in all build plans, cause cabal-helper requires it for windows, to avoid a bug in previous versions (for windows)
  * To being able to use it we had to request revisions for haddock-api, see https://github.com/DanielG/cabal-helper/issues/116. We are using them in stack.yaml's
  * Other packages has been upgraded, to get versions compatibles with Cabal-3.0.2.0
  * There were some conditions over Cabal api changes using the ghc version instead, the code failed when changing the cabal version and now it is more correct

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -4,11 +4,6 @@ packages:
 
          -- ./submodules/HaRe
 
-source-repository-package
-    type: git
-    location: https://github.com/DanielG/cabal-helper.git
-    tag: a18bbb2af92e9b4337e7f930cb80754f2408bcfd
-
 tests: true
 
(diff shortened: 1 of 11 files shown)


jneira:
I think ci failing builds are the same introduced since aca3241

jneira:
All ci builds failed at unit test step so the build step works

PR #1787, opened 2020-06-16
Update README.md

JohnTasto:
`zlib1g-dev` is missing in at least the WSL2 version of Ubuntu 20.04, which causes the build to fail.


jneira:
@JohnTasto thanks a lot for complete docs!

PR #1794, opened 2020-07-23
Rename configuration section languageServerHaskell => haskell

bubba:
Brings it inline with other language servers, and also fixes it with vscode-hie-server 0.1.
A breaking change for people's configurations, but the ship has already sailed on vscode-hie-server, whoops!

### DIFF
--- a/hie-plugin-api/Haskell/Ide/Engine/Config.hs
+++ b/hie-plugin-api/Haskell/Ide/Engine/Config.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Haskell.Ide.Engine.Config where
 
+import           Control.Applicative
 import           Data.Aeson
 import           Data.Default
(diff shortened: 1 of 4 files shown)


hie-plugin-api/Haskell/Ide/Engine/Config.hs
```diff
@@ -54,7 +54,7 @@ instance Default Config where
 -- TODO: Add API for plugins to expose their own LSP config options
 instance FromJSON Config where
   parseJSON = withObject "Config" $ \v -> do
-    s <- v .: "languageServerHaskell"
+    s <- v .: "haskell"
```
alanz, line 57: I think we need to accept both versions initially, here and in hls, otherwise we are forcing every client to make the change immediately, and banning them from running older versions of hie/hls.

bubba:
The failing tests seem to be affecting all other PRs, merging

PR #1802, opened 2020-10-02
Use default config on missing configuration section

aufarg:
Use default config on missing configuration section

On serving initialize request, the deserialization of HIE configuration
embedded in InitializeParam passed by client will result in an error if
during the deserialization process the server cannot find HIE specific
configuration key under initializationOptions.

This commit changes the initializationOptions deserialization to return
the default configuration if configuration key cannot be found under
initializationOptions. Here, setting the key with a value of null will
also be considered as part of not found condition to accommodate clients
that fills missing user options as null.


alanz:
@aufarg This project is not actually active any more, so PRs will not be merged.  I see you did do something on haskell-language-server too, which is the project superseding this one

aufarg:
@alanz Okay, I'll close the PR then. Thanks for telling me this.

